Issue #86, opened 2016-10-05
provide a way to forward exceptions in the recv-thread


umlaeute:
i'm using `socketcan` as backend.

whenever the interface goes down (`ifconfig vcan0 down`), the actual receiver gets an `OSError` exception.
unfortunately this happens in the receiver-thread, and there is no way to be notified of this event.

i'd propose a new (optional) argument `execption_callback` to the `Notifier` class that get's called (with the execption as argument) whenever such an exception is caught.

umlaeute:
so after sleeping this over i think that callback is probably an overkill.

instead i would like to propose to just catch "all" exceptions than can be appear while receiving data (e.g. `OSError`) and return `None`, like with other exceptions.

the exceptions should probably be logged...

hardbyte:
I suggest we catch the `OSError`s in `Bus.recv` and then throw a `can.CanError`. 

When does `recv` return `None` other than timeout?

Agree that exceptions in the `Notifier` object's  recv thread should be logged.

umlaeute:
i'm pretty sure that i had to catch exceptions in two distinct places (using `socketcan_native`) in order to properly deal with a `can0` device that was taken down during operation.

i'll try to be more specific when i have access to the the system again...

PR #87, opened 2016-10-18
capture OSErrors while receiving data

umlaeute:
both sock.recv() and select.select(sock) both throw errors when the device
to which the socket is bound goes down.
this patch should make the notifier thread survive such a condition
(and bail out on the log)

Closes: https://github.com/hardbyte/python-can/issues/86

### DIFF
--- a/can/interfaces/socketcan_native.py
+++ b/can/interfaces/socketcan_native.py
@@ -278,6 +278,10 @@ def capturePacket(sock):
     except socket.timeout:
         log.debug('Captured no data, socket read timed out.')
         return None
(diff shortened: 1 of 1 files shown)


umlaeute:
btw, this code is *slightly* untested as it is a cleaned up version of what i actually used (the original hacked version would catch any `Exception` (rather than just an `OSError`) and wrap almost the entire body of `SocketcanNative_Bus.recv` into a a big `try/catch` clause.

i'm pretty confident that the code does work though, but have currently no possibility to actually test it.

PR #88, opened 2016-10-18
compat: sqlite3's BLOB requires a buffer() in py2

umlaeute:
on python2, it seem like we need to convert the `msg.data` to `buffer` first, before we can insert it into a `BLOB` column.

otoh, in python3 there is no `buffer`.

so we just define a function `msg2buf()` that is either `buffer()` or an identity function.

### DIFF
--- a/can/CAN.py
+++ b/can/CAN.py
@@ -18,6 +18,12 @@ try:
 except ImportError:
     import Queue as queue
 
+try:
+    msg2buf = buffer
+except NameError:
+    # in Py3 there is no buffer(), so we use an identity function
+    msg2buf = lambda x: x
+
 log = logging.getLogger('can')
 log.debug("Loading python-can")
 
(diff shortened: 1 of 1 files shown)


hardbyte:
Can we use `memoryview` for [Python3](https://hg.python.org/cpython/file/3.5/Lib/sqlite3/dbapi2.py#l53) and `buffer` for Python2?

I haven't thought too much about how to best detect versions but maybe this would be better than catching the name error:
```
if sys.version_info > (3,):
    buffer = memoryview
```

(from http://python3porting.com/differences.html#buffer)

umlaeute:
regarding `memoryview`: maybe that's even better, i don't know; I used the identity function as a quick working solution; there might be better ones.

regarding `sys.version_info` i guess this is a matter of taste: personally i find it preferable to not clutter the code with a-priori knowledge of what a specific interpreter version can do or not, and just `try` whether a given construct is supported. but `python-can` already has a number of `sys.version_info` checks, so using them would be more consistent.

hardbyte:
Thanks @umlaeute I've just committed the `memoryview` version.

PR #92, opened 2016-11-10
National Instruments interface support

christiansandberg:
This adds an interface for [NI-CAN](http://www.ni.com/can/).

I did not seem to get filtering working but can't find more information on how to do it. Maybe it only works for some hardware...

Raising timeout error on failed transmission was not possible since multiple threads could not wait for a state change on the same channel at the same time.


can/interfaces/nican.py
```diff
@@ -0,0 +1,282 @@
...
+            can_id = can_filters[0]["can_id"]
+            can_mask = can_filters[0]["can_mask"]
+            logger.info("Filtering on ID 0x%X, mask 0x%X", can_id, can_mask)
+            # This did not filter out anything when tested but I don't know why
+            config[NC_ATTR_CAN_COMP_STD] = can_id
+            config[NC_ATTR_CAN_MASK_STD] = can_mask
+            config[NC_ATTR_CAN_COMP_XTD] = can_id
+            config[NC_ATTR_CAN_MASK_XTD] = can_mask
+        else:
+            logger.warning("Only one filter is supported")
```
christiansandberg: Do you mean a logging error or an exception? If the filter is used as an optimization then it will probably not be a big deal, just that the application uses unnecessary resources processing all messages on the bus.

christiansandberg:
Filtering should be done by the book now but it still doesn't filter anything. :confused: I think I'll just leave it as it is for now.

christiansandberg:
Seems like the USB device I have for testing do not actually support HW filtering:

> Given that certain NI-CAN devices, such as the USB-8473, do not support CAN Objects with the NI-CAN API, it is not possible to filter the data read in the bus for one particular device at the hardware/driver level.

https://decibel.ni.com/content/docs/DOC-16992

hardbyte:
Ok looks good to me. You're happy with how it handles multiple thread access?

Can you just add a couple of lines to the installation docs at `doc/installation.rst` - specifically where to find the requisite drivers and any other constraints (OS versions etc).

christiansandberg:
It will handle `recv()` and `send()` being called from separate threads which I think is what can be expected.

PR #95, opened 2016-11-24
Optimize the sqlite3 logger. Closes #94

### LINKED ISSUE #94: sqlite logger not keeping up with messages
Using a Raspberry PI and a PICAN 2 using the `socketcan_ctypes` backend I can't log all the traffic to an sqlite database. However writing to csv or txt seems to do fine.

This is probably because the `on_message_received` method of the `SqliteWriter` inserts the new CAN frame directly into the database - incurring the overhead of creating a transaction for each frame.

Instead we could buffer a number of frames (using the `BufferedReader`) then bulk insert them into the database using `executemany`. Would have to ensure that we flush the buffer when the Listener is stopped.

http://stackoverflow.com/questions/18219779/bulk-insert-huge-data-into-sqlite-using-python

hardbyte:
This PR adds a new thread to the sqlite logger and buffers messages and writes *many* can frames to the database instead of individually.


hardbyte:
Needs rebasing/merging.

hardbyte:
Performance is much better now. Anyone able to do a code review?

hardbyte:
Cheers, made those changes.

Issue #82, opened 2016-09-14
Ixxat on Windows 7 x64


hardbyte:
Originally reported by: **Mathias Giacomuzzi (Bitbucket: [mgiaco](https://bitbucket.org/mgiaco), GitHub: [mgiaco](https://github.com/mgiaco))**

---

Hello,

I already used peak can works. Today I get an Ixxat V2 I tried to use it now with python-can with no success. So can someone give me a hint how I get the ixxat to work with python-can?

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/82

PR #100, opened 2016-12-02
Added some docs about IXXAT-windows configuration file options

cowo78:
See issue #82 . Couple of documentation lines about config file. Based on develop branch.


christiansandberg:
Shouldn't it be merged to develop instead?

PR #102, opened 2016-12-04
Feature canplayer

hardbyte:
Add framework support for replaying messages from the files we create using `can_logger.py`.
I implemented the sql db reader, and added a new script to replay with correct delays to another bus.

### DIFF
--- a/bin/can_player.py
+++ b/bin/can_player.py
@@ -8,10 +8,9 @@ Similar to canplayer in the can-utils package.
 from __future__ import print_function
 import datetime
 import argparse
-import time
-import socket
 
 import can
+from CAN import LogReader, MessageSync
 
 if __name__ == "__main__":
 
@@ -35,7 +34,6 @@ if __name__ == "__main__":
                         fall back to reading from configuration files.''',
                         choices=can.interface.VALID_INTERFACES)
 
-
     parser.add_argument('--ignore-timestamps', dest='timestamps',
                         help='''Ignore timestamps (send all frames immediately with minimum gap between
     frames)''', action='store_false')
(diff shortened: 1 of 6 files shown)


bin/can_player.py
```diff
@@ -0,0 +1,74 @@
...
+        datetime.datetime.now()))
+    print('Replaying data from {}'.format(
+        player.recorded_start_datetime))
+
+    try:
+        for m in in_sync:
+            bus.send(m)
+
+    except KeyboardInterrupt:
+        bus.shutdown()
```
christiansandberg: `shutdown()` should be called also after completed execution.

can/CAN.py
```diff
@@ -1,308 +1,379 @@
...
+        self.recorded_start_datetime = datetime.fromtimestamp(self.recorded_start_time)
+
+    @staticmethod
+    def create_frame_from_db_tuple(frame_data):
+        ts, id, is_extended, is_remote, is_error, dlc, data = frame_data
+        return Message(
+            ts, is_remote, is_extended, is_error, id, dlc, data
+        )
+
+    def __iter__(self):
```
christiansandberg: Should any open resources be closed after successful iteration, in a separate `.close()` method, or can we rely on the garbage collection?

can/CAN.py
```diff
@@ -1,308 +1,379 @@
...
+
+class SqlReader:
+    def __init__(self, filename):
+        log.debug("Starting sqlreader with {}".format(filename))
+        conn = sqlite3.connect(filename)
+
+        self.c = conn.cursor()
+        self.c.execute("SELECT ts FROM messages LIMIT 1")
+
+        self.recorded_start_time = self.c.fetchone()[0]
```
christiansandberg: Wouldn't it be better if this was done by can_player.py instead? If it is needed at all...

hardbyte: Ahh good catch I had already moved that to the `MessageSync` in `util.py`:
```
        for m in self.raw_messages:
            if recorded_start_time is None:
                recorded_start_time = m.timestamp
```

hardbyte: I think I'll remove that message about when the trace was recorded though.

can/CAN.py
hardbyte: I think that we can rely on the garbage collector - especially because we aren't writing to the db here.

Issue #80, opened 2016-08-31
Periodic send api


hardbyte:
Originally reported by: **Brian Thorne (Bitbucket: [hardbyte](https://bitbucket.org/hardbyte), GitHub: [hardbyte](https://github.com/hardbyte))**

---

The interfaces in `can/broadcastmanager.py` should take python-can `Bus` instances not channels - this would be more backend agnostic.

It might be worth adding this periodic api directly to the bus itself too.

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/80

hardbyte:
So here is what I'm thinking, as always I'd appreciate any thoughts and feedback. Primarily I want the API to be as simple as possible for users, and also straightforward to implement for each backend.

I'm in favor of extending the `Bus` directly with a `send_periodic` method. This method could either take a `CyclicSendTask` argument or, a `Message` and period information. I'm not quite convinced either way yet so I'd like to discuss the pros and cons...

Using a `CyclicTask` object:
- (-) More complexity for users. Most users probably just want the most basic of periodic sending to work really well and I suspect an api like `bus.send_periodic(msg, period, repeats)` might cater to most use cases.
- (+) More flexibility for users. We could use the same Bus method to pass a few different types of tasks (e.g `MultiRateCyclicSendTask` or more complex instructions). We could have a library dispatcher and default pure python implementations and leave it up to each interface to override pieces.
- (+) Might be more explicit because the user would have created a "Task" object which could later be used to stop and modify the task. As opposed to sending a periodic message and having to keep the returned `PeriodicTask`.
- What else?

This public api Bus level method would probably return an object like `PeriodicTask` which could be used to modify or cancel it later. This has the potential for introducing threading issues so would have to be designed with something like queues or events to prevent them.

We can implement a concrete default implementation of this entire idea in the Bus's abstract base class and each interface would be encouraged to pass off to the hardware where it makes the most sense for them. IMO The tricky part is working out the api.

christiansandberg:
It's a tricky question. I think it should be started using an existing Bus instance, since some interfaces require a handle and even if they only require a channel name then that could be read from the Bus instance. I guess usually you would like to both send and receive messages.

We could have both an API with a CyclicTask class where you pass a Bus instance as argument and a ``bus.send_periodic()`` method which could be a short-hand for creating that class.

[...]

PR #106, opened 2016-12-07
Remote TCP/IP interface

christiansandberg:
This is pull request adds a special kind of interface which acts like a TCP/IP bridge between the application using python-can and a server which handles the physical interface on a different computer.

I think it's working pretty well now after I fixed an issue with closing the connection correctly.

The cyclic transmit API has not been fully implemented mostly because it doesn't fit well with backends other than socketcan. I think after #80 has been fixed, I can continue working on supporting that part.

I think it would be a nice addition to 2.0.

[Original pull request on Bitbucket](https://bitbucket.org/hardbyte/python-can/pull-requests/27/).

### DIFF
--- a/bin/can_server.py
+++ b/bin/can_server.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python
+import logging
+FORMAT = '%(asctime)-15s %(message)s'
+logging.basicConfig(format=FORMAT)
+import argparse
+import can
+from can.interfaces import remote
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Remote CAN server")
+
+    parser.add_argument('-v', action='count', dest="verbosity",
+                        help='''How much information do you want to see at the command line?
+                        You can add several of these e.g., -vv is DEBUG''', default=2)
+
+    parser.add_argument('-c', '--channel', help='''Most backend interfaces require some sort of channel.
+    For example with the serial interface the channel might be a rfcomm device: "/dev/rfcomm0"
+    With the socketcan interfaces valid channel examples include: "can0", "vcan0"''')
+
+    parser.add_argument('-i', '--interface',
+                        help='''Specify the backend CAN interface to use. If left blank,
+                        fall back to reading from configuration files.''',
+                        choices=can.interface.VALID_INTERFACES)
+
+    parser.add_argument('-p', '--port', type=int,
+                        help='''TCP port to listen on (default %d).''' % remote.DEFAULT_PORT,
+                        default=remote.DEFAULT_PORT)
+
+    results = parser.parse_args()
+
+    verbosity = results.verbosity
+    logging_level_name = ['critical', 'error', 'warning', 'info', 'debug', 'subdebug'][min(5, verbosity)]
+    can.set_logging_level(logging_level_name)
+
+    server = remote.RemoteServer(results.port,
+                                 channel=results.channel,
+                                 bustype=results.interface)
+    server.start()
--- a/bin/can_server.py
+++ b/bin/can_server.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
 import logging
-FORMAT = '%(asctime)-15s %(message)s'
-logging.basicConfig(format=FORMAT)
+logging.basicConfig(format='%(asctime)-15s %(message)s')
 import argparse
 import can
 from can.interfaces import remote
 
-if __name__ == "__main__":
+
+def main():
     parser = argparse.ArgumentParser(description="Remote CAN server")
 
     parser.add_argument('-v', action='count', dest="verbosity",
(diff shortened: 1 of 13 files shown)


bin/can_server.py
```diff
@@ -0,0 +1,38 @@
...
+from can.interfaces import remote
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Remote CAN server")
+
+    parser.add_argument('-v', action='count', dest="verbosity",
+                        help='''How much information do you want to see at the command line?
+                        You can add several of these e.g., -vv is DEBUG''', default=2)
+
+    parser.add_argument('-c', '--channel', help='''Most backend interfaces require some sort of channel.
```
christiansandberg: I guess that would make sense. My thought was that since we do not have any official default bitrate today, then if the application does not specifiy a bitrate it will be backend dependent. However, since the application and physical CAN bus is separated here, the application might not know what the bitrate should be, therefore not providing one and relies on the server to know what bitrate is appropriate.

bin/can_server.py
```diff
@@ -0,0 +1,38 @@
...
+    results = parser.parse_args()
+
+    verbosity = results.verbosity
+    logging_level_name = ['critical', 'error', 'warning', 'info', 'debug', 'subdebug'][min(5, verbosity)]
+    can.set_logging_level(logging_level_name)
+
+    server = remote.RemoteServer(results.port,
+                                 channel=results.channel,
+                                 bustype=results.interface)
+    server.start()
```
christiansandberg: Good idea!

doc/interfaces/remote.rst
```diff
@@ -0,0 +1,94 @@
...
+The remote interface works as a networked bridge between the computer running
+the application and the computer owning the physical CAN interface.
+
+Server
+------
+
+The computer which owns the CAN interface must start a server which accepts
+incoming connections. If more than one channel is to be shared, multiple
+servers must be started on different ports.
+
```
christiansandberg: Will do. Since we create one bus per connection, it might be backend dependent if it will allow multiple connections.

doc/interfaces/remote.rst
```diff
@@ -0,0 +1,84 @@
...
+.. autoclass:: can.interfaces.remote.server.ClientBusConnection
+
+Protocol
+~~~~~~~~
+
+The protocol is a stream of events over a network socket.
+Each event starts with one byte that represents the event id, followed by
+event specific data of arbitrary length in big-endian byte order.
+
+The client start with sending a :class:`can.interfaces.remote.events.BusRequest`
```
christiansandberg: Ah I didn't know that trick.

doc/interfaces/remote.rst
```diff
@@ -0,0 +1,94 @@
...
+
+      Stop the server.
+
+
+.. autoclass:: can.interfaces.remote.server.ClientBusConnection
+
+Protocol
+~~~~~~~~
+
+The protocol is a stream of events over a network socket.
```
christiansandberg: :thumbsup:

can/interfaces/remote/events.py
```diff
@@ -0,0 +1,451 @@
...
+class BaseEvent(object):
+    """Events should inherit this class."""
+
+    def encode(self):
+        """Convert event data to bytes.
+
+        :return:
+            Bytestring representing the event data.
+        """
+        return b''
```
christiansandberg: I'm taking advantage of the default when the event does not contain any data. Since it is not a public API I think it makes sense to have a default behavior.

can/interfaces/remote/events.py
```diff
@@ -0,0 +1,451 @@
...
+
+    def __init__(self, version, bitrate):
+        """
+        :param int version:
+            Network protocol version
+        :param int bitrate:
+            Bitrate to use on CAN
+        """
+        #: Network protocol version
+        self.version = version
```
christiansandberg: The version is constant (can.interfaces.remote.PROTOCOL_VERSION) so it might be unnecessary to have it in the constructor. It was more for consistency.

can/interfaces/interface.py
```diff
@@ -96,6 +99,9 @@ def __new__(cls, other, channel, *args, **kwargs):
         elif can.rc['interface'] == 'socketcan_native':
             from can.interfaces.socketcan_native import CyclicSendTask as _nativeCyclicSendTask
             cls = _nativeCyclicSendTask
+        #elif can.rc['interface'] == 'remote':
```
christiansandberg: I started preparing for adding support for the cyclic API but it got too complicated so I put it on hold. I think the proposal made in #80 to have it associated with an already existing bus object would result in a cleaner implementation. I can add some info in this comment and also hide all cyclic API from the documentation until it is fully implemented.

hardbyte:
Ok thanks for making those changes - looks good to me.

HeathKang:
Hi, I was confused about your code about events.py in the remote interface.It seems like TCP server will send specific event id to client, after client send data to server. Am I right?

christiansandberg:
Well, the client starts with sending two events (1 and 10) and the server should respond with a response event (2). I've tried to describe the protocol [in the documentation](http://python-can.readthedocs.io/en/latest/interfaces/remote.html#protocol).

HeathKang:
Thanks for your help. And I wonder if there any library about CAN message decode by dbc file.

christiansandberg:
Have a look at this [canmatrix pull request](https://github.com/ebroecker/canmatrix/pull/98).

HeathKang:
Thanks!!

Issue #109, opened 2016-12-20
Extended ID flag ignored for period messages when using socketcan_native


jboomer:
I found that when attempting to send a period message using the 'socketcan_native' interface, the 'extended_id' argument in the constructor for the message is ignored.

I have a small test program to test this : 
```python
#! /usr/bin/env python3

import can
import time

INTERFACE = 'socketcan_native'
CHANNEL = 'vcan0'

can.rc['interface'] = INTERFACE
can.rc['channel'] = CHANNEL

from can.interfaces.interface import Bus

bus = Bus()
msg = can.Message(arbitration_id = 0x12345678,
                  data = [0xAA] * 8,
                  extended_id = True)
period = 1.0

task = (can.send_periodic(CHANNEL, msg, period))

time.sleep(10)

task.stop()
bus.shutdown()
```

The candump output for this is something like : 
```
 (1482224446.405665)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224447.405739)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224448.405804)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224449.405874)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224450.405956)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224451.406010)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224452.406099)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224453.406164)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224454.406242)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224455.406313)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
 (1482224456.406322)  vcan0  678   [8]  AA AA AA AA AA AA AA AA
```
So it looks like the ID is the standard 11 bits only.

When using 'socketcan_ctypes' the id is 12345678 as expected.

I found this can be worked around by setting bit 32 in the arbitration id, so something like : 
```python
arbitration_id= 0x12345678 | (1 << 31)
```

But this should not be necessary right?

I am using Ubuntu 16.04, with Python version 3.5.2

hardbyte:
You're quite correct you shouldn't have to manually change the `can_id`. This should fix it.

PR #110, opened 2016-12-20
Extended ID flag ignored on socketcan_native

hardbyte:
Addresses #109 

Could be further refactored as it is the same handling as `socketcan_ctypes`.


christiansandberg:
I believe ``modify_data()`` would fail now.

hardbyte:
Thanks! I did catch that but for some reason pushing the third commit failed.

PR #111, opened 2016-12-20
Small fixes in IXXAT exception handling plus default handling in `Bus.__new__`

cowo78:
Exceptions are now simple CanError subclasses and do not depend on ctypes to decode error codes. Decoding is now done in canlib.
Not related to IXXAT but `interface.Bus.__new__` was not updating can.rc from its kwargs interface and bitrate.
Apply to the develop branch.


can/interfaces/ixxat/canlib.py
```diff
@@ -40,6 +45,25 @@
     log.warning("IXXAT VCI library does not work on %s platform", sys.platform)
     _canlib = None
 
+def __vciFormatErrorExtended(library_instance, function, HRESULT, arguments):
+    """ Format a VCI error and attach failed function, decoded HRESULT and arguments
+        @TODO: make sure we don't generate another exception
+    """
+    buf = ctypes.create_string_buffer(VCI_MAX_ERRSTRLEN)
```
hardbyte: To avoid repeating the same code just call `__vciFormatError` here and add the arguments to the returned string.

can/interfaces/ixxat/canlib.py
```diff
@@ -308,11 +336,10 @@ def recv(self, timeout=None):
...
             except VCITimeout:
                 return None
+            except VCIRxQueueEmptyError:
+                return None
             except VCIError as e:
-                if (e.HRESULT == constants.VCI_E_RXQUEUE_EMPTY):
-                    return None
-                else:
-                    raise e
+                raise e
```
hardbyte: Unless you want to add some logging in the event of a `VCIError` you don't need to catch and re-raise this.

hardbyte:
Thanks Giuseppe

PR #112, opened 2016-12-24
Refactoring interfaces into own packages

hardbyte:
housework


hardbyte:
Need to update this now that @cowo78's ixxat fixes have been merged into develop.

PR #114, opened 2017-01-04
Add Vector CAN Interface

JulienGrv:
This is the CAN interface for Vector devices. Of course, the windows dll is required in order to make it work, it is available on Vector's website.
The API is a Ctypes wrapper around the windows dll.


can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        :param list channel:
+            The channel indexes to create this bus with.
+        """
+        if vxlapi is None:
+            raise ImportError("The Vector API has not been loaded")
+        if isinstance(channel, (list, tuple)):
+            self.channel = channel
+        else:
+            raise ValueError('channel must be a list of integers')
```
christiansandberg: Could you also add support for accepting a string of comma separated integers as well? Otherwise it won't be possible to use config file or command line tools.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+        xl_status = self._driver_init()
+        if xl_status.value == vxlapi.XL_SUCCESS:
+            xl_status = self._go_on_bus()
+        if xl_status.value != vxlapi.XL_SUCCESS:
+            self.shutdown()
```
christiansandberg: Raise an exception with error description from driver.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+        return xl_status
+
+    def recv(self, timeout=None):
+        event_count = ctypes.c_uint(1)
+        event_list = vxlapi.XLevent(0)
+        xl_status = vxlapi.receive(self.port_handle, event_count, event_list)
+        if xl_status.value != vxlapi.XL_ERR_QUEUE_IS_EMPTY:
+            if xl_status.value != vxlapi.XL_SUCCESS:
+                err_string = vxlapi.get_error_string(xl_status)
+                LOG.debug(err_string.value)
```
christiansandberg: This might be a serious issue and should be raised as an exception.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+            else:
+                ev_string = vxlapi.get_event_string(event_list)
```
christiansandberg: Is it really necessary to parse the string to get CAN data? Isn't that information already available in ``event_list``?

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+                        rx_msg = Message(
+                            timestamp=long(can_msg.group('timestamp')),
+                            is_error_frame=True,
+                            arbitration_id=long(can_msg.group('type')))
+                        return rx_msg
+                    else:
+                        return ev_string.value
+                elif event_list.tag == vxlapi.XL_CHIP_STATE.value:
+                    can_msg = CHIP_STATE_RE(ev_string.value)
+                    return ev_string.value
```
christiansandberg: Returning anything other than can.Message will likely break implementations using this.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+        message_count = ctypes.c_uint(1)
+        xl_event = vxlapi.XLevent()
+        xl_event.tag = ctypes.c_ubyte(vxlapi.XL_TRANSMIT_MSG.value)
+        xl_event.timeStamp = vxlapi.XLuint64(msg.timestamp)
+        xl_event.tagData.msg.id = ctypes.c_ulong(msg.arbitration_id)
+        xl_event.tagData.msg.dlc = ctypes.c_ushort(msg.dlc)
+        xl_event.tagData.msg.flags = ctypes.c_ushort(0)
+        for idx in range(0, msg.dlc):
+            xl_event.tagData.msg.data[idx] = ctypes.c_ubyte(msg.data[idx])
+        vxlapi.can_transmit(self.port_handle, self.mask, message_count,
```
christiansandberg: Check for errors and raise exception.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+        if xl_status.value != vxlapi.XL_ERR_QUEUE_IS_EMPTY:
+            if xl_status.value != vxlapi.XL_SUCCESS:
+                err_string = vxlapi.get_error_string(xl_status)
+                LOG.debug(err_string.value)
+            else:
+                ev_string = vxlapi.get_event_string(event_list)
+                if event_list.tag == vxlapi.XL_RECEIVE_MSG.value:
+                    can_msg = RX_MSG_RE(ev_string.value)
+                    if can_msg.group('id') is not None:
+                        rx_msg = Message(
```
christiansandberg: What about extended IDs?

JulienGrv:
Thank you @christiansandberg for the review, I will improve the code as soon as I have time for that.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+                            timestamp=long(can_msg.group('timestamp')),
```
christiansandberg: ``timestamp`` must be in units of seconds, i.e. divide with 1000000000.0. Preferably it should also be a Unix timestamp but since the driver does not seem to support absolute timestamps it gets a bit complicated. If you want to you can look at the kvaser implementation.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+                    return ev_string.value
+                else:
+                    return ev_string.value
+        return None
+
+    def send(self, msg):
+        message_count = ctypes.c_uint(1)
+        xl_event = vxlapi.XLevent()
+        xl_event.tag = ctypes.c_ubyte(vxlapi.XL_TRANSMIT_MSG.value)
+        xl_event.timeStamp = vxlapi.XLuint64(long(msg.timestamp))
```
christiansandberg: Don't think you need to set timestamp when sending messages.

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,157 @@
...
+).match
+CHIP_STATE_RE = re.compile(
+    r'^CHIP_STATE c=(?P<channel>\d+), t=(?P<timestamp>\d+), busStatus=\s?(?P<busStatus>\w+), txErrCnt=(?P<txErrCnt>\d+), rxErrCnt=(?P<rxErrCnt>\d+)$'
+).match
+
+
+class VectorBus(BusABC):
+    """The CAN Bus implemented for the Vector interface."""
+
+    def __init__(self, channel, can_filters=None, **config):
```
christiansandberg: It would be nice to support ``can_filters`` also since there is the ``xlCanSetChannelAcceptance()`` function. Since the HW only supports one filter you could ignore it if more than one is given.

hardbyte:
Sorry you'll now also need to merge in my recent api changes which moved `can/interfaces/interface.py` -> `can/interface.py`.

JulienGrv:
Many changes happened since last month. I will try to integrate all the changes as soon as I can but I wonder if it would not be better to simply close this PR and submit a new one when I am finished. What do you think?

hardbyte:
@JulienGrv are you able to address the feedback left by @christiansandberg in the review?

mgiaco:
@JulienGrv  are there any news on your pull for vetor?
many thanks
cheers mathias

christiansandberg:
I have a branch on my fork where I've tried to fix the remaining things. I should be able to test it in the coming weeks.

JulienGrv:
Hi everyone,
Sorry for the very very late feedback... I quitted at the beginning of March the company where I had this Vector CANCase XL device, so I could not finished the wrapper. Besides, I lost interest for this side project.

christiansandberg:
Final touches will be made in #182. Thank you Julien for doing this. I know supporting Vector interfaces is going to be appreciated by many.

Issue #99, opened 2016-11-30
Refactor CAN.py


hardbyte:
- [ ] move the listeners into own file
- [ ] move logging function into `util.py`

christiansandberg:
My idea is to move the log readers and writers to their own modules. Suggestion:

can.log.sqlite
can.log.asc
can.log.csv
etc

hardbyte:
Agree, this sounds sensible. Perhaps `can.io` instead of `can.log` though.

PR #115, opened 2017-01-05
BLF logging support

### LINKED ISSUE #108: BLF logging support
I did some investigation on the [BLF](http://kb.vector.com/entry/520/) format and thought I might share my findings here in case I or someone else decides to give it a go in the future. The format is quite popular due to its compact footprint. On the other hand there is also the [MDF](https://wiki.asam.net/display/STANDARDS/ASAM+MDF) format which is now an official industry standard but seems even more complex and the spec seems only available to members.

BLF is invented by Vector Informatik GmbH, and as with the ASCII-format, no specification is easily available. It could just be a matter of asking them though. [BUSMASTER](https://github.com/rbei-etas/busmaster/tree/master/Sources/BUSMASTER/Format%20Converter/BlfLibrary/Src/Kernel) has implemented parsing support for it so I have looked at their C++ implementation.

The file starts with a 144 byte header. It contains some info that is not known until the log has stopped so it will have to be written last. The rest is one or more "log containers" which consists of a header and some zlib compressed data, usually up to 8 kB of uncompressed data each. This data contains the actual CAN messages.

Seems certainly doable to implement but might require some trial and error.

christiansandberg:
Adds both read and write support for BLF (Binary Logging File).
Fixes #108.

It is also a start for #99 where it has been placed under a new can.io package in order to keep CAN.py from being cluttered.


test/listener_test.py
```diff
@@ -12,6 +12,22 @@
 logging.getLogger("").setLevel(logging.DEBUG)
 
 
+# List of messages of different types that can be used in tests
+TEST_MESSAGES = [
```
hardbyte: Nice, we should have had this for a very long time! More specific to BLF though perhaps add a data directory and put an externally created blf file. Eg the test file from the C++ repo: https://bitbucket.org/tobylorenz/vector_blf/raw/3039a27a2cb24bfafb971103e4d61944c4883ca9/src/Vector/BLF/tests/unittests/data/logfile.blf

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+            systemtime[0], systemtime[1], systemtime[3],
+            systemtime[4], systemtime[5], systemtime[6], systemtime[7] * 1000)
+        return time.mktime(t.timetuple()) + systemtime[7] / 1000.0
+    except ValueError:
+        return 0
+
+
+class BLFReader(object):
+    """
+    Iterator of CAN messages from a Binary Logging File.
```
hardbyte: Probably worth mentioning that any non CAN message objects are dropped when encountered.

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+import struct
+import zlib
+import datetime
+import time
+
+from can.message import Message
+from can.CAN import Listener
+
+
+logger = logging.getLogger(__name__)
```
hardbyte: While you started with the best of intentions, this logger isn't used.

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+Implements support for BLF (Binary Logging Format) which is a proprietary
+CAN log format from Vector Informatik GmbH.
+
+No offical specification is available. The implementation is based on the works
+by Toby Lorenz (https://bitbucket.org/tobylorenz/vector_blf).
+The file starts with a header. The rest is one or more "log containers"
+which consists of a header and some zlib compressed data, usually up to 128 kB
+of uncompressed data each. This data contains the actual CAN messages and other
+objects types.
+"""
```
hardbyte: Typo in *official*. Suggest we add a comment about the license of the Vector BLF library too. Something like:

> No official specification of the binary logging format is available. This implementation is based on Toby Lorenz' C++ library "Vector BLF" which is licenced under GPLv3. https://bitbucket.org/tobylorenz/vector_blf

hardbyte: Actually on a similar note can you add a line to the `doc/history.rst#acknowledgements` to acknowledge both Toby Lorenz and your work for this.

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+        else:
+            obj_size = OBJ_HEADER_STRUCT.size + CAN_ERROR_STRUCT.size
+            header = OBJ_HEADER_STRUCT.pack(
+                b"LOBJ", OBJ_HEADER_STRUCT.size, 1, obj_size, CAN_ERROR,
+                2, 0, timestamp)
+            data = CAN_ERROR_STRUCT.pack(self.channel, 0)
+        self._add_data(header + data)
+
+    def log_event(self, message, timestamp=None):
+        """Add an arbitrary message to the log file."""
```
hardbyte: Since we already deal in "messages" I'd change this parameter to `text`. As it is a public api could you add the expected types to the docstring too.

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+        if timestamp is None:
+            timestamp = time.time()
+        if self.start_timestamp is None:
+            self.start_timestamp = timestamp
+        self.stop_timestamp = timestamp
+        try:
+            # Only works on Windows
+            text = message.encode("mbcs")
```
hardbyte: Ehh I don't even want to know

can/io/blf.py
```diff
@@ -0,0 +1,263 @@
...
+        except LookupError:
+            text = message.encode("ascii")
+        timestamp = int((timestamp - self.start_timestamp) * 1000000000)
+        comment = b"Added by python-can"
```
hardbyte: Do most other blf tools add a comment like this?

test/listener_test.py
```diff
@@ -161,5 +178,23 @@ def test_sql_reader(self):
         self.assertEqual(len(ms), 1)
         self.assertEqual(0xDADADA, ms[0].arbitration_id)
 
+
+class BLFTest(unittest.TestCase):
+
+    def test_reader_writer(self):
+        writer = can.BLFWriter("testlog.blf")
```
hardbyte: Use a `tempfile.NamedTemporaryFile` here.

test/listener_test.py
```diff
@@ -161,5 +178,23 @@ def test_sql_reader(self):
...
+        for msg in TEST_MESSAGES:
+            writer(msg)
+        writer.log_event("Log stops here", TEST_MESSAGES[-1].timestamp + 1)
+        writer.log_event("Another comment", TEST_MESSAGES[-1].timestamp + 2)
```
hardbyte: One of these should be without a timestamp.

can/io/blf.py
christiansandberg: I know, but according to Toby it should be encoded in MBCS but it is only available in Python for Windows... 💩

can/io/blf.py
christiansandberg: Don't know, usually I think it is added manually.

can/io/blf.py
rahulmohang: Hi, 

If this implementation is based on Toby Lorenz' C++ library "Vector BLF" which is licensed under GPLv3, do you think that there is a chance that the license of this file (if this file is licensed under GPL v3.0 too) could affect licensing of other files too and make the whole component GPL v3.0?

christiansandberg: It's just used as a reference for the format, none of the code is used. So I would say no if I understand the license correctly.

rahulmohang: Ok

Issue #25, opened 2015-04-20
can_filters should be backported to socketcan_ctypes


hardbyte:
Originally reported by: **Michael Muldoon (Bitbucket: [mbmuldoon07](https://bitbucket.org/mbmuldoon07), GitHub: [mbmuldoon07](https://github.com/mbmuldoon07))**

---

The can filters does not appear to work at least with socketcan_ctypes.  I've tried anything I can think of when adding the can_filter to the bus but no matter what there appears to be no filtering.

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/25

felixdivo:
Isn't this fixed in #116?

PR #116, opened 2017-01-06
Update socketcan_ctypes.py

mdjarrel:
Added filter support based on socketcan_native implementation and comments in issues #25.

### DIFF
--- a/can/interfaces/socketcan/socketcan_constants.py
+++ b/can/interfaces/socketcan/socketcan_constants.py
@@ -81,3 +81,10 @@ MAX_11_BIT_ID = STD_ACCEPTANCE_MASK_ALL_BITS
 
 EXT_ACCEPTANCE_MASK_ALL_BITS = (2**29 - 1)
 MAX_29_BIT_ID = EXT_ACCEPTANCE_MASK_ALL_BITS
+
+def parseCanFilters(can_filters = {}):
+  can_filter_fmt = "={}I".format(2 * len(can_filters))
+  filter_data = []
+  for can_filter in can_filters:
+    filter_data.append(can_filter['can_id'])
+    filter_data.append(can_filter['can_mask'])
--- a/can/interfaces/socketcan/socketcan_constants.py
+++ b/can/interfaces/socketcan/socketcan_constants.py
@@ -81,18 +81,3 @@ MAX_11_BIT_ID = STD_ACCEPTANCE_MASK_ALL_BITS
 
 EXT_ACCEPTANCE_MASK_ALL_BITS = (2**29 - 1)
 MAX_29_BIT_ID = EXT_ACCEPTANCE_MASK_ALL_BITS
-
-def parseCanFilters(can_filters=None):
-  if can_filters is None:
-    # Pass all messages
-    can_filters=[{
-      'can_id': 0,
-      'can_mask': 0
-    }]
-  
-  can_filter_fmt = "={}I".format(2 * len(can_filters))
-  filter_data = []
-  for can_filter in can_filters:
-    filter_data.append(can_filter['can_id'])
-    filter_data.append(can_filter['can_mask'])
-  return can_filter_fmt, filter_data
--- a/can/interfaces/socketcan/socketcan_constants.py
+++ b/can/interfaces/socketcan/socketcan_constants.py
@@ -82,9 +82,17 @@ MAX_11_BIT_ID = STD_ACCEPTANCE_MASK_ALL_BITS
 EXT_ACCEPTANCE_MASK_ALL_BITS = (2**29 - 1)
 MAX_29_BIT_ID = EXT_ACCEPTANCE_MASK_ALL_BITS
 
-def parseCanFilters(can_filters = {}):
+def parseCanFilters(can_filters=None):
+  if can_filters is None:
+    # Pass all messages
+    can_filters=[{
+      'can_id': 0,
+      'can_mask': 0
+    }]
+  
   can_filter_fmt = "={}I".format(2 * len(can_filters))
   filter_data = []
   for can_filter in can_filters:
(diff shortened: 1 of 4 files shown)


can/interfaces/socketcan/socketcan_ctypes.py
```diff
@@ -50,13 +51,56 @@ def __init__(self,
...
+        """
+        if can_filters is not None and len(can_filters) > 0:
+            can_filter_fmt = "={}I".format(2 * len(can_filters))
+            filter_data = []
+            for can_filter in can_filters:
+                filter_data.append(can_filter['can_id'])
+                filter_data.append(can_filter['can_mask'])
+            res = libc.setsockopt(self.socket, SOL_CAN_RAW,
+                CAN_RAW_FILTER, struct.pack(can_filter_fmt, *filter_data),
+                4*2*len(kwargs['can_filters'])
```
hardbyte: `kwargs` isn't defined in this method, should just be `can_filters`

can/interfaces/socketcan/socketcan_ctypes.py
```diff
@@ -50,13 +51,56 @@ def __init__(self,
         self.socket = createSocket()
 
         log.debug("Result of createSocket was %d", self.socket)
+        
+        # Add any socket options such as can frame filters
+        if 'can_filters' in kwargs and len(kwargs['can_filters']) > 0:
+            log.debug("Creating a filtered can bus")
+            can_filter_fmt = "={}I".format(2 * len(kwargs['can_filters']))
+            filter_data = []
```
hardbyte: We can just call the `set_filters` method here and remove the duplicated code.

can/interfaces/socketcan/socketcan_ctypes.py
```diff
@@ -50,13 +51,56 @@ def __init__(self,
         self.socket = createSocket()
 
         log.debug("Result of createSocket was %d", self.socket)
+        
+        # Add any socket options such as can frame filters
+        if 'can_filters' in kwargs and len(kwargs['can_filters']) > 0:
+            log.debug("Creating a filtered can bus")
+            can_filter_fmt = "={}I".format(2 * len(kwargs['can_filters']))
```
hardbyte: I'm tempted to say this should be moved into `socketcan_constants.py` as it is shared with the native interface.

can/interfaces/socketcan/socketcan_ctypes.py
hardbyte: Replace the magic `4*2` numbers with a variable

can/interfaces/socketcan/socketcan_ctypes.py
mdjarrel: Agreed

mdjarrel:
Thank you for the review @hardbyte . I have moved the code to convert the filter dictionary array to a sockopt arg array into socketcan_constants.py. I have also updated both the ctypes and native implementations to use the new function in socketcan_constants.py.
I also moved the code to take a None argument for can_filters and populate a wide open filter to the same function as the conversion code.
I am using the set_filters function now for setting filters in the constructor in socketcan_ctypes.py and fixed the reference to kwargs as well.
I also removed the magic 4*2 number and replaced it with a call to ctypes.sizeof(ctypes.c_uint32) and the length of the converted filter array instead of the filter dictionary array.
I changed my print call when setting the filters to use log.debug as well.
I quickly tested this with a vcan interface on Ubuntu in both python 2.7.12 and python 3.4.5 with ctypes in 2.7 and native in 3.4 and both seem to work.
This should be ready for another review.

can/interfaces/socketcan/socketcan_constants.py
```diff
@@ -81,3 +81,18 @@
...
+      'can_id': 0,
+      'can_mask': 0
+    }]
+  
+  can_filter_fmt = "={}I".format(2 * len(can_filters))
+  filter_data = []
+  for can_filter in can_filters:
+    filter_data.append(can_filter['can_id'])
+    filter_data.append(can_filter['can_mask'])
+  return can_filter_fmt, filter_data
```
hardbyte: I think it would be better just return `struct.pack(can_filter_fmt, *filter_data)` - or am I missing something?

Slightly uncomfortable adding code to a constants file, perhaps add this in a new `socketcan_common.py` module.

can/interfaces/socketcan/socketcan_ctypes.py
```diff
@@ -88,18 +77,13 @@ def set_filters(self, can_filters=None):
...
-                )
-            if res != 0:
-                print(str(res))
+        can_filter_fmt, filter_data = parseCanFilters(can_filters)
+        res = libc.setsockopt(self.socket, SOL_CAN_RAW,
+                              CAN_RAW_FILTER, struct.pack(can_filter_fmt, *filter_data),
+                              len(filter_data)*ctypes.sizeof(ctypes.c_uint32)
+                             )
+        if res != 0:
+            log.debug('Setting filters failed: ' + str(res))
```
hardbyte: This should be a much higher priority log message (e.g. `log.error`).

I'm tempted to say we should also raise a `CanError` exception. Have you seen it fail, do you know when it might?

mdjarrel: I have not seen it fail and I am not sure when it might.

mdjarrel:
I moved parseCanFilters out of socketcan_constants.py and into socketcan_common.py.
I am also now returning the filter struct from parseCanFilters.
Updated optlen param in socketcan_ctypes setsockopt to be len(filter_struct) .
I elevated the log.debug to log.error. I made a note to revisit the possibility of raising a CanError exception.

hardbyte:
Thanks for the contribution!

PR #117, opened 2017-01-08
Refactoring all functionality out of CAN.py

hardbyte:
Fairly minor changes moving functions out of CAN.py into more sensible locations.

Do the module names make sense?
I thought it best to keep `CAN.py` incase users of this library are using that as the primary api... thoughts?


can/util.py
```diff
@@ -188,6 +188,19 @@ def __iter__(self):
             yield m
 
 
+def set_logging_level(level_name=None):
+    """Set the logging level for python-can.
+    Expects one of: 'critical', 'error', 'warning', 'info', 'debug', 'subdebug'
+    """
+    try:
+        log.setLevel(getattr(logging, level_name.upper()))
```
christiansandberg: Will this work? Previously the log level was set for the 'can' logger so that it was the default for all logger under that package. Now it is set for 'can.util' so I don't think it will have much effect.

can/io/sql.py
```diff
@@ -0,0 +1,127 @@
+from can.listener import BufferedReader
+from can.message import Message
+
+import sys
+import time
+import threading
+import sqlite3
+import logging
+
+log = logging.getLogger('can.io.sql')
```
christiansandberg: Just in case someone would create a writer for MySQL or PostreSQL or whatever, maybe this module should be named sqlite?

hardbyte:
Thanks for the code review, hopefully I've addressed your points?

PR #118, opened 2017-01-13
Added linux support for pcanbasic

DPastl:
Tried using python-can in linux but the PCANBasic file appears to be the old windows version from Peak-systems.  The updated file includes a platform check that allows one PCANBasic.py file to be used across both OS's.

### DIFF
--- a/can/interfaces/pcan/PCANBasic.py
+++ b/can/interfaces/pcan/PCANBasic.py
@@ -3,6 +3,7 @@
 #  more Info at http://www.peak-system.com
 
 import logging
+import platform
 
 from can.interfaces.pcan.pcan_constants import *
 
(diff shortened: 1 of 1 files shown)


DPastl:
It would also be great if this could get updated on the master branch as well!

hardbyte:
Thanks David. I was under the impression that peak systems now support the linux socketcan interface? What advantage would you have from manually linking to their library?
http://www.peak-system.com/fileadmin/media/linux/index.htm

That aside, I have no problem with your changes. This will be included in master when we are ready for the [2.0 release](https://github.com/hardbyte/python-can/milestone/1) - still have a few bits of work outstanding .

DPastl:
Hello Brian,

I believe you are correct about socketcan, although I've never used it. I generally use the python api they provide, which your library was attempting to make use of on my system.  Adding those lines allowed it to find the required file (libpcanbasic.so).  I'm not sure how python-can uses the pcan (this is my first time looking at it), but this was the fix I needed to get going, and something I've had to do before in my own code base to provide cross platform support.

Cheers,

PR #124, opened 2017-01-23
Adding support for ICS (INTREPID CONTROL SYSTEMS) neoVI interface via pyneovi

pierreluctg:


pierreluctg:
@hardbyte please review

can/interfaces/neovi_api/neovi_api.py
```diff
@@ -0,0 +1,172 @@
...
+        return msg
+
+    def send(self, msg):
+        data = tuple(msg.data)
+        ics_msg = icsSpyMessage()
+        ics_msg.ArbIDOrHeader = msg.arbitration_id
+        ics_msg.NumberBytesData = len(data)
+        ics_msg.Data = data
+        ics_msg.StatusBitField = 0
+        ics_msg.StatusBitField2 = 0
```
christiansandberg: These should reflect `msg.is_extended_id` and `msg.is_remote_request` like in recv().

can/interfaces/neovi_api/neovi_api.py
```diff
@@ -0,0 +1,172 @@
...
+    }
+    return names.get(device_type, 'Unknown neoVI')
+
+
+class NeoVIBus(BusABC):
+    """
+    The CAN Bus implemented for the pyneovi interface.
+    """
+
+    def __init__(self, channel=None, can_filters=None, **config):
```
christiansandberg: I think I saw that you can set the bitrate of the device as well.

can/interfaces/neovi_api/neovi_api.py
```diff
@@ -0,0 +1,172 @@
...
+    from queue import Queue, Empty
+except ImportError:
+    from Queue import Queue, Empty
+
+try:
+    from neovi import neodevice
+    from neovi import neovi
+    from neovi.structures import icsSpyMessage
+except ImportError:
+    print
```
christiansandberg: Use `logger.error()`.

can/interfaces/neovi_api/neovi_api.py
```diff
@@ -0,0 +1,172 @@
...
+            if not (arb_id ^ can_filter['can_id']) & can_filter['can_mask']:
+                return True
+
+        logging.info("%s not matching" % arb_id)
+        return False
+
+    def recv(self, timeout=None):
+        msg = None
+        try:
+            while msg is None:
```
christiansandberg: There is a risk to get stuck here depending on filter configuration. You could skip the while loop and just return `None` if the message didn't match the filter because the calling function will probably call `recv()` again anyway.

can/interfaces/neovi_api/neovi_api.py
pierreluctg: Can we leave that for a future improvement for now? I do not have a ms can bus to test at this point.

christiansandberg: Sure, as long as there are other ways to configure the bitrate.

pierreluctg: Yes there is. This configuration is saved on the device. So one can use the small configuration (neoVI Explorer) application that is included with the driver from ICS.

christiansandberg:
Could you also add a chapter in doc/installation.rst and shortly describe how to install pyneovi?

Issue #125, opened 2017-01-24
Unable to get correct status after Pcan bus initialization


Delgan:
Hi.

First of all, I must thank you for this library and all the work you have put into it. This is of immeasurable help to me.

I am working with a Pcan-USB from Peak System and I encountered trouble trying to initialize a simple bus controller. I am using Python 3.5 on Linux and python-can current development branch (2.0.0.a1):

```python
>>> from can.interfaces.pcan import PcanBus
>>> bus = PcanBus("PCAN_USBBUS1")
>>> bus.status()
67108864
```

This corresponds to `PCAN_ERROR_ILLPARAMVAL`.

I may have found the cause of this error, so I will open a PR too.

PR #126, opened 2017-01-24
Fix wrong channel value used to call PCANBasic API (#125)

Delgan:
It seems there was confusion with the attributes `channel_info` and `m_PcanHandle` used in the pcan interface. So, I replaced each call to the PCANBasic object using `channel_info` with `m_PcanHandle`.

Once applied, the `.status()` method returns no error:

```python
>>> from can.interfaces.pcan import PcanBus
>>> bus = PcanBus("PCAN_USBBUS1")
>>> bus.status()
0
```

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -125,7 +125,7 @@ class PcanBus(BusABC):
 
         :return: The status code. See values in pcan_constants.py
         """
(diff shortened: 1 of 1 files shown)


hardbyte:
Cheers Delgan. 

@christiansandberg feel free to merge if/when you're happy with a PR

PR #127, opened 2017-01-24
Add timeout argument to send()

christiansandberg:
Today, all interfaces except Kvaser queue messages to be sent, but never block. Kvaser, IXXAT and USB2CAN have support for blocking until the message has been ACK:ed. This behaviour could sometimes be desirable since it gives you an indication that something is wrong with the CAN bus or you are using the wrong bitrate.

This proposal adds a timeout argument to the `send()` method which _can_ be used by the interface to do a blocking transmission. Since a timeout of 0 makes no sense here as opposed to the `recv()` method, I suggest to only block if timeout is > 0, otherwise it should be queued as normal which should be the default.

Unfortunately I can only test this on Kvaser.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -56,8 +56,9 @@ class BusABC(object):
 
         :param msg: A :class:`can.Message` object.
         :param float timeout:
-            If given, wait for message to be ACK:ed.
-            Might not be supported by the interface.
+            If > 0, wait up to this many seconds for message to be ACK:ed.
+            If timeout is exceeded, an exception will be raised.
+            Might not be supported by all interfaces.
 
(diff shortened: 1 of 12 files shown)


cowo78:
Sounds good to me. I will test IXXAT stuff. A question, mostly unrelated, I admit. What is the flush_tx_buffer() method supposed to do? Send out every message in the output mailboxes or just wipe them and mark them clean? If the answer is to send out every message this should need a timeout too.

christiansandberg:
flush_tx_buffer() as I understand it should wipe the transmit buffer. So I don't think it should need a timeout.

PR #130, opened 2017-01-25
 Update PCANBasic to v4.1.0 and refactor its import (fixes #129)

### LINKED ISSUE #129: PCAN Messages are incorrectly formed on Linux 64bits
Hello.

The messages received using a Peak System PCAN-USB make no sense on my computer, and are completely different from what is sent.

```python
# Sender (armv7l)
>>> from can.interfaces.pcan import PcanBus
>>> from can import Message
>>>
>>> bus = PcanBus("PCAN_USBBUS1")
>>> msg = Message(data=[8, 7, 6, 5, 4, 3, 2, 1])
>>> bus.send(msg)
```
```python
# Receiver (x86_64)
>>> from can.interfaces.pcan import PcanBus
>>>
>>> bus = PcanBus("PCAN_USBBUS1")
>>> _ = bus.recv()  # Just some flushing
>>> bus.recv()
can.Message(timestamp=131448817700072.39, is_remote_frame=False, extended_id=True, is_error_frame=False, arbitration_id=0x708080200000000, dlc=5, data=[0x4, 0x3, 0x2, 0x1, 0x0])
```

Using the example provided by Peak System worked just fine. So, I made some tests and it appears that the PCANBasic API of python-can would deserve an update. I will open a Pull Request.

Delgan:
The version of PCANBasic API used by python-can seems quite old. Since then, Peak System updated the Python API and fixed several bugs.

* [Windows Version History](http://www.peak-system.com/PCAN-Basic.126.0.html)
* [Linux Version History](http://www.peak-system.com/PCAN-Basic-Linux.433.0.html)

One of the changes which fixes #129 is:
> 2013-11-12 - Version 2.0.1
> _Changed all "c_ulong" types to "c_uint" to ensure 32-bits implementation whatever Linux arch data model is used_

I took advantage of this pull request to entirely update the PCANBasic API and use the last published version (v4.1.0.2). I guess it is a good time to do that, since python-can is updating to v2.0.0. This is probably related to [this hardbyte's comment](https://github.com/hardbyte/python-can/pull/118#issuecomment-272607848).

I compared the Linux and Windows API, [they are slightly the same](https://www.diffchecker.com/wH0HkUEE).

Then, what I did:
* Separate constants from the API code as you implemented it
* Fix `TPCANBitrateFD()` calls in `pcan_constants.py` to be compatible with Python 3 (in Python 2 `str == bytes` but this is no longer the case in Python 3)
* Replace all `print()` with `log.error()`
* Fix a typo in API (replace `touple` with `tuple`)
* Remove some comments in the API which was redundant with docstrings
* Format some code to conform to the PEP8 guide
* Update metadata header


christiansandberg:
I think this may be a good opportunity to use the original PCANBasic.py directly without modifications. Then it will be easier to upgrade it in the future.

Delgan:
I agree with you. Although there will still need an interface file between `pcan.py` and `PCANBasic.py` to monkey-patch `print()` and `TPCANBitrateFD()`. Or maybe directly propose the changes to a Peak System software engineer. I do not know which solution is better.

christiansandberg:
I would say that changing print() to logger.error() doesn't matter much because you will get an exception anyway. Also we don't support CAN-FD so TPCANBitrateFD() won't be needed. If we need to do some patching to make it work that's OK, but it would be nice to keep it to a minimum.

Delgan:
@christiansandberg So, it appears that Peak-System developpers [are aware of the incompatiblity with Python 3](https://www.peak-system.com/forum/viewtopic.php?f=41&t=1930#p5573) but did not fix it, so contacting them about this would be pointless.

I implemented a solution using an interface file. I tried to keep it as simple and readable as possible, but mocking the import was not that easy.
This is probably a little bit overkill compared to just manually delete the `TPCANBitrateFD` lines but at least the changes are now explicit and may avoid mistakes and confusions if someone want to update the module in the future.

christiansandberg:
I also think it is bit overkill with the interface file. Patching their code is fine, but maybe just add a comment where patches has been made, then it will be easy to spot it in the diff when someone updates the file in the future.

Delgan:
All right, it is done. I do not expect PCANBasic to change a lot in the future so it is not too much annoying to patch it manually.

Issue #120, opened 2017-01-23
Which socketcan interface to use not correctly identified


remusmp:
Hi,

I've installed the latest python-can package this way:

`pip3 install git+https://github.com/hardbyte/python-can.git`

In pip list it appears as 2.0.0a1.

Then I typed in these commands in a python terminal:
```
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import can
>>> can.rc['interface'] = 'socketcan'
>>> bus = can.interface.Bus()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/interface.py", line 79, in __new__
    raise NotImplementedError("CAN interface '{}' not supported".format(interface))
NotImplementedError: CAN interface 'socketcan' not supported
```

I'm able to see the CAN bus under Linux via `candump` (I have installed all necessary modules). Here is an example:
```
candump slcan0
  slcan0  001   [4]  1B 00 00 00
  slcan0  002   [2]  38 13
  slcan0  001   [4]  1B 00 00 00
```

hardbyte:
Thanks for the report - I'll have a look into it. Would you mind trying with the value `socketcan_native`?

remusmp:
This works:
```
import can
can.rc['interface'] = 'socketcan_native'
can.rc['channel'] = 'slcan0'
bus = can.interface.Bus()
```

Once again thanks!

PR #134, opened 2017-01-27
Automatically choose correct socketcan interface

hardbyte:
Note I also moved `can.interfaces.interface` to `can.interface` as it seems to make more sense there.

Fix for #120


can/interface.py
```diff
@@ -27,21 +23,12 @@ def __new__(cls, other, channel=None, *args, **kwargs):
             can.rc['interface'] = kwargs['bustype']
```
christiansandberg: I've been thinking about this. Wouldn't be more logical to never change can.rc if a custom bustype has been given? One might think that the default interface and channel should always be used when nothing else has been specified when initiating the Bus, but now it depends on the last created bus.

hardbyte: Yes - I agree in principal. A read only global config variable would be an improvement. I'll take a look and merge in the 'neovi' + io changes.

christiansandberg: It doesn't have to be read only (one might still want to change it programmatically), just that creating a bus different from the default should not modify the default.

hardbyte: Yup I meant the library should treat it as read only - the user may of course change at will.

hardbyte:
I'd appreciate it if you could cast your eyes over those changes too

christiansandberg:
I think it looks good!

PR #135, opened 2017-01-30
Avoid using polling for PCAN

### LINKED ISSUE #53: PCAN non-polling recv timeouts
Originally reported by: **Brian Thorne (Bitbucket: [hardbyte](https://bitbucket.org/hardbyte), GitHub: [hardbyte](https://github.com/hardbyte))**

---

Implementing a non-polling `recv` function for the PCAN backend using `PCAN_RECEIVE_EVENT`.

See attached PCAN documentation.

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/53

christiansandberg:
Utilize events instead of polling when the [pywin32](https://sourceforge.net/projects/pywin32) or [pypiwin32](https://pypi.python.org/pypi/pypiwin32) package is available.

Fixes #53.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -154,9 +154,15 @@ class PcanBus(BusABC):
         return status == PCAN_ERROR_OK
 
     def recv(self, timeout=None):
-        start_time = timeout_clock()
-
-        rx_msg = Message()
+        if win32event:
+            # We will utilize events for the timeout handling
+            timeout_ms = int(timeout * 1000) if timeout is not None else win32event.INFINITE
+        elif timeout is not None:
+            # Calculate max time
+            end_time = timeout_clock() + timeout
+        else:
+            # Skip timeout handling
+            end_time = 0
 
         log.debug("Trying to read a msg")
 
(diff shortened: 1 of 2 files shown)


can/interfaces/pcan/pcan.py
```diff
@@ -27,8 +32,8 @@
     timeout_clock = time.clock
 
 # Set up logging
-logging.basicConfig(level=logging.WARNING)
 log = logging.getLogger('can.pcan')
+log.setLevel(logging.WARNING)
```
hardbyte: IMO we shouldn't be setting the logging level at all outside of our scripts. So :+1: for removing the `basicConfig` but I'd say also remove the `setLevel` call.

can/interfaces/pcan/pcan.py
```diff
@@ -83,6 +88,13 @@ def __init__(self, channel, *args, **kwargs):
         if result != PCAN_ERROR_OK:
             raise PcanError(self._get_formatted_error(result))
 
+        if win32event:
+            self._recv_event = win32event.CreateEvent(None, 0, 0, None)
+            result = self.m_objPCANBasic.SetValue(
```
hardbyte: Just confirming that the pcan api dictates that this must be a win32event? We can't pass in a more standard Python type such as `threading.Event`?

can/interfaces/pcan/pcan.py
```diff
@@ -83,6 +88,13 @@ def __init__(self, channel, *args, **kwargs):
         if result != PCAN_ERROR_OK:
             raise PcanError(self._get_formatted_error(result))
 
+        if win32event:
```
hardbyte: nit: `if win32event is not None:`

can/interfaces/pcan/pcan.py
christiansandberg: Unfortunately this is a Windows API thing. Would be nice if Peak could provide a more general way like a blocking function or callbacks like many other interfaces have. pywin32 is quite common on Windows installations but I should definitely add it to the installation instructions.

PR #137, opened 2017-02-01
Minor aesthetic changes

cowo78:
- Use a BACKENDS dict to describe implemented backends, substitute the if's tree in Bus.__init__
- Try to import nican only on win32 systems
- kvaser: dlsym() only if dlopen() did not fail

### DIFF
--- a/can/interface.py
+++ b/can/interface.py
@@ -1,7 +1,23 @@
 import can
+import importlib
+
 from can.broadcastmanager import CyclicSendTaskABC, MultiRateCyclicSendTaskABC
 from can.util import load_config
 
+# interface_name => (module, classname)
+BACKENDS = {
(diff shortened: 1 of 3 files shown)


can/interface.py
```diff
@@ -36,42 +52,26 @@ def __new__(cls, other, channel=None, *args, **kwargs):
...
-            cls = RemoteBus
-        elif interface == 'virtual':
-            from can.interfaces.virtual import VirtualBus
-            cls = VirtualBus
-        elif interface == 'neovi':
-            from can.interfaces.neovi_api import NeoVIBus
-            cls = NeoVIBus
-        else:
+        try:
+            (module_name, class_name) = BACKENDS[interface]
```
hardbyte: :+1: I was cringing looking at this last night.

PR #138, opened 2017-02-01
Now using neovi.GetTimeStampForMsg to get correct time stamp

pierreluctg:
Now using neovi.GetTimeStampForMsg since it is now working after commit https://bitbucket.org/Kemp_J/pyneovi/commits/0804ce1c35aef184e32ce063ab5481d086dfde11 in pyneovi.

Remove time scaling constant since they are no longer used.
Removed win32 check for neovi import since it's not win32 only.

### DIFF
--- a/can/interfaces/neovi_api/neovi_api.py
+++ b/can/interfaces/neovi_api/neovi_api.py
@@ -11,26 +11,19 @@ Implementation references:
 
 import logging
 
-import sys
-
 logger = logging.getLogger(__name__)
 
 try:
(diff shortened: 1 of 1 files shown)


christiansandberg:
Looks like GetTimeStampForMsg takes two arguments and returns a tuple of result and timestamp. Or is there some other change I missed?

pierreluctg:
@christiansandberg you are right, I have pushed the wrong code. Fixed it now.

PR #140, opened 2017-02-07
Fix listener test

hardbyte:
Instead of waiting an arbitrary amount of time, actually use the blocking queue.get timeout that we are waiting for.

### DIFF
--- a/can/io/sqlite.py
+++ b/can/io/sqlite.py
@@ -104,11 +104,12 @@ class SqliteWriter(BufferedReader):
                     m.dlc,
                     buffer(m.data)
                 ))
-                m = self.get_message(self.GET_MESSAGE_TIMEOUT)
 
                 if time.time() - last_write > self.MAX_TIME_BETWEEN_WRITES:
                     log.debug("Max timeout between writes reached")
                     break
+
+                m = self.get_message(self.GET_MESSAGE_TIMEOUT)
 
             if len(messages) > 0:
                 with self.conn:
--- a/can/io/sqlite.py
+++ b/can/io/sqlite.py
@@ -39,6 +39,24 @@ class SqliteWriter(BufferedReader):
 
     The sqlite database may already exist, otherwise it will
     be created when the first message arrives.
+
+    Messages are internally buffered and written to the SQL file in a background
+    thread.
+
+    .. note::
+
+        When the listener's :meth:`~SqliteWriter.stop` method is called the
+        thread writing to the sql file will continue to receive and internally
+        buffer messages if they continue to arrive before the
+        :attr:`~SqliteWriter.GET_MESSAGE_TIMEOUT`.
+
+        If the :attr:`~SqliteWriter.GET_MESSAGE_TIMEOUT` expires before a message
+        is received, the internal buffer is written out to the sql file.
+
+        However if the bus is still saturated with messages, the Listener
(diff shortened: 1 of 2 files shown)


christiansandberg:
I think the root issue is that when `stop_running_event` is set, the `messages` list may not be empty and we jump out of the loop and close the file. Therefore the last messages will not be saved.

christiansandberg:
And if a message is received after `MAX_TIME_BETWEEN_WRITES` has passed, it will be discarded.

hardbyte:
I'm not following you Christian. When we call `stop()` from _thread 1_ an event is set and _thread 1_ blocks on the `join()` call waiting for _thread 2_ (which is executing `db_writer_thread()`).

At this point in _thread 2_ we are *probably* waiting in the middle of the blocking `get_message` calls. This returns either a message or `None` after the timeout `GET_MESSAGE_TIMEOUT` expires. In one case we write the received message to the sql database, in the other there was no message and when we get back to the top of the outer while loop we find out that we've been told to stop running.

What I assume was happening was that the test (in _thread 1_) was finished after the smaller sleep period expired while _thread 2_ was in it for the long haul - still happily waiting to buffer another message before it would write to the sql file.

```
    def db_writer_thread(self):
        num_frames = 0
        last_write = time.time()
        self._create_db()

        while not self.stop_running_event.is_set():
            messages = []

            m = self.get_message(self.GET_MESSAGE_TIMEOUT)
            while m is not None:
                log.debug("sqlitewriter buffering message")

                messages.append((
                    m.timestamp,
                    m.arbitration_id,
                    m.id_type,
                    m.is_remote_frame,
                    m.is_error_frame,
                    m.dlc,
                    buffer(m.data)
                ))
                m = self.get_message(self.GET_MESSAGE_TIMEOUT)

                if time.time() - last_write > self.MAX_TIME_BETWEEN_WRITES:
                    log.debug("Max timeout between writes reached")
                    break

            if len(messages) > 0:
                with self.conn:
                    log.debug("Writing %s frames to db", len(messages))
                    self.conn.executemany(SqliteWriter.insert_msg_template, messages)
                    num_frames += len(messages)
                    last_write = time.time()

        self.conn.close()
        log.info("Stopped sqlite writer after writing %s messages", num_frames)

    def stop(self):
        self.stop_running_event.set()
        log.debug("Stopping sqlite writer")
        self.writer_thread.join()
```

christiansandberg:
There are some places in the code where if a message is received while `stop_running_event` has been set, it will not be written.

1. At the start, if `stop_running_event` is already set, database will be closed without writing any messages in the queue.
2. At the second `get_message`, if this is received after `MAX_TIME_BETWEEN_WRITES` has passed, it will not be added to `messages`, instead `m` will be garbage collected.
3. When writing to the database, any new messages during that time will not be checked.

My idea was to also write the rest of the messages in the queue before closing the database.

hardbyte:
Thanks for the clarification. I'm on board - this has issues I didn't notice.

My biggest concern is your second point. That second blocking `get_message` has to be the last thing in the inner loop to avoid receiving a message that it would drop.

The other points regarding shutdown behavior I'd like to think on a bit longer. I'm not convinced there is a problem with the current approach - and I don't want to introduce an issue where we continue receiving and writing messages after we've been explicitly told to stop.

Issue #147, opened 2017-03-03
Split filtering for standard and extended IDs


christiansandberg:
Right now there is no clear distinction between 11- and 29-bit messages for filtering. Seems like most interfaces have support for having different filters but right now they apply the same filter settings for both or have some other non-standard way of specifying which one to apply the filtering on.

Maybe it should be something like `[{"can_id_ext": 0x123456, "can_mask_ext": 0xFFFFFF}, {"can_id_std": 0x123, "can_mask_std": 0xFFF}]`.

If can_id and can_mask is specified, it could be automatically translated to both std and ext variants in can.interface.Bus constructor to be fairly backwards compatible.

christiansandberg:
On second thought it would be even better to add an "extended" key to the dictionary. Then it would not be necessary to update all interfaces at once.

`[{"can_id": 0x123456, "can_mask": 0xFFFFFF, "extended": True}, {"can_id": 0x123, "can_mask": 0xFFF, "extended": False}]`

christiansandberg:
I'll go ahead and close this since the pull request has been merged.

PR #153, opened 2017-04-01
Split filtering for standard and extended IDs

christiansandberg:
Allows for filtering differently on extended and standard IDs. The default is to filter on standard IDs if the "extended" key is not specified so it is fairly backwards compatible but not fully.

Need help to review the changes made to the different interfaces. @cowo78, could you check if the IXXAT changes are OK?

Update socketcan, kvaser, ixxat, nican, and remote filter configuration.
Remove SW filtering in kvaser and neovi. If it is needed at all it should be implemented in a way that all interfaces that cannot do HW filtering for some reason can use it.
Fixes issue #147.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -28,9 +28,10 @@ class BusABC(object):
             The can interface identifier. Expected type is backend dependent.
 
         :param list can_filters:
-            A list of dictionaries each containing a "can_id" and a "can_mask".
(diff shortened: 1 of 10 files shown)


cowo78:
Looks fine, but I've switched employer and I don't have access to an IXXAT setup anymore.

christiansandberg:
Created issue #157 for implementing a common software filter.

PR #155, opened 2017-04-03
changing default argument on socketcan_ctypes channel name

CatherineH:
The default value for the channel in the documentation, "channel=0", results in an error because these lines:

https://github.com/hardbyte/python-can/blob/develop/can/interfaces/socketcan/socketcan_ctypes.py#L247
https://github.com/hardbyte/python-can/blob/develop/can/interfaces/socketcan/socketcan_ctypes.py#L281

assume that the channel name is a string, not an integer. 

I suggest that this default variable be "vcan0", as suggested by the existing docstring.


hardbyte:
Thanks!

Issue #163, opened 2017-05-29
IndexError when printing messages with data of length zero


radiocane:
A can.Message created with a "data" of zero length can be correcly sent, but trying to print it will raise IndexError. I think it should be treated as a message with data=None.

radiocane:
I noticed only now that in order to raise, not only data must have length zero, but dlc must be greater than zero. Such a message (no data and dlc > 0) is allowed by the standard for Remote frames (CAN 2.0 Part A 3.1.2).

radiocane:
Fixed in 879019d3b3671be0ad007108899cec78c8f61dd5

PR #164, opened 2017-05-29
Print messages with zero-length data as if data=None. Addresses #163

radiocane:


can/message.py
```diff
@@ -65,7 +65,7 @@ def __str__(self):
 
         field_strings.append("DLC: {0:d}".format(self.dlc))
         data_strings = []
-        if self.data is not None:
+        if self.data is not None and len(self.data) > 0:
             for index in range(0, self.dlc):
```
hardbyte: Isn't the core issue that we are iterating to `dlc` instead of iterating over the data array?

radiocane: Yes. I didn't think enough before sending this PR. It'd make more sense to change "range(0, self.dlc)" to "range(0, len(self.data))". Is it ok if I close this PR and open another?

christiansandberg:
Some interfaces use all 8 bytes for the data even if DLC is less than 8 so those should be considered garbage.

hardbyte:
@radiocane you can just commit over top if you like - we can squash it into one commit when merging.

It seems like the number of elements can be given by:
```
min(len(self.data),  self.dlc)
```

The existing `data is not None` check should take care of the remote frame case.

radiocane:
I hope the new version makes more sense :)

PR #166, opened 2017-05-31
Implement receive_own_messages option for more buses

### LINKED ISSUE #160: Kvaser reads its own transmitted messages
Since the kvaser interface uses two handles by default and canlib uses local echo between handles, transmitted messages are received back. I'm not sure how other interfaces work but in many cases this is not a desirable behavior. On the other hand it might be desirable to receive messages from other programs when logging CAN traffic for example.

I suggest that local echo should be turned off when two handles are used but it should be possible to turn it back on with an argument.

christiansandberg:
Fixes #160.


christiansandberg:
I changed the solution and scope a bit. I saw that socketcan_ctypes had receive_own_messages as an option so I used that in kvaser, socketcan_native and virtual buses as well.

PR #168, opened 2017-06-20
Add asyncio interfaces

christiansandberg:
This pull request adds support for receiving (and to some extent sending) messages asynchronously using [asyncio](https://docs.python.org/3/library/asyncio.html). This can be useful when writing protocol implementations that should handle parallel communication channels with many nodes.

The default implementation first tries a non-blocking receive, then passes of a blocking call to a thread. Sending is just a non-blocking send operation. The SocketCAN and remote interfaces have custom implementations that are true asynchronous methods.

Since the files must be parseable by Python 2.7, there are some ugly workarounds to avoid using `yield from` directly in the code. Sometimes `exec` is used and sometimes they are regular functions returning a Future object.

To make the coroutines be documented as such correctly, I had to add a dependency to sphinxcontrib.asyncio when generating documentation. I tried to make this work with RTD.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,0 +1,4 @@
+requirements_file: doc/requirements.txt
+
+python:
+    version: 3
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,10 +3,10 @@ python:
   - "2.7"
   - "pypy"
   - "pypy3"
+  - "3.3"
   - "3.4"
   - "3.5"
-  - "3.5-dev" # 3.5 development branch
-  - "3.6-dev" # 3.6 development branch
+  - "3.6"
   - "nightly" # currently points to 3.7-dev
 
 # command to run tests
--- a/can/bus.py
+++ b/can/bus.py
@@ -3,6 +3,12 @@ from __future__ import print_function
 
 import abc
 import logging
+import textwrap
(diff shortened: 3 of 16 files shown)


hardbyte:
I don't like the `exec` text hacks. Maybe instead we should consider stopping compatibility with Python 2.7? We could keep a long lived maintenance branch alive for the 2.x release (when we get there) which will keep the current many-version compatibility, then this and other features like type annotations can go into a 3.x series.

christiansandberg:
I know, it's not pretty, but it seems like a normal way to solve it. Some examples are [here](https://github.com/tornadoweb/tornado/blob/c2c0f383f491215eaf8129c8ebc6625c441761bf/tornado/concurrent.py#L181) and [here](https://github.com/mongodb/motor/blob/88bf7d8e6530a6fdc794dbf5a391f47f3814d06a/motor/motor_gridfs.py#L189).

Right now I don't have a specific use case for it. I added it for possible future needs, so it can wait.

christiansandberg:
I've been thinking of skipping making coroutine versions of recv and send and instead using callbacks. Most real life applications will end up with some code being executed when a message is received anyway. The code would be regular Python and many interface APIs implement callback functionality so it would work "natively" on more interfaces.

PR #174, opened 2017-08-02
Remote protocol update

christiansandberg:
The remote interface has been a bit of a learning experience for me. It made sense to make it very low level to make it efficient, but the thought of implementing this over again in some other language made me reconsider. So I've decided to reimplement everything using WebSocket and JSON instead.

A nice side-effect of this is that I could also implement a basic web interface for simple monitoring of CAN traffic using a browser. This could be improved in the future but it's a start. I implemented a JavaScript package on [npm](https://www.npmjs.com/package/python-can-remote) that can be re-used.

Sending data using JSON will of course cause quite a bit of overhead so I added support for sending CAN messages in binary format which could be useful on busy CAN-FD buses for example.

Unfortunately I had a hard time finding a WebSocket library that implemented both the server and client side and worked in Python 2 and 3. So I decided to roll my own minimal WebSocket implementation for now.


hardbyte:
Cool I'll review this and try it out early next week.

Funny story... guess who else rolled their own [websocket implementation](https://stackoverflow.com/questions/12879274/websocket-implementation-in-python-3/12903383) in python 5 years ago!

hardbyte:
I should avoid saying things like "I'll review this early next week"... Sorry for the broken promise.

Ignoring your existing and the updated implementation for a minute, I'm leaning towards having a separate repository for the rest/websocket api. I'd like `python-can` as a library to stay more focused on abstracting different `CAN` devices while allowing such a webserver to be easily built using it.

mgiaco:
very nice.

christiansandberg:
If #179 would be accepted it would be possible to move the remote interface to its own package. The HTTP+WebSocket server is needed for the interface to work so I don't think it's possible to separate it. The web interface could be perhaps be splitted out but then you would have to start a separate server and that would make things more complicated to use.

hardbyte:
I'm sold on #179 so hopefully that will be merged in by `2.1`

What would you like to do with the remote interface in the meantime?

christiansandberg:
I will remove it before 2.0 is released. I will create a separate package for it which will require creating the bus using its own RemoteBus class until we can plug in new interfaces.

christiansandberg:
Moved to new PyPI package: [python-can-remote](https://pypi.python.org/pypi/python-can-remote).

hardbyte:
Sounds good!

PR #178, opened 2017-08-05
Implementation of the serial interface extended and documented.

boris-wenzlaff:
**Extensions for the serial interface**
* Send method
* Shutdown method
* Class and function documentation
* Interface documentation
* Example for the serial interface


can/interfaces/serial/serial_can.py
```diff
@@ -1,9 +1,14 @@
...
+Enable basic can communication over a serial device.
 
-E.g. over bluetooth with "/dev/rfcomm0"
+E.g. over bluetooth with "/dev/rfcomm0" or with Arduino "/dev/ttyACM0" on Linux
+or "COM3" on Windows systems.
 
 """
+# TODO implement test cases
+# TODO implement timeout for send and receive
+# TODO implement cyclic send
```
hardbyte: Could you please open these as issues in github to make it easier to track.

can/interfaces/serial/serial_can.py
```diff
@@ -12,35 +17,92 @@
 try:
     import serial
 except ImportError:
-    logger.error("You won't be able to use the serial can backend without the serial module installed!")
+    logger.error("You won't be able to use the serial can backend without the "
```
hardbyte: For consistency with the other interfaces can you please switch this to use the `warning` level here.

can/interfaces/serial/serial_can.py
```diff
@@ -12,35 +17,92 @@
...
+
+            .. note:: Some serial port implementations don't care about the bit
+                      rate.
+
+        :param float timeout:
+            Timeout for the serial device in seconds (default 0.1).
         """
+
         if channel == '':
             raise TypeError("Must specify a serial port.")
```
hardbyte: This should be a `ValueError` as the type (`str`) is correct

doc/interfaces/serial.rst
```diff
@@ -2,9 +2,18 @@
 
 CAN over Serial
 ===============
-
 A text based interface. For example use over bluetooth with
-``/dev/rfcomm0``
+``/dev/rfcomm0`` or with Arduino ``/dev/ttyACM0``. The interface is a simple
+implementation that is primarily intended for recording CAN traces.
```
hardbyte: Instead of saying the interface has a "primary intended use" say it "has been used for".

doc/interfaces/serial.rst
```diff
@@ -2,9 +2,18 @@
 
 CAN over Serial
 ===============
-
 A text based interface. For example use over bluetooth with
-``/dev/rfcomm0``
+``/dev/rfcomm0`` or with Arduino ``/dev/ttyACM0``. The interface is a simple
```
hardbyte: Let's include a couple of really common serial channels: ``COM1``, ``/dev/ttyUSB0``, ``/dev/ttyS1``.  I don't have a problem linking to your Arduino example, but perhaps a bit later in the docs - after establishing the basics. We want readers to feel confident that this interface will allow them to communicate with anything that can speak serial.

If you feel inspired perhaps you could write a standalone demo in the docs using `socat` and a pseudo term? See e.g. https://stackoverflow.com/questions/52187/virtual-serial-port-for-linux

doc/interfaces/serial.rst
```diff
@@ -2,9 +2,18 @@
...
+implementation that is primarily intended for recording CAN traces.
+To demonstrate the interface, the Arduino Sketch arduino-python-can-serial-demo_
+can be used.
+
+.. _arduino-python-can-serial-demo: https://github.com/boris-wenzlaff/arduino-python-can-serial-demo
+
+.. note::
+    The properties extended_id, is_remote_frame and is_error_frame
+    from the class can.Message are not in use. This interface will not
+    send or receive flags for this properties.
```
hardbyte: Typo: this -> these

doc/interfaces/serial.rst
```diff
@@ -13,5 +22,80 @@ Bus
 
 Internals
 ---------
+The frame that will be send and receive over the serial interface consists of
+six parts. The start and the stop byte for the frame, the timestamp, DLC,
+arbitration ID and the payload. Only the payload has a variable length of bytes
+(0 - 8 bytes), the other parts are fix. Both, the timestamp and the arbitration
+ID will be interpreted as 4 byte unsigned integers. The DLC is also a unsigned
+integer with a length of 1 byte.
```
hardbyte: A few minor english errors in this paragraph.
```
The frames that will be sent and received over the serial interface consist of
six parts. The start and the stop byte for the frame, the timestamp, DLC,
arbitration ID and the payload. The payload has a variable length of between 0 and
8 bytes, the other parts are fixed. Both, the timestamp and the arbitration
ID will be interpreted as 4 byte unsigned integers. The DLC is also an unsigned
integer with a length of 1 byte.
```

doc/interfaces/serial.rst
```diff
@@ -13,5 +22,80 @@ Bus
...
+| **Length (Byte)** | 1              | 4                                             | 1                             | 4                       | 0 - 8   | 1            |
++-------------------+----------------+-----------------------------------------------+-------------------------------+-------------------------+---------+--------------+
+| **Data type**     | Byte           | Unsigned 4 byte integer                       | Unsigned 1 byte integer       | Unsigned 4 byte integer | Byte    | Byte         |
++-------------------+----------------+-----------------------------------------------+-------------------------------+-------------------------+---------+--------------+
+| **Byte order**    | \-             | Little-Endian                                 | Little-Endian                 | Little-Endian           | \-      | \-           |
++-------------------+----------------+-----------------------------------------------+-------------------------------+-------------------------+---------+--------------+
+| **Description**   | Must be 0xAA   | Usually s, ms or µs since start of the device | Length in byte of the payload | \-                      | \-      | Must be 0xBB |
```
hardbyte: As the timestamp could have different scalars for different serial devices (which I'm not entirely sure is reasonable), then we should allow the user to inform the `Serial` class of this scalar to ensure that when serializing/deserializing a frame we apply a correction to the timestamp.

Even if we set some standard (e.g. integer microseconds), there will be some smarter devices (e.g. a raspberry pi) which will actually have epoch time, and some use cases that will want to collect more than `MAX_INT` microseconds. 

@christiansandberg care to weigh in here?

doc/interfaces/serial.rst
```diff
@@ -13,5 +22,80 @@ Bus
...
++-------------------+----------------+-----------------------------------------------+-------------------------------+-------------------------+---------+--------------+
+
+Examples of serial frames
```
hardbyte: These are great - thanks for taking the time.

doc/interfaces/serial.rst
christiansandberg: Is this implemented for some specific device or is it just something custom that is supposed to be implemented on the other side? In that case I think it should be clear how the protocol is specified. Right now it is in seconds but it should probably be in milli or microseconds which means it must be 8 bytes.

Also we might just as well add support for extended/std and remote frames to avoid backwards incompatible changes in the future. Add one byte with various flags.

boris-wenzlaff:
The last commit is only part of the fixes.
**Pending changes:**
* Test cases for the serial interface
* Remove arduino example
* Add pure Python example
* Check backward compatibility  with Python 2

**This will not be part of this deployment:**
* Any changes of the serial interface specification

The specification of the serial interface was derived from the receive method in the interface.  I do not know any public specification or a device that implements the interface. I think it was only implemented for a special case.
Currently I implement the can232 interface. When I'm done, we can bring them face to face and discuss the serial interface specification over the mailing list.
I share your concerns, but I did not want to change the interface specification without a geenrelle discussion.

boris-wenzlaff:
Ready for round two.

I have fixed all open points from the review. As already written, I have made no changes to the protocol itself. Here we should discuss in advance what should be changed and what cross-effects arise when we break the interface.

Thanks for the review Brian and Christan.

boris-wenzlaff:
I merged the develop branch into mine and Travis canceled the Python 2.7 tests (job #306).
After that I make small change to rerun the test (order of imports), the next run of the tests failed (job #308).
On my machine the tests run without any errors (Python 2.7.12 and 3.6.2).

Any ideas? Maybe a problem with Travis?


Error from job #308.1:
ERROR: testSQLWriterReceives (listener_test.ListenerTest)

----------------------------------------------------------------------

  File "/home/travis/build/hardbyte/python-can/test/listener_test.py", line 99, in testSQLWriterReceives

    self.assertEqual(msg[1], 0xDADADA)

TypeError: 'NoneType' object has no attribute '__getitem__'

-------------------- >> begin captured logging << --------------------

can: DEBUG: can config: {'interface': 'virtual', 'channel': None}

can.io.sql: INFO: Creating sqlite db

can.io.sql: DEBUG: Stopping sqlite writer

can.io.sql: INFO: Stopped sqlite writer after writing 0 messages

--------------------- >> end captured logging << ---------------------

examples/serial_com.py
```diff
@@ -0,0 +1,72 @@
...
+
+def time_millis():
+    return int(round(time.time() * 1000))
+
+
+def send_cyclic(bus, msg, stop_event):
+    print("Start to send a message every 1s")
+    start_time = time_millis()
+    while not stop_event.is_set():
+        msg.timestamp = time_millis() - start_time
```
christiansandberg: The timestamp is defined in seconds so this is incompatible with the rest of interfaces.

test/serial_test.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertEqual(msg, msg_receive)
+
+    def test_rx_tx_max_timestamp(self):
+        """
+        Tests the transfer with the highest possible timestamp
+        """
+        msg = can.Message(timestamp=0xFFFFFFFF)
+        self.bus.send(msg)
+        msg_receive = self.bus.recv()
+        self.assertEqual(msg, msg_receive)
```
christiansandberg: When testing two messages if they are equal, timestamp is not compared. You need to test `msg.timestamp` and `msg_receive.timestamp` instead.

test/serial_test.py
```diff
@@ -0,0 +1,150 @@
...
+        self.assertRaises(ValueError, self.bus.send, msg)
+
+    def test_rx_tx_min_timestamp(self):
+        """
+        Tests the transfer with the lowest possible timestamp
+        """
+        msg = can.Message(timestamp=0)
+        self.bus.send(msg)
+        msg_receive = self.bus.recv()
+        self.assertEqual(msg, msg_receive)
```
christiansandberg: Same here...

hardbyte:
Happy to merge once the minor timestamp comments Christian raised are addressed.

PR #179, opened 2017-08-14
Using pluggy to allow plug-able can interfaces

pierreluctg:
Allows user of python-can to create their own plug-able interface without the need to modify the python-can source.

### DIFF
--- a/can/util.py
+++ b/can/util.py
@@ -161,7 +161,7 @@ def load_config(path=None, config=None):
         can.log.debug(
             "interface {} in VALID_INTERFACES".format(system_config['interface'])
         )
-    elif len(get_pluginmanager().pythoncan_interface(
+    elif len(get_pluginmanager().hook.pythoncan_interface(
             interface=system_config['interface'])):
         can.log.debug("interface {} in a plugin".format(system_config['interface']))
     else:


pierreluctg:
I had to modify `.travis.yml` configuration in order to install the dependencies defined in `setup.py`

pierreluctg:
Hi, is there anything missing to get this accepted?

can/interface.py
```diff
@@ -21,6 +24,27 @@
...
+    and possibly input arguments.
+    """
+    return BACKENDS.get(interface)
+
+
+def get_pluginmanager(load_entrypoints=True):
+    pm = PluginManager("pythoncan")
+    pm.add_hookspecs(hookspecs)
+    # XXX load internal plugins here
+    pm.register(sys.modules[__name__])
```
hardbyte: Can you help me understand this code & comment? Is this _just in case_ we wanted to define other hookspecs in the current namespace and have them self register into the plugin manager? Or does this pick up implementations/plugins?

can/interfaces/hookspecs.py
```diff
@@ -0,0 +1,11 @@
+from pluggy import HookspecMarker
+
+
+hookspec = HookspecMarker("pythoncan")
+
+
+@hookspec
+def pythoncan_interface(interface):
+    """ return tuple (module, classname) containing python-can interface info
```
hardbyte: What does it return if the plugin doesn't support the given interface? None/False/Exception?

This would return the `modulename` not the `module`.
Please document the parameter interface (`str`).

can/util.py
```diff
@@ -206,6 +215,16 @@ def set_logging_level(level_name=None):
     log.debug("Logging set to {}".format(level_name))
 
 
+def get_pluginmanager(load_entrypoints=True):
```
hardbyte: This appears to be duplicated?

pierreluctg:
Replaced by #221 since my fork was removed :(

@hardbyte you may want to re-tag #221 

I will add the documentation soon

PR #182, opened 2017-08-22
Vector interface (continued)

christiansandberg:
This pull requests continues the work done in #114 by @JulienGrv. Thanks for your hard work!


can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,176 @@
...
+import sys
+import time
+
+# Import Modules
+# ==============
+from can import BusABC, Message
+
+# Define Module Logger
+# ====================
+LOG = logging.getLogger('can.vector')
```
hardbyte: lowercase `log`

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,176 @@
...
+# Import Standard Python Modules
+# ==============================
+import ctypes
+import logging
+import sys
+import time
+
+# Import Modules
+# ==============
+from can import BusABC, Message
```
hardbyte: No `CanError`?

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,176 @@
...
+# Import safely Vector API module for Travis tests
+if sys.platform == 'win32':
+    from . import vxlapi
+else:
+    LOG.warning('Vector API does not work on %s platform', sys.platform)
+    vxlapi = None
+
+
+class VectorBus(BusABC):
+    """The CAN Bus implemented for the Vector interface."""
```
hardbyte: Should have a default `channel_info` string

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,176 @@
...
+                hw_type = ctypes.c_uint(0)
+                hw_index = ctypes.c_uint(0)
+                hw_channel = ctypes.c_uint(0)
+                xl_status = vxlapi.get_appl_config(
+                    app_name=self._app_name,
+                    app_channel=app_channel,
+                    hw_type=hw_type,
+                    hw_index=hw_index,
+                    hw_channel=hw_channel)
+                if xl_status.value == vxlapi.XL_SUCCESS:
```
hardbyte: Would we also want to log if `xl_status.value` isn't success?

can/interfaces/vector/canlib.py
```diff
@@ -0,0 +1,176 @@
...
+    def recv(self, timeout=None):
+        end_time = time.time() + timeout if timeout is not None else None
+        event = vxlapi.XLevent(0)
+        while True:
+            event_count = ctypes.c_uint(1)
+            try:
+                vxlapi.receive(self.port_handle, event_count, event)
+            except vxlapi.VectorError as exc:
+                if exc.status != vxlapi.XL_ERR_QUEUE_IS_EMPTY:
+                    raise
```
hardbyte: Should we be logging the caught exceptions?

can/interfaces/vector/vxlapi.py
```diff
@@ -0,0 +1,859 @@
...
+_xlapi_dll = None
+if sys.platform == 'win32':
+    architecture = platform.architecture()[0]
+    try:
+        if architecture == '64bit':
+            _xlapi_dll = ctypes.windll.LoadLibrary('vxlapi64')
+        else:
+            _xlapi_dll = ctypes.windll.LoadLibrary('vxlapi')
+    except OSError:
+        if architecture == '64bit':
```
hardbyte: Don't need to check arch again. Just save the libname once:
```
lib_name = 'vxlapi64' if architecture == '64bit' else 'vxlapi'
```

can/interfaces/vector/vxlapi.py
```diff
@@ -0,0 +1,859 @@
...
+            LOG.error('Cannot load Dynamic XL Driver Library (vxlapi64.dll)')
+        else:
+            LOG.error('Cannot load Dynamic XL Driver Library (vxlapi.dll)')
+else:
+    LOG.warning('This library is only availabe for Windows')
+
+# Bus types
```
hardbyte: Your call @christiansandberg but might be nice to have all the constants in own file

can/interfaces/vector/vxlapi.py
```diff
@@ -0,0 +1,859 @@
...
+
+if _xlapi_dll is not None:
+    # xlOpenDriver();
+    xlOpenDriver = _xlapi_dll.xlOpenDriver
+    xlOpenDriver.argtypes = []
+    xlOpenDriver.restype = XLstatus
+    xlOpenDriver.errcheck = check_status
+
+
+    def open_driver():
```
hardbyte: We could use the `ctypesutil.LibraryMixin` or try abstract this argtype/resulttype/status checking better.

doc/interfaces/vector.rst
```diff
@@ -0,0 +1,15 @@
+Vector
+======
+
+This interface adds support for CAN controllers by `Vector`_.
```
hardbyte: Any limitations, a wishlist of things that could be better?

can/interfaces/vector/canlib.py
christiansandberg: I moved all the error raising to vxlapi.

can/interfaces/vector/canlib.py
christiansandberg: Perhaps, pylint actually complains if it is lowercase since it is a module constant...

can/interfaces/vector/canlib.py
christiansandberg: We don't want to log empty queue exception since that will happen hundreds of times per second.

can/interfaces/vector/vxlapi.py
christiansandberg: I don't really understand the purpose of that mixin. I would still need to specify the same amount of information so I don't see what the abstraction is.

can/interfaces/vector/vxlapi.py
christiansandberg: True. Most of the code in this module is not in use so I was thinking of stripping it down to make it easier to find the stuff that matters.

can/interfaces/vector/canlib.py
christiansandberg: I forgot to remove this check since an exception will be raised if it is not successful anyway.

PR #192, opened 2017-10-05
windows install missing win32com step

LivingInSyn:
Added a step to install pywin32 (win32com) to the windows instructions which is a dependency of usb2canSerialFindWin.py and others


hardbyte:
Thanks @LivingInSyn I've added that change to the develop branch in commit https://github.com/hardbyte/python-can/commit/2ce0d55f2ce0f1a48f64b4724e6a99b48690992a

PR #194, opened 2017-10-05
Boolean coercion to True, even for CAN frames with dlc==0

OscarStarink:
The following code did not print CAN frames that had zero data (dlc==0), because boolean coercion uses the overloaded `__len__()` 
```
msg = bus.recv(timeout=1)
if msg :
    print(msg)
```
By overloading `__bool__()` to return True, all CAN frames will be printed. Timeouts return None and will not be printed.

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -83,6 +83,12 @@ class Message(object):
     def __len__(self):
         return len(self.data)
 
+    def __bool__(self):
+        return True
+
+    def __nonzero__(self):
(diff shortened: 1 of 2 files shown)


christiansandberg:
I think it’s a sensible change. In general though it’s safest to explicitly compare against None instead.

If this is to be merged it might be best to do it for the 2.0 release just in case someone relies on the old behavior.

PR #195, opened 2017-10-10
add slcan-support

### LINKED ISSUE #119: Windows compatibility for Lawicel CANUSB
Hello,

Is `python-can` compatible with [Lawicel CANUSB](http://www.can232.com/)? What configuration should I use (interface, channel)?

I'm running Windows python 3.5 x64.

Many thanks!

ebroecker:
add slcan/lawicel serial interface
these where supported with socketcan (slcand) already.
But this simple interfaces makes any slcan-compatible device usable for windows also

Note: no need to rename on of current modules any more.

Currently only tested with python2.7.

this should fix #119 and #5

please have a look on it.
I'll do further work on it, if needed.


can/interfaces/slcan.py
```diff
@@ -0,0 +1,147 @@
...
+        super(slcanBus, self).__init__(channel, **kwargs)
+
+    def recv(self, timeout=None):
+        if timeout is not None:
+            self.serialPortOrig.timeout = timeout
+
+        canId = None
+        remote = False
+        frame = []
+        str = self.serialPort.readline()
```
christiansandberg: I would suggest to not use `str` as it is defined by Python.

can/interfaces/slcan.py
```diff
@@ -0,0 +1,147 @@
...
+            timeout in seconds when reading message
+        """
+
+
+        if channel == '':
+            raise TypeError("Must specify a serial port.")
+        if '@' in channel:
+            (channel, ttyBaudrate) = channel.split('@')
+
+        self.poll_interval = poll_interval
```
christiansandberg: This is not used.

can/interfaces/slcan.py
```diff
@@ -0,0 +1,147 @@
...
+                               extended_id=extended,
+                               timestamp=time.time(),   # Better than nothing...
+                               is_remote_frame=remote,
+                               dlc=dlc,
+                               data=frame)
+            else:
+                return None
+
+    def send(self, msg, timeout=None):
+        if bool(msg.is_remote_frame):
```
christiansandberg: No need to convert to bool. Same for the other below.

doc/conf.py
```diff
@@ -204,6 +204,6 @@ def __getattr__(cls, name):
         else:
             return Mock()
 
-MOCK_MODULES = ['serial']
+MOCK_MODULES = ['basic_serial']
```
christiansandberg: Is this change needed now?

can/interfaces/slcan.py
ebroecker: yes, you are right. 
poll_intervall is a copy-paste artefact
I'll remove

ebroecker:
@christiansandberg 
All your commends should be fixed.

christiansandberg:
Thanks, but the documentation must also be updated with the new interface.

ebroecker:
some base documentation added

ebroecker:
@christiansandberg 
Do I have anything else to do?

christiansandberg:
Sorry. Update also doc/configuration.rst (and doc/history.rst if you want). Then I think you got the of it covered. :)

hardbyte:
Thanks Eduard!

PR #198, opened 2017-10-16
Remove intermediate namedtuple in socketcan_native interface.

eerimoq:
Hello,

first of all, thanks for a useful and easy to use package that serves its purpose well.

When I started implementing this PR I didn't realize that the `capturePacket()`-function was part of the user API, though, listed under "Internals" in the documentation. This PR makes non-backwards compatible changes to this interface since the function was both renamed and changed its return type from a namedtuple to a Message. Is this ok? Who is the intended user of the internal functions? Usually users use the Bus-object to communicate over a CAN BUS, right?

Anyway, the purpose of this PR is to improve performance and reduce the code size to make it easier to read.

### DIFF
--- a/can/interfaces/socketcan/socketcan_native.py
+++ b/can/interfaces/socketcan/socketcan_native.py
@@ -10,7 +10,6 @@ import select
 import threading
 import socket
 import struct
-from collections import namedtuple
 
 import errno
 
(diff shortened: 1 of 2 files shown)


hardbyte:
Thanks Erik - well spotted! If you notice any other code smells like that please raise issues or open more pull requests.

PR #199, opened 2017-10-18
Add support for asc reading and for .log-format  

ebroecker:
Here is my next pull request:

* Add (io-)Support for reading .asc files
* Add read and write support for  .log-Format (linux canutils)  

any interest?


can/io/asc.py
```diff
@@ -1,7 +1,49 @@
...
+    def __iter__(self):
+        for line in self.fp:
+            temp = line.strip()
+            if len(temp) > 0:
+                if temp[0].isdigit():
+                    lineArray = temp.split()
+                    if lineArray[1].isdigit() and lineArray[2] != "Statistic:":
+                        time = float(lineArray[0])
+                        channel = lineArray[1]
+                        can_id = int(lineArray[2],16)
```
christiansandberg: Extended frames have an x suffixed after the CAN ID (see [here](http://zone.ni.com/reference/en-XX/help/370859J-01/dlgcanconverter/dlgcanconverter/canconverter_ascii_logfiles/)), so this will probably fail.

Have a look at listener_test.py and BLFTest and make a test that writes an ASC log file and then reads it back again.

ebroecker: Thanks for the feedback.
I'll add extended support for IDs and add some documentation and tests.

ebroecker:
most should be done.
could you do another review?

regards,
Eduard

test/listener_test.py
```diff
@@ -216,6 +218,42 @@ def test_reader_writer(self):
             self.assertEqual(msg1, msg2)
             self.assertAlmostEqual(msg1.timestamp, msg2.timestamp)
 
+class canutilsLog(unittest.TestCase):
```
hardbyte: I think we should move these new tests to their own files, and rename the Test class.

test/listener_test.py
```diff
@@ -216,6 +218,42 @@ def test_reader_writer(self):
...
+            writer(msg)
+        writer.stop()
+
+        messages = list(can.canutilsLogReader(filename))
+        self.assertEqual(len(messages), len(TEST_MESSAGES))
+        for msg1, msg2 in zip(messages, TEST_MESSAGES):
+            self.assertEqual(msg1, msg2)
+            self.assertAlmostEqual(msg1.timestamp, msg2.timestamp)
+
+class ascFileFormat(unittest.TestCase):
```
hardbyte: Move to own file and rename to CamelCase and include `Test`

test/listener_test.py
```diff
@@ -216,6 +218,42 @@ def test_reader_writer(self):
...
+
+    def test_reader_writer(self):
```
hardbyte: This test looks good. If you have them, would you mind also including a `.asc` file sourced using another tool like Vector CANalyzer or CANoe and add it to e.g., `test/data/vectorlogfile.asc` etc?

can/io/asc.py
```diff
@@ -1,7 +1,59 @@
...
+                            dlc = int(lineArray[5])
+                            frame = bytearray()
+                            for byte in lineArray[6:6 + dlc]:
+                                frame.append(int(byte,16))
+                            msg = Message(timestamp=time,
+                                        arbitration_id=can_id & 0x1FFFFFFF,
+                                        extended_id=isExtended,
+                                        is_remote_frame=False,
+                                        dlc=dlc,
+                                        data=frame)
```
hardbyte: Many of these inner most branches could be extracted into clearly named helper methods.

can/io/__init__.py
```diff
@@ -5,7 +5,8 @@
 
 from .logger import Logger
 from .player import LogReader
-from .asc import ASCWriter
+from .log import canutilsLogReader, canutilsLogWriter
```
hardbyte: Classes in python should start with Upper case

can/io/asc.py
```diff
@@ -1,7 +1,59 @@
...
+                        if lineArray[2].endswith("x") or lineArray[2].endswith("X"):
+                            isExtended = True
+                            can_id = int(lineArray[2][0:-1], 16)
+                        else:
+                            isExtended = False
+                            can_id = int(lineArray[2],16)
+                        frameType = lineArray[4]
+                        if frameType == 'r' or frameType == 'R':
+                            msg = Message(timestamp=time,
+                                        arbitration_id=can_id & 0x1FFFFFFF,
```
hardbyte: Extract the constant `0x1FFFFFFF`

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+import datetime
+import time
+from can.message import Message
+
+CAN_MSG_EXT = 0x80000000
+CAN_ERR_FLAG = 0x20000000
+CAN_ERR_BUSERROR = 0x00000080
+CAN_ERR_DLC = 8
+
+class canutilsLogReader(object):
```
hardbyte: Start classes with a Capital letter please

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    def __init__(self, filename):
+        self.fp = open(filename, "r")
+
+    def __iter__(self):
+        for line in self.fp:
+            temp = line.strip()
+            if len(temp) > 0:
+                (timestamp, bus, frame) = temp.split()
```
hardbyte: This is much nicer that the `.asc` version knowing the names! 
You can remove the unnessasary quotes on the left hand side though.

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+                timestamp = float(timestamp[1:-1])
```
hardbyte: You want to skip the first and last character of `timestamp` before parsing as a float - Is this what you want? What are those characters? Can we `strip()` to make this clearer? Otherwise I think it deserves a comment for people like me who are unfamiliar with the format.

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+                (canId, data) = frame.split("#")
+                if len(canId) > 3:
+                    isExtended = True
+                else:
+                    isExtended = False
+                canId = int(canId, 16)
+                if len(data) > 0 and (data[0] == "R" or data[0] == "r"):
```
hardbyte: Another option:
```
data[0].lower() == "r"
```

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+                    isRemoteFrame = True
+                    if len(data) > 1:
+                        dlc = int(data[1:])
+                    else:
+                        dlc = 0
+                else:
+                    isRemoteFrame = False
+
+                    dlc = len(data) / 2
```
hardbyte: I'd convert this to an `int` here

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+                    dataBin = bytearray()
+                    for i in range(0, int(dlc)):
```
hardbyte: This might be simpler - one idea is to do your index math here using `range(0, 2 * dlc, 2)`.

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+
+                if canId & CAN_ERR_FLAG and canId & CAN_ERR_BUSERROR:
+                    msg = Message(timestamp=timestamp, is_error_frame=True)
+                else:
+                    msg = Message(timestamp=timestamp, arbitration_id=canId & 0x1FFFFFFF,
+                              extended_id=isExtended, is_remote_frame=isRemoteFrame, dlc=dlc, data=dataBin)
+                yield msg
+
+
+class canutilsLogWriter(Listener):
```
hardbyte: class name

can/io/log.py
```diff
@@ -0,0 +1,95 @@
...
+    def on_message_received(self, msg):
+        if msg.is_error_frame:
+            self.log_file.write("(%f) vcan0 %08X#0000000000000000\n" % (msg.timestamp, CAN_ERR_FLAG | CAN_ERR_BUSERROR, ))
+            return
+
+        timestamp = msg.timestamp
+        if timestamp >= self.started:
+            timestamp -= self.started
+
+        if self.log_file is not None:
```
hardbyte: This check should be either removed or moved up before any attempt to write to the `log_file` (currently an error frame would be attempted to be written to `None`)

examples/simpleLogConvert.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env python
+import sys
```
hardbyte: Just a one line comment showing example arguments is always handy

ebroecker:
@hardbyte 
You are right with most/all of your comments. 
I'll fix them these days.
There are one or two comments, which I currently don't understand. 
But let me fix the others first and discuss the remaining later...

can/io/log.py
ebroecker: maybe somewhere a example .log-Entry should be documented:
.log-file looks like:
(0.0) vcan0 001#8d00100100820100
after splitting (0.0) remains, thus I need [1:-1] to cut of the braces

hardbyte:
Thanks Eduard - happy to clarify anything further. Don't feel bad about pushing back on some of my comments too. Let me know when you'd like me to take another look.

Lastly as this is about ready for merging - do you want to merge or rebase your changes on top of my develop branch?

can/io/log.py
ebroecker: I changed .asc to work with named vars also.

ebroecker:
hopefully got all your comments in the code.
could you just have another look?

I merged your top develop branch as rebase made errors I was to lazy to fix.

Give me some hint, if I should change some further things (maybe I got some comments wrong).

Feel free to assign issues to me which may appear in future if they are about log or asc,  ...

can/io/log.py
hardbyte: Ok that sounds reasonable, I think it would be slightly clearer to use `strip("()")`

can/io/asc.py
```diff
@@ -1,7 +1,65 @@
...
+class ASCReader(object):
+    """
+    Iterator of CAN messages from a ASC Logging File.
+    """
+
+    def __init__(self, filename):
+        self.fp = open(filename, "r")
+
+    def __iter__(self):
+        def extractCanId(strCanId):
```
hardbyte: This should probably be a class staticmethod

ebroecker:
anything open from my side?

hardbyte:
Thanks for the contribution! 

👍 all good

PR #203, opened 2017-11-19
ixxat: In recv(), avoid the use of canChannelWaitRxEvent()

### LINKED ISSUE #201: Possible Bug in IXXAT Implementation
I'm not quite shure if it's an actual bug or a missunderstanding on my side.

If we look at the main part of the [IXXATBus.recv() method](https://github.com/hardbyte/python-can/blob/develop/can/interfaces/ixxat/canlib.py#L387), mainly at the part which handles the timeout != 0 case ([Line 408 ff](https://github.com/hardbyte/python-can/blob/develop/can/interfaces/ixxat/canlib.py#L408)):

```python
# Wait if no message available
t0 = _timer_function()
elapsed_ms = 0
remaining_ms = 0
while elapsed_ms <= timeout:
	elapsed_ms = int((_timer_function() - t0) * 1000)
	remaining_ms = timeout - elapsed_ms
	# Wait until at least one frame is in the buffer
	try:
		_canlib.canChannelWaitRxEvent(self._channel_handle, remaining_ms)
	except VCITimeout:
		log.debug('canChannelWaitRxEvent timed out after %dms', remaining_ms)
		return None

	# In theory we should be fine with a 0 timeout since the rxEvent was already
	# set but I've seen timeouts appearing here and there
	try:
		_canlib.canChannelReadMessage(self._channel_handle, 0, ctypes.byref(self._message))
	except VCITimeout:
		continue

	# See if we got a data or info/error messages
	if self._message.uMsgInfo.Bits.type == constants.CAN_MSGTYPE_DATA:
		tm = _timer_function()
		break

	elif self._message.uMsgInfo.Bits.type == constants.CAN_MSGTYPE_INFO:
		log.info(CAN_INFO_MESSAGES.get(self._message.abData[0], "Unknown CAN info message code {}".format(self._message.abData[0])))

	elif self._message.uMsgInfo.Bits.type == constants.CAN_MSGTYPE_ERROR:
		log.warning(CAN_ERROR_MESSAGES.get(self._message.abData[0], "Unknown CAN error message code {}".format(self._message.abData[0])))

	elif self._message.uMsgInfo.Bits.type == constants.CAN_MSGTYPE_TIMEOVR:
		pass
	else:
		log.warn("Unexpected message info type")
```

Now consider multiple messages arrive very fast and a RxEvent occours. We start processing the first message, but it happens to be a CAN info or error message. This is fine, we just ignore the data and return to the top of the loop, where we wait for the next RxEvent. In my understanding `canChannelWaitRxEvent` will timeout despite having data in the Queue, therefore `None` is returned instead of a new message.

The same would happen if multiple messages arrive very fast and one readout cycle lasts longer than the time between two messages.

I might be wrong, so please correct me if the behaviour is not as described. If someone can confirm this behaviour, I can provide a fix for this.

benjarobin:
The function canChannelWaitRxEvent() does not always return even if there is one message in the FIFO. This looks like a regression with at least VCI V4 4.0.482 since with VCI v3 3.5.4 there is no problem.

This commit also refactor the recv() function in order to be more robust.
    
While testing this fix, I discover that canChannelReadMessage(), with VCI V4, can randomly return the error code VCI_E_RXQUEUE_EMPTY, which by the documentation should never happen, so catch it...

Fix #201

### DIFF
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -388,62 +388,56 @@ class IXXATBus(BusABC):
         """ Read a message from IXXAT device. """
 
         # TODO: handling CAN error messages?
-        if timeout is None:
-            timeout = constants.INFINITE
-        else:
-            timeout = int(timeout * 1000)
+        data_received = False
 
-        tm = None
         if timeout == 0:
             # Peek without waiting
             try:
                 _canlib.canChannelPeekMessage(self._channel_handle, ctypes.byref(self._message))
-            except VCITimeout:
-                return None
-            except VCIRxQueueEmptyError:
+            except (VCITimeout, VCIRxQueueEmptyError):
                 return None
             else:
                 if self._message.uMsgInfo.Bits.type == constants.CAN_MSGTYPE_DATA:
-                    tm = _timer_function()
+                    data_received = True
         else:
             # Wait if no message available
-            t0 = _timer_function()
-            elapsed_ms = 0
-            remaining_ms = 0
-            while elapsed_ms <= timeout:
-                elapsed_ms = int((_timer_function() - t0) * 1000)
-                remaining_ms = timeout - elapsed_ms
-                # Wait until at least one frame is in the buffer
+            if timeout is None or timeout < 0:
+                remaining_ms = constants.INFINITE
+                t0 = None
+            else:
+                timeout_ms = int(timeout * 1000)
+                remaining_ms = timeout_ms
+                t0 = _timer_function()
+
+            while True:
                 try:
-                    _canlib.canChannelWaitRxEvent(self._channel_handle, remaining_ms)
-                except VCITimeout:
-                    log.debug('canChannelWaitRxEvent timed out after %dms', remaining_ms)
-                    return None
-
-                # In theory we should be fine with a 0 timeout since the rxEvent was already
-                # set but I've seen timeouts appearing here and there
-                try:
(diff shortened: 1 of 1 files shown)


cowo78:
Mark the VCI library version you're using (and getting the error with), could be useful as a future reference.
Just my 2c, since I don't remember getting such an error and I'm not using/developing the module anymore.

benjarobin:
@cowo78 The latest which is VCI V4 4.0.482, tested with an (old) ixxat USB to CAN II (v1), which is a dual CAN USB module

cowo78:
I was using VCI V3, as stated in doc/installation.rst. We could decide to switch to VCI V4 and remove compatibility with V3.
Would it feasible for you to perform your tests with VCI V3 and vanilla python-can?

benjarobin:
Interesting, with VCI v3 3.5.4, and vanilla python-can (latest Git commit), the recv is working normally (canChannelWaitRxEvent does not look buggy).  
I also test this pull request with VCI v3, and this is working as expected.

cowo78:
So it seems that there's some bug in canChannelWaitRxEvent for VCI V4. I would suggest to merge this pull request, just drop in some comments to mark the buggy behaviour. If you're feeling courageous you may also report this to IXXAT.

benjarobin:
When I will have 5 minutes, I will force push new commits :
 - Remove commit about exclusive mode : this really should not be in this PR
 - Cleanup comments...

benjarobin:
@cowo78 Pull request updated, I only changed the commits messages

PR #206, opened 2017-11-30
Clarify flags in can.Message.__str__

jcassette:
This is a proposal to make the output of `can.Message.__str__` more understandable, especially the flags.

Current:
```
>>> print(can.Message())
Timestamp:        0.000000    ID: 00000000    010    DLC: 0
>>> print(can.Message(is_remote_frame=True))
Timestamp:        0.000000    ID: 00000000    110    DLC: 0
>>> print(can.Message(is_error_frame=True))
Timestamp:        0.000000    ID: 00000000    011    DLC: 0
```

Proposed:
```
>>> print(can.Message())
Timestamp:        0.000000    ID: 00000000     X     DLC: 0
>>> print(can.Message(is_remote_frame=True))
Timestamp:        0.000000    ID: 00000000    RX     DLC: 0
>>> print(can.Message(is_error_frame=True))
Timestamp:        0.000000    ID: 00000000     XE    DLC: 0
```

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -47,18 +47,11 @@ class Message(object):
             arbitration_id_string = "ID: {0:04x}".format(self.arbitration_id)
         field_strings.append(arbitration_id_string.rjust(12, " "))
 
-        flag_string = "".join(
-            map(
-                str,
-                map(
-                    int, [
-                        self.is_remote_frame,
-                        self.id_type,
-                        self.is_error_frame,
-                    ]
-                )
-            )
-        )
+        flag_string = " ".join([
+            "X" if self.id_type else "S",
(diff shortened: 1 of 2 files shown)


hardbyte:
Hi Julien, thanks for your proposal.

I agree that it makes the representation more easily understood. If you would update the docs in `doc/message.rst` I'd be happy to merge this in for the 2.0 release.

jcassette:
Hi Brian, and thank you for your reply.
I would be happy to update the doc.
Also, I am not sure RXE is ideal, maybe X R E (with spaces) is better so that there is no confusion with RX/TX. What do you think?

PR #207, opened 2017-12-03
Add CAN-FD support

### LINKED ISSUE #173: Support CAN-FD
The current car CAN specification is moving to FD. 
FD is also supported in Kvaser's library. It would be very useful if you could add FD's functionality to a very powerful Python-can library

christiansandberg:
Initial support for socketcan_native and kvaser. Tested on virtual CAN only.
Experimental support for BLF logger
Only warn on DLC too long since some loggers support long data lengths but not FD frames

Fixes #173.


felixdivo:
This addresses #173 and #207.

can/interfaces/socketcan/socketcan_native.py
```diff
@@ -308,7 +338,7 @@ def captureMessage(sock):
     """
     # Fetching the Arb ID, DLC and Data
     try:
-        cf, addr = sock.recvfrom(can_frame_size)
+        cf, addr = sock.recvfrom(CAN_FRAME_HEADER_STRUCT.size + 64)
```
hardbyte: Maybe pull this magic number out?

`CANFD_MTU`?

can/interfaces/socketcan/socketcan_native.py
```diff
@@ -338,6 +368,9 @@ def captureMessage(sock):
     is_extended_frame_format = bool(can_id & 0x80000000)
     is_remote_transmission_request = bool(can_id & 0x40000000)
     is_error_frame = bool(can_id & 0x20000000)
+    is_fd = len(cf) > 16
```
hardbyte: Reading through the [can kernel docs](https://www.kernel.org/doc/Documentation/networking/can.txt) we could probably be more explicit here - checking the len(cf) is exactly what we expect.

>   Example:
>    [ remember: CANFD_MTU == sizeof(struct canfd_frame) ]

```
    struct canfd_frame cfd;

    nbytes = read(s, &cfd, CANFD_MTU);

    if (nbytes == CANFD_MTU) {
            printf("got CAN FD frame with length %d\n", cfd.len);
	    /* cfd.flags contains valid data */
    } else if (nbytes == CAN_MTU) {
            printf("got legacy CAN frame with length %d\n", cfd.len);
	    /* cfd.flags is undefined */
    } else {
            fprintf(stderr, "read: invalid CAN(FD) frame\n");
            return 1;
    }
```

can/interfaces/kvaser/canlib.py
```diff
@@ -437,7 +481,7 @@ def recv(self, timeout=None):
         Read a message from kvaser device.
         """
         arb_id = ctypes.c_long(0)
-        data = ctypes.create_string_buffer(8)
+        data = ctypes.create_string_buffer(64)
```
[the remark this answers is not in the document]
christiansandberg: Yes it will be limited by the dlc later.

can/interfaces/kvaser/canlib.py
```diff
@@ -518,11 +572,11 @@ def flash(self, flash=True):
 
     def shutdown(self):
         # Wait for transmit queue to be cleared
-        try:
-            canWriteSync(self._write_handle, 100)
-        except CANLIBError as e:
-            log.warning("There may be messages in the transmit queue that could "
-                        "not be transmitted before going bus off (%s)", e)
+        # try:
```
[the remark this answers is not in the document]
christiansandberg: It was actually when I tested on virtual CAN on Linux that it failed if there were no messages in the buffer. That’s why I removed it for now.

can/interfaces/socketcan/socketcan_native.py
christiansandberg: Good idea.

PR #212, opened 2017-12-11
Added length to blf reader by objectCount

zorce:
Added objectCount as length to keep track of read progression

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -87,7 +87,11 @@ class BLFReader(object):
         header = FILE_HEADER_STRUCT.unpack(data)
         #print(header)
         assert header[0] == b"LOGG", "Unknown file format"
+        self.file_size = header[10]
+        self.uncompressed_size = header[11]
+        self.object_count = header[12]
         self.start_timestamp = systemtime_to_timestamp(header[14:22])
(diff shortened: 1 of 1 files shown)


can/io/blf.py
```diff
@@ -87,7 +87,12 @@ def __init__(self, filename):
         header = FILE_HEADER_STRUCT.unpack(data)
         #print(header)
         assert header[0] == b"LOGG", "Unknown file format"
+        self.objectCount = header[12]
```
christiansandberg: In order to comply with PEP8 this should be `self.object_count`.

can/io/blf.py
```diff
@@ -87,7 +87,12 @@ def __init__(self, filename):
...
         self.start_timestamp = systemtime_to_timestamp(header[14:22])
+        self.stop_timestamp = systemtime_to_timestamp(header[22:30])
+
+    def __len__(self):
+        return self.objectCount
```
christiansandberg: This is not really accurate as object count includes all objects while `__len__()` should return number of CAN messages + error frames as that is what you get when you iterate over this object.

zorce:
I agree ideally it should show only the can frames, but it at least gives an indication on how mush you still have to process. Can remove the __len__ and just have object_count then.

Issue #98, opened 2016-11-30
Document log file formats better


hardbyte:
The documentation doesn't for instance mention the sqlite schema.

felixdivo:
This adds some information about the sqlite format: #214, or more specifically 5b097fcf2e27e899acfc518c53b0d630e245a12d

felixdivo:
Is more documentation needed anywhere?

PR #214, opened 2017-12-14
Fixes an bunch of minor problems, adds some documentation of the SQL database format

felixdivo:


felixdivo:
I am currently writing more unit tests for the writer/reader combinations.

felixdivo:
I added some more messages to the tests, but now some of them fail. The messages are declared at the top of `test/logformats_test.py`. 

I generally have the impression that many test could be simplified using the method I copied and parameterized at the top of that file. Should I do that?

felixdivo:
Another thing that got me thinking: It is possible to create a message with `data=bytearray(0x01, 0x02, 0x03)` but set `dlc=6`. Shouldn't the DLC be checked against the length of the data, or even just be derived from that? Or is that allowed on purpose?

felixdivo:
The fix in 65e1800 actually fixed a bug. The "fix" in 682e055 simply removed the problematic test case: `CanutilsLogWriter`/`CanutilsLogReader` and `ASCWriter`/`ASCReader` seem to fail when working with error frame messages. Is that known / acceptable? I'm asking since it isn't documented anywhere ...

hardbyte:
Hi Felix - thanks for the flurry of changes! It might take a while to go through the changes all combined together like this - if you want to split related items into their own PR we can probably start folding in some of your changes straight away.

Regarding setting the DLC to be different from the actual number of bytes - that was intentional as other software/hardware implementations or protocols might do so and I want python-can to be able to reproduce or interact with such systems.

I'm a fan of simpler tests - please be my guest!

felixdivo:
Okay, I created the issues for the error frame handling. I will create a new PR in which the message object is better documented; it will mention what the constructor checks and what not. The PR will also contain some more structured tests for the `io` modules.

felixdivo:
I just fixed a small merge conflict with the current `develop` branch. I will open another pull request (with more docs & better tests) if this one gets merged.

felixdivo:
@hardbyte Is this still being reviewed?

hardbyte:
Hey sorry minor communication error - I thought you might break the changes into separate pull requests. 

I've gone through and merged in these changes now - thanks a whole lot!

Closes #98

felixdivo:
Oh, I thought I would break future changes in more PRs. ^^ Sorry.

Issue #157, opened 2017-04-04
Software filtering of messages


christiansandberg:
Some interfaces support multiple filters, some support only one filter, and some do not support filtering at all or uses ranges instead of ID and mask (PCAN).

It could reduce the burden of the application if python-can filtered out messages according to the configured filtering in software. I [did this in the kvaser interface](https://github.com/hardbyte/python-can/blob/1.5.2/can/interfaces/kvaser/canlib.py#L448) before but I think it's better to implement something that all interfaces can use.

Suggestion is to implement this in either the BusABC class or in the utils package so that specific interface implementations can take advantage of them.

felixdivo:
It could be implemented in `BusABC` and some interfaces could override this if they are able to do it in the OS kernel or in hardware.

felixdivo:
I have thought a little bit about this and want to propose this new `BusABC` implementation:

```Python

from abc import ABCMeta, abstractmethod

class BusABC():
    """A sketch to show the suggested design.

    The comments/docs/implementation are of course shortened
    to make it more readable. In the release this would be
    documented way better.

    New interfaces would need to implement:

        * __init__()
        * _recv_internal()
        * send()
        * set channel_info atribute

    That is pretty much the same as perviously, only that
    recv() is now called _recv_internal() and must return
    `(msg, False)` instead of only `msg` if it didn't support
    filters before and `(msg, True)` if it did.

    It doesn't change *ANY* method signatures used by the actual
    users of the library, and only requires minimal changes in
    the infterfaces.
    """

    @abstractmethod
    def __init__(self, channel=None, can_filters=None, **config):
        """Same as always, except it now has to call self._apply_filters()
        insteat of setting the field itself."""
        self._apply_filters(can_filters)

    def recv(self, timeout=None):
        """Not abstract any more."""

        msg, already_filtered = self._recv_internal(timeout=timeout)

        if msg and (already_filtered or self._matches_filter(msg)):
            return msg

        else:
            # discard message & try next one if there still is time

            elapsed_time = 0 # actually measure this in the real implementation

            if elapsed_time >= timeout:
                return None
            else:
                # is a call stack overflow a real scenario here?
                return self.rcev(timeout=timeout - elapsed_time)

    @abstractmethod
    def _recv_internal(self, timeout=None):
        """The old implementation of recv() moves to here.
        This method returns a tuple (msg, already_filtered)
        containing the message that was rceived (or None) and
        whether it was already filtered by the bus (for example
[...]

PR #220, opened 2018-01-09
Replace neovi inteface inplementation

pierreluctg:
Replacing the pyneovi implementation with ICS own python interface python-ics.

`neovi` interface is replaced by `icsneovi` interface.

`neovi` interface is now pointing to the the `icsneovi` interface for compatibility.
  
python-ics is maintained by Intrepid Control Systems them self and is available in pypi.


pierreluctg:
FYI @hardbyte and @christiansandberg

can/interface.py
```diff
@@ -18,7 +18,8 @@
     'nican':            ('can.interfaces.nican', 'NicanBus'),
     'iscan':            ('can.interfaces.iscan', 'IscanBus'),
     'virtual':          ('can.interfaces.virtual', 'VirtualBus'),
-    'neovi':            ('can.interfaces.neovi_api', 'NeoVIBus'),
+    'neovi':            ('can.interfaces.ics_neovi', 'NeoViBus'),
+    'icsneovi':         ('can.interfaces.ics_neovi', 'NeoViBus'),
```
christiansandberg: Why not keep using "neovi"?

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -0,0 +1,255 @@
...
+            self.dev.SerialNumber,
+            channel
+        )
+        logger.info("Using device: {}".format(self.channel_info))
+
+        ics.load_default_settings(self.dev)
+
+        self.sw_filters = None
+        self.set_filters(can_filters)
+        self.rx_buffer = queue.Queue()
```
christiansandberg: A Queue is overkill since the code is single threaded. You can use a list instead.

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def _process_msg_queue(self, timeout=None):
+        try:
+            messages, errors = ics.get_messages(self.dev, False, timeout)
+        except ics.RuntimeError:
+            return
+        for ics_msg in messages:
+            if ics_msg.NetworkID != self.network:
+                continue
+            if ics_msg.ArbIDOrHeader == 0:
```
christiansandberg: 0 is actually a common CAN-ID (e.g. in the CANopen protocol).

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -0,0 +1,255 @@
...
+            if not self._is_filter_match(ics_msg.ArbIDOrHeader):
+                continue
+            self.rx_buffer.put(ics_msg)
+        if errors:
+            logger.warning("%d errors found" % errors)
+
+            for msg in ics.get_error_messages(self.dev):
+                logger.warning(msg)
+
+    def _is_filter_match(self, arb_id):
```
christiansandberg: I've removed the SW filtering from the other interfaces and created issue #157 to do a more general solution, so you could remove the filtering for now.

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -0,0 +1,255 @@
...
+            #
+            # This timestamp is useful for time comparing with other system
+            # events or data which is not synced with the neoVI timestamp.
+            #
+            # Currently, TimeSystem2 is not used.
+            return ics_msg.TimeSystem
+        else:
+            # This is the hardware time stamp.
+            # The TimeStamp is reset to zero every time the OpenPort method is
+            # called.
```
christiansandberg: Since the timestamp is zeroed on OpenPort I see a possibility to create an offset using `time.time()` and add that to the timestamp to have an absolute timestamp.

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -0,0 +1,255 @@
...
+                ics_msg.StatusBitField & ics.SPY_STATUS_XTD_FRAME
+            ),
+            is_remote_frame=bool(
+                ics_msg.StatusBitField & ics.SPY_STATUS_REMOTE_FRAME
+            )
+        )
+
+    def recv(self, timeout=None):
+        try:
+            self._process_msg_queue(timeout=timeout)
```
christiansandberg: Shouldn't we check the rx_buffer for available messages first?

can/interface.py
pierreluctg: Agree, will change to keep the `neovi` interface name.

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: When are you planing to add this feature in? We have code that currently depend on filtering the messages using this method.

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Agree, this is some left over code from some tests. However, I am still not set on if a separated "messages pump" thread/task is needed or not. The driver can only hold 20000 messages before dropping older messages.

I may keep the queue and have a single use thread thread get the messages when `recv` is called.

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Will add this

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Agree, however this is because the driver can only hold 20000 messages. So pumping the messages on each `recv` call is to avoid dropping messages. But, this is also downside to that.

See comment above regarding the Queue usage.

can/interfaces/ics_neovi/neovi_bus.py
christiansandberg: The other interfaces also have a limited receive buffer. It is up to the application to process messages continuously, often using a thread. So you don’t have to solve that here. The BufferedReader class can also be used.

can/interfaces/ics_neovi/neovi_bus.py
christiansandberg: No plan. If you need it then definitely keep it. But it won’t work for all interfaces.

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: For that the user should use http://python-ics.readthedocs.io/en/latest/#ics.set_rtc. The "OpenPort" is not necessarily called only called by us. The driver allow multiple clients.

Do you think we should set the RTC when we open the device? Should we provide a API in this class toto set the RTC?

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Make sense.

Will use `collections.deque` instead of a `list` for efficiency.
https://docs.python.org/3/tutorial/datastructures.html#using-lists-as-queues

can/interfaces/ics_neovi/neovi_bus.py
christiansandberg: I'm not familiar with how the NEOVI interfaces work so I'll leave that up to you to decide.

pierreluctg:
Thank you @christiansandberg

PR #221, opened 2018-01-09
Using pluggy to allow plug-able can interfaces.

pierreluctg:
Allows user of python-can to create their own plug-able interface without the need to modify the python-can source.

- [x] Add documentation


can/plugin.py
```diff
@@ -0,0 +1,13 @@
+from pluggy import PluginManager
+
+from can.interfaces import hookspecs
```
hardbyte: There doesn't appear to be a hookspecs?

can/util.py
```diff
@@ -4,8 +4,11 @@
 """
 from __future__ import absolute_import
 
+from pluggy import PluginManager
+
 import can
-from can.interfaces import VALID_INTERFACES
+from can.interfaces import VALID_INTERFACES, hookspecs
```
hardbyte: Missing import (`hookspecs`) breaks the build - https://travis-ci.org/hardbyte/python-can/jobs/328610852

pierreluctg:
Hi @hardbyte, I have simplified the plugin interface and add documentation.

pierreluctg:
Hi @hardbyte, is there anything else you would like to see changed in this PR?

Thank you

PR #222, opened 2018-01-11
Fix the KeyError: 'interface'

pierreluctg:
Existing code throw in a `KeyError: 'interface'` when a invalid interface name is used.


pierreluctg:
FYI @hardbyte and @christiansandberg

PR #232, opened 2018-01-17
Adding string representation to CAN Bus Abstract Base Class

pierreluctg:
Adding \_\_str\_\_ to CAN Bus Abstract Base Class for a nicely printable
string representation of an the Bus using channel_info.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -47,6 +47,9 @@ class BusABC(object):
         """
         pass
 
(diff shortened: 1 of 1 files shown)


felixdivo:
Maybe the bus type should be added?

pierreluctg:
Hi @felixdivo, most interfaces are already setting `channel_info`.

PR #234, opened 2018-01-31
add interface example definition for beginners

mgiaco:
I think that would be great for beginners.

cheers

### DIFF
--- a/examples/send_one.py
+++ b/examples/send_one.py
@@ -3,7 +3,10 @@ import can
 
 
 def send_one():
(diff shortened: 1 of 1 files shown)


christiansandberg:
I'm thinking that we could also put a quick start example in the README with some connection examples like you have and how to receive and send messages.

hardbyte:
My only issue is it doesn't reference the auto config option - but I agree we should have more of a quick start guide in the readme/docs.

PR #235, opened 2018-02-01
Add a possibility to use LISTEN ONLY for the CAN Hardware.

### LINKED ISSUE #78: Exposing bus mode in consistent way
Originally reported by: **Brian Thorne (Bitbucket: [hardbyte](https://bitbucket.org/hardbyte), GitHub: [hardbyte](https://github.com/hardbyte))**

---

What possible modes do different bits of hardware support?

E.g. NORMAL, LOOPBACK, SILENT

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/78

mgiaco:
For a real logger it is very important that the logger device is not an active member. I think most of the can devices do support that. I have added some code for that but currently only the pcan is supported. This is a basic idea for that. 

I think i would be great to add some state to the base class. Maybe there is some other state needed in the future. I have tested the code with python 3.6 and python 2.7. 

Sorry the pull contains also the changes from the other pull.

Closes #78.


christiansandberg:
What other states are there? Do you need to change between passive and active in one bus instance? Maybe it's unnecessarily complicated and a `passive=True` in the constructor could cover most needs.

mgiaco:
I think 4 states are meaningful: INITIALISING, OPERATIONAL, LISTEN_ONLY, SLEEP. Initializing and sleep is not needed now but it would be great to have these states in the future I think. But at least operational and listen_only... or active and passiv is really important. 

Yes I have to change active and passive. Lets say you would do some auto baud in a app. So you do not know which baud rate is correct therefore you cold try to go from active to passive and switch the baud rate and try to receive some data. If you get some data the baud rate seems to be correct. And then If you would like to send some message from that baud rate you have to change from passive to active mode otherwise it is not possible to send.  So for me this feature could be really very useful. It is a good idea to add active as default.

mgiaco:
Any news here? What should I change?

can/bus.py
```diff
@@ -141,4 +143,13 @@ def shutdown(self):
         """
         self.flush_tx_buffer()
 
-    __metaclass__ = abc.ABCMeta
+    @property
+    @abc.abstractmethod
+    def state(self):
+        return 'Should never read this'
```
hardbyte: Instead of the unreachable code add a docstring - it will serve a dual purpose of making the method valid and documenting what the state will be for users/implementors.

can/bus.py
```diff
@@ -141,4 +143,13 @@ def shutdown(self):
...
+
+    @state.setter
+    def state(self, new_state):
+        pass
+
+    __metaclass__ = abc.ABCMeta
```
hardbyte: nitpick: missing newline

can/bus.py
```diff
@@ -9,7 +9,9 @@
 import logging
 import threading
 from can.broadcastmanager import ThreadBasedCyclicSendTask
+from collections import namedtuple
 
+BusState = namedtuple('BusState', 'INITIALISING, OPERATIONAL, LISTEN_ONLY, SLEEP')
```
hardbyte: As Christian commented earlier - can we use ACTIVE/PASSIVE.

Perhaps we should introduce an unknown and error state too - with unkown being the default for interfaces which haven't yet implemented it.

Also it seems like you are mixing bus mode and bus state. The bus mode is up to the caller, while the state is really up to the interface/hardware.

can/logger.py
```diff
@@ -53,10 +54,15 @@ def main():
     parser.add_argument('-b', '--bitrate', type=int,
                         help='''Bitrate to use for the CAN bus.''')
 
+    parser.add_argument('-s', '--state', type=int, dest='state', default=1,
```
hardbyte: Instead of the user having to look up or remember that `--state=1` is active and `--state=0` is passive this should instead be a flag (e.g. `--passive`)

examples/send_one.py
```diff
@@ -3,7 +3,10 @@
 
 
 def send_one():
-    bus = can.interface.Bus()
+    bus = can.interface.Bus(bustype='pcan', channel='PCAN_USBBUS1', bitrate=250000)
```
hardbyte: Try not to commit your configuration - the `can.interface.Bus()` will load from configuration files and environment variables by default.

hardbyte:
I think the bus mode should also be an optional argument to `Bus.__init__`.

felixdivo:
These changes should get documented as well, if that hasn't already been done.

can/bus.py
felixdivo: If we have many states, using an enum would be nice. But that would require [some library](https://pypi.python.org/pypi/enum34) for Python < 3.3.

examples/send_one.py
mgiaco: But this comes from the master branch so i will not change something here.

hardbyte: My bad!

can/bus.py
hardbyte: We do try to! But over many years with many developers things are hard to keep "normal".

mgiaco:
I have updated the pull

felixdivo:
I really like how it is implemented now. 👍 

But I'm not sure if we should make it an abstract property, since many interfaces might not implement this. How about providing a default implementation that locks the interface in `Active` and can be overridden in needed?

mgiaco:
I was also not sure if I should use an abstract property but I thought if I realize it that way it fits best to the current style of implementation.

felixdivo:
But supposing that some interface does not allow changing the bus state because the underlying system does not support it. How should the setter be implemented? Throw an exception really unexpectedly? Because silently failing and not changing anything would be quite a bad design in my opinion.

mgiaco:
Yes you are right not every hardware support active/passive mode. Hm, the changes I have made are not throw/raise any exception. 

I think if the interface does not support active/passive state it is better to do nothing and stay active than throw any exception. But that is my opinion @hardbyte what do you think about that?

felixdivo:
I my opinion, it should not silently fail, because the user intended to change something and should be notified that it did not work.

can/bus.py
```diff
@@ -146,4 +150,22 @@ def shutdown(self):
         """
         self.flush_tx_buffer()
 
+    @property
+    @abc.abstractmethod
```
felixdivo: You need to remove the `abc.` here. Shouldn't the setter be marked as abstract as well?

felixdivo:
@hardbyte What do you think?

mgiaco:
Is here an chance to get this mainline?

mgiaco:
What do you mean with mixing? I have updated my pull. I only use these states:

```
BusState = namedtuple('BusState', 'ACTIVE, PASSIVE, ERROR')
```
This is the normal behavior of a lot of devices in CAN. I have added ERROR but i haven´t used it now. Also till now only peak is supported but that was a first try. I will improve that in the future i need it for vector and ixxat to. So maybe if no one else contribute that I will try to add it for that devices too if i have time.
For me a CAN Hardware can only be in ACTIVE, PASSIVE and ERROR State. For me be in multiple States is not possible.

ERROR is not used at the moment. But i will try to add ERROR for peak in the future.

felixdivo:
Well, I want to get this merged too; thus this proposal:

1. I would make the default implementation of the setter raise a `NotImplementedError` to address the concerns mentioned @christiansandberg.

2. The default implementation should return `ACTIVE` so not all buses have to implement some behaviour they do not use anyways. But that would still make the API universal across many interface backends, which is sort of the whole point of this library. 

BTW: I have just resolved the merge conflict that arose from making the virtual bus interface thread safe.

hardbyte:
Sorry I dropped the ball on this one. Looks good - I'm glad you went ahead and merged @felixdivo  :-)

felixdivo:
Okay good, you're welcome. ;-)

PR #237, opened 2018-02-02
Added errors='replace' for system with other languages than English

alexalexein:
When the language of the system is set to French or Spanish, a decoding error is shown.


hardbyte:
@mgiaco does this resolve your issue also?

felixdivo:
This seems to be way cleaner than #238

mgiaco:
many thanks

PR #238, opened 2018-02-05
Decode error with primary language: Deutsch 

mgiaco:
Hi, i do get a decode error for "Deutsch". So I use that fix which also works for "English". I did not check the other languages but I think the fix works also for the others.

cheers


felixdivo:
Is this fixed by #237? You might add a test under `python-can/tests/` to test the mentioned fix and prevent regression later on.

mgiaco:
Hm yes it looks like it is fixed now.

Issue #243, opened 2018-02-10
test/simplecyclic_test.py often fails


felixdivo:
The fails can be seen on [Linux](https://travis-ci.org/hardbyte/python-can/jobs/339738832) and [OSX](https://travis-ci.org/hardbyte/python-can/jobs/339866694). It seems like that test works rather non-deterministic.

felixdivo:
`test/back2back_test.py::Back2BackTestCase::test_timestamp` might also be problematic. See [here](https://travis-ci.org/hardbyte/python-can/jobs/339885966).

These tests might fail on the CI server but not on our own machines since the CI servers may be slower and are virtualized.

felixdivo:
bb23d5b adds a better failure message.

hardbyte:
Maybe a better approach would be to skip the tests are running on travis

hardbyte:
How about using the `unittest.skipIf` decorator to always skip running these flaky tests _if_ on travis?

felixdivo:
`unittest.skipIf` would be the simplest and does the job I think.

PR #244, opened 2018-02-10
More test platforms, most notably OSX

felixdivo:
I added some more test platforms, most notably OSX tests for Python 3.6-dev, 3.7-dev and nighly. I also changed the Travis CI file to allow nightly Python builds to fail, but they are still tested and the results can be viewed. 

Now, there are generally more comments in `.travis.yml`. For someone reviewing this, I would kindly ask them to read the comments in there and decide whether the chosen approach for OSX support & disallowing/allwoing dev/nightly builds was OK.

For the failing tests, see #243. I think they are not directly connected to the changes in this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,12 @@ matrix:
       python: "3.7-dev"
     - os: osx
       python: "nightly"
+    - os: osx
+      python: "3.6-stable"
+    - os: osx
+      python: "3.6-dev"
+    - os: osx
+      python: "3.6-nightly"
   allow_failures:
     - python: "3.7-dev"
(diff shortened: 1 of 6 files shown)


felixdivo:
"Squash and merge" would probably be best for this since I needed many small commits to test the Travis CI script.

christiansandberg:
The default cyclic transmission is a bit problematic since it depends on available CPU resources to work reliably. Travis' OS X machines are probably choked.

felixdivo:
That's true, but it's not about their issue. The problem exists on Linux as well, as can be seen in #243.

PR #245, opened 2018-02-10
Adds canal interface

felixdivo:
This adds the changes that were requested in #161. That older PR can now be closed.


felixdivo:
The failing test is the problematic `test/simplecyclic_test.py`, see #243. It is not related to this PR.

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,196 @@
...
+        Flags to directly pass to open function of the CANAL abstraction layer.
+    """
+
+    def __init__(self, channel, *args, **kwargs):
+
+        # TODO force specifying dll
+        if 'dll' in kwargs:
+            dll = kwargs["dll"]
+        else:
+            raise Exception("please specify a CANAL dll to load, e.g. 'usb2can.dll'")
```
hardbyte: I think we should try getting the value from `can.rc` - does it make sense to have a default instead of raising an exception?

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,196 @@
...
+
+            if not deviceID:
+                raise can.CanError("Device ID could not be autodetected")
+
+        self.channel_info = "CANAL device " + deviceID
+
+        # set baudrate in kb/s from bitrate
+        # (eg:500000 bitrate must be 500)
+        if 'bitrate' in kwargs:
+            br = kwargs["bitrate"]
```
hardbyte: How about we fall back to `can.rc` and then a default
```
bitrate = kwargs.get('bitrate', can.rc.get('bitrate', 500))
```

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,196 @@
+# this interface is for windows only, otherwise use socketCAN
+
+import logging
+
+from can import BusABC, Message
+from can.interfaces.canal.canal_wrapper import *
+
+bootTimeEpoch = 0
+try:
+    import uptime
```
hardbyte: This kind of optional dependency should be documented in the interface docs

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,196 @@
...
+        used.
+
+    :param str dll:
+        dll with CANAL API to load
+
+    :param int bitrate:
+        Bitrate of channel in bit/s. Values will be limited to a maximum of 1000 Kb/s.
+        Default is 500 Kbs
+
+    :param str serial (optional)
```
hardbyte: Have a look at how this renders in sphinx - I suspect this won't register as a parameter without a trailing `:`

can/interfaces/canal/serial_selector.py
```diff
@@ -0,0 +1,39 @@
...
+    for objItem in colItems:
+        string = objItem.Dependent
+        # find based on beginning of serial
+        if serialMatcher in string:
+            # print "Dependent:" + ` objItem.Dependent`
+            string = string[len(string) - 9:len(string) - 1]
+
+            return string
+
+    return None
```
hardbyte: nitpick: missing newline

can/interfaces/socketcan/socketcan_ctypes.py
```diff
@@ -125,12 +125,16 @@ def recv(self, timeout=None):
 
     def send(self, msg, timeout=None):
         frame = _build_can_frame(msg)
+
         if timeout:
             # Wait for write availability. write will fail below on timeout
-            select.select([], [self.socket], [], timeout)
+            _, ready_send_sockets, _ = select.select([], [self.socket], [], timeout)
```
hardbyte: This shouldn't be part of this PR

examples/canal_demo.py
```diff
@@ -0,0 +1,50 @@
...
+# alternative: specify device serial:
+# bus = can.interface.Bus(bustype="canal", dll="canusbdrv64.dll", serial = "LW19ZSBR", bitrate=100000, flags=4)
+
+def send_one():
+    msg = can.Message(arbitration_id=0x00c0ffee,
+                      data=[0, 25, 0, 1, 3, 1, 4, 1],
+                      extended_id=True)
+    try:
+        bus.send(msg)
+        print ("Message sent:")
```
hardbyte: I suggest adding a `from future import print` and removing the space after `print` here

felixdivo:
All requested changes are valid. ;) I will address them as soon as I can, but that will have to wait for one or too weeks probably, since I am very busy right now.

can/interfaces/canal/canalInterface.py
felixdivo: I now tries to get it from `kwargs` then from `can.rc` and then fails. I don't know where to look for that dll typically, do you?

can/interfaces/canal/canalInterface.py
felixdivo: Actually this can be completely removed since I was not able to find any usage of  `bootTimeEpoch`.

can/interfaces/canal/serial_selector.py
felixdivo: Now I learned the word `nitpick`. ^^ Done.

can/interfaces/socketcan/socketcan_ctypes.py
felixdivo: Yeah right, I must've started with the wrong branch or something. I would just leave it in this time. Ok?

can/interfaces/canal/canalInterface.py
felixdivo: I guess that is now solved.

felixdivo:
> I'm unlikely to merge this without a page added to the documentation.
> 
> It wouldn't need to cover much - links to the hardware vendor, mention and link to the required libraries. > Any comments on compatibility, limitations etc.

I would like someone else to do that since I do not really get Sphinx and do not know the hardware behind the interface at all. Can someone else pick that up?

felixdivo:
Does it need to be registered in `can/interface.py :: BACKENDS`?

felixdivo:
@krixkrix Could you write some docs? That would be really nice.

felixdivo:
This still needs to be done:

- [ ] register this interface in `can/interface.py :: BACKENDS`? Is this required?
- [ ] add a page to the documentation. It wouldn't need to cover much - links to the hardware vendor, mention and link to the required libraries. Any comments on compatibility, limitations etc.

Can anyone help to close this ancient issue?

hardbyte:
@krixkrix while I appreciate that you originally opened your PR a full year ago - do you have any time over the next month to help wrap this up so it can be part of python-can from version 2.2?

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,185 @@
...
+                raise can.CanError("Device ID could not be autodetected")
+
+        self.channel_info = "CANAL device " + deviceID
+
+        # get baudrate in bit/s
+        baudrate = kwargs.get('bitrate', can.rc.get('bitrate', 500000))
+        # convert to kb/s
+        baudrate = baudrate // 1000
+        # cap: max rate is 1000 kb/s
+        baudrate = max(baudrate, 1000)
```
christiansandberg: Should use `min`.

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,185 @@
...
+            return message_convert_rx(messagerx)
+
+        elif status == 19:
+            # CANAL_ERROR_RCV_EMPTY
+            log.warn("recv: status: CANAL_ERROR_RCV_EMPTY")
+            return None
+
+        elif status == 32:
+            # CANAL_ERROR_TIMEOUT
+            log.warn("recv: status: CANAL_ERROR_TIMEOUT")
```
christiansandberg: Timeout or empty recv queue should be normal so no warning should be logged.

can/interfaces/canal/canalInterface.py
```diff
@@ -0,0 +1,185 @@
...
+    config = deviceID + '; ' + baudrate
+    """
+    return "%s; %s" % (deviceID, baudrate)
+
+
+def message_convert_tx(msg):
+    messagetx = CanalMsg()
+
+    length = len(msg.data)
+    messagetx.sizeData = length
```
christiansandberg: `msg.dlc` maybe? Otherwise remote frames might not work.

felixdivo:
I think docs is the only thing missing, apart from tests.

felixdivo:
Can someone test this? I do not have the required hardware at hand.

felixdivo:
What shall we do with this PR? Keep it lying around here forever?

krixkrix:
Sorry for not responding earlier. I am not really using this lib anymore, and I am not really sure I can find the time to dig into it again. But I do have the hardware so maybe if all that is needed is doc and test?

felixdivo:
Cool! Yes, I think all that is needed now is:

- **documentation** similar to these: [python-can/doc/interfaces/](https://github.com/hardbyte/python-can/tree/develop/doc/interfaces). It should tell at least: The supported hardware and the required drivers. Great would be the limitations of the interface as well, like whether it supports error frames or CAN FD. It would be great if you could do that, maybe simplay add a comment below.
- **testing**: Like sending, receiving with normal frames and error frames or remote frames. And testing the automatic detection of available channels/devices. It would be great if you could do that. 
- Well, fixing something is the library if we find any problems. I could probably do that for you.

felixdivo:
@krixkrix Just mentioning you because I'm not sure if you else get notified by new comments. See above for my comment on docs & tests.

felixdivo:
@krixkrix Have you already found time to look at this? As version 3.0 is now released, I want to get this working early in this "iteration" so we can finally release this into 3.1.

acolomb:
What is the difference between this backend and the USB2CAN one?  I do have hardware for the latter and it also uses an abstraction layer DLL named CANAL.  If it's in any way related, I could do some testing.

Would be good to have at least **some** kind of documentation, to answer questions like this ;-)

felixdivo:
This might be the case, yeah. I actually have no idea of the required DLLs, I only copied PR #161 and applied some general patches. Also, I found this through the issue #465: https://github.com/krumboeck/usb2can_canal.

If this is redundant, we can close this PR, sure. Apparently, no one had interest to solve this, so that might explain it.

acolomb:
So the real question is **why** did @krixkrix make a copy of the usb2can backend?  I don't get it from the [comment](https://github.com/hardbyte/python-can/pull/161#issue-119646540), but didn't check the source code either.

@felixdivo Would you mind diffing the two backends against each other, to figure out the main differences?

felixdivo:
Yeah, I could do that in a few days/weeks. I'm quite busy right now.

acolomb:
According to the [VSCP documentation](https://grodansparadis.gitbooks.io/the-vscp-daemon/level_i_drivers.html#vscp-daemon-level-i-drivers):

> For Python developers python-can is a good tool. Unfortunately the CANAL interface is named USB2CAN but it is there.

Apparently the CANAL API is implemented by a bunch of drivers there, not only USB2CAN.  But I wouldn't bet on more people using this interface backend if it were named CANAL, especially since probably no one ever tested with other CANAL implementation DLLs.

felixdivo:
> But I wouldn't bet on more people using this interface backend if it were named CANAL.

I wouldn't either.

acolomb:
@felixdivo I've started a comparison of the backends, see [this Gist](https://gist.github.com/acolomb/945cc5e5b29ced411f1df5b5f9c4a4d0).

Besides renaming the files and classes, there are some differences in error handling (exceptions) and fewer comments about the backend being targeted to Windows mainly.  In some places, I get the impression that some general python-can changes, like `msg.is_extended_id` vs. `msg.id_type` have not been incorporated properly into the current PR state.

My knowledge of python-can internals is very limited though, so I won't try to judge which backend is "better".  Some obvious documentation and cosmetic changes could be ported to usb2can easily, though.

felixdivo:
I didn't know Github candisplay .diff files that nicely. ^^

Yeah, there are some changes that should be ported to `usb2can` to replace this interface.

I can do it this weekend, but I have a question about `serial_selector.py`'s `find_serial()`: Is `serialMatcher = "PID_6001"` a sensible default, or is `"ED"` better? I don't know how they usually look like, so ...

acolomb:
The USB2CAN adapters from 8devices apparently always have a serial number starting with 'ED' prefix.  I don't know what the resulting strings from the Windows API query will look like, but it seems to slice out the last 8 bytes trying to get the serial number.  On the other hand, the USB product ID [PID_6001](https://www.the-sz.com/products/usbid/index.php?p=0x6001) probably matches an FTDI UART interface chip on the adapter, which will result in pretty much any USB developer gadget without an own vendor id, plus the ones in the linked query.

Since the method only returns the first match, the latter might cause problems.  It's a rather fragile approach either way. Looking for 'ED' will probably only work for adapters sold by **E**ight **D**evices, not @rusoku's TouCAN, for example.

Since `_detect_available_configs()` is never called with an argument, it's also not easy to override the selection substring.

I'd probably go with the more flexible approach from `canal/`, but keep the default prefix at 'ED' until someone complains.

rusoku:
"ED" prefix comes from my old company edevices.lt ;-)
usb2can USB PID is 1234 and not 6001.
BTW TouCAN converter has different init string format like: 0 ; 12345678; 1000.
https://www.rusoku.com/storage/app/media/CANAL%20Library%20User%20Guide.pdf
Added symbol before serial number which means device type: 0 - TouCAN, TouCAN Marine, TouCAN micro, 1 - TouCAN duo, 2 - TouCAN FD etc...
filter/mask values now moved from init string to its own extended functions.
All rusoku devices will have 8 symbols serial number.

felixdivo:
Thanks for all the input. ;)

Maybe the Win32 API can even reveal the serial numbers directly? Does anyone know the API well enough for that? I couldn't find it via the first search results ...

> BTW TouCAN converter has different init string format like: 0 ; 12345678; 1000

Then we could simply look for the last regex group with `(\d+)` and take that as the serial number. Is it safe to assume that it's always in decimal notation?

>Added symbol before serial number which means device type

Do we need to strip the device type number from the serial ID or is it okay to leave it in there? Because how can we detect when to strip it and when not?

> I'd probably go with the more flexible approach from `canal/`, but keep the default prefix at 'ED' until someone complains.

Sounds good, combined with variable length IDs.

rusoku:
> > BTW TouCAN converter has different init string format like: 0 ; 12345678; 1000
> 
> Then we could simply look for the last regex group with `(\d+)` and take that as the serial number. Is it safe to assume that it's always in decimal notation?
> 
> > Added symbol before serial number which means device type
> 
> Do we need to strip the device type number from the serial ID or is it okay to leave it in there? Because how can we detect when to strip it and when not?

For TouCAN devices init string always consists of 3 blocks: **device_id ; serial number ; speed.**
Or for custom speed : **device_id; serial_number ; 0 ; tseg1; tseg2; sjw; brp**
serial number is always 8 decimal characters

PR #247, opened 2018-02-11
Fixing #217 & #218

felixdivo:
Fix #217. Fix #218.

Depends on PR #245 & PR #246.

PR #246, opened 2018-02-11
Fixing #196

felixdivo:
Fix #196.

EDIT: I wrote in the name of the branch and in the initial title something about issue 192, which was a typo. This is all about issue #196.

Depends on PR #245.

PR #248, opened 2018-02-11
Fix usb2can issue #219

### LINKED ISSUE #219: usb2can with Python3 dead (fix included)
Hi,
I found a couple of show-stoppers with usb2can using Python3.
Here's my patch:
```
        # set flags on the connection
        if 'flags' in kwargs:
            enable_flags = kwargs["flags"]

        else:
            enable_flags = 0x00000008

        # code to get the serial number of the device
        if 'serial' in kwargs:
            deviceID = kwargs["serial"]
        elif channel is not None:
            deviceID = channel
        else:
            from can.interfaces.usb2can.serial_selector import serial
            deviceID = serial()

        # set baudrate in kb/s from bitrate
        # (eg:500000 bitrate must be 500)
        if 'bitrate' in kwargs:
            br = kwargs["bitrate"]

            # max rate is 1000 kbps
            baudrate = min(1000, int(br/1000))   ########## <------- This was max(...)
        # set default value
        else:
            baudrate = 500

        connector = format_connection_string(deviceID, baudrate)

        self.handle = self.can.open(connector.encode(), enable_flags)   ########### <------- here .encode() was missing
```
Apparently in Python 3 ctypes strings must be passed as bytearray, I don't know if this breaks compatibility with Python 2 though. The above works for me.
Oh, and the max() was clearly wrong.

felixdivo:
Fixes #219.

It would be nice if someone could test this. 

Depends on #245, #246 and #247. (I seemed to have failed to branch off from the `develop` branch, so the PRs depend on each other for this time.)

### DIFF
--- a/can/interface.py
+++ b/can/interface.py
@@ -21,7 +21,8 @@ BACKENDS = {
     'virtual':          ('can.interfaces.virtual', 'VirtualBus'),
     'neovi':            ('can.interfaces.ics_neovi', 'NeoViBus'),
     'vector':           ('can.interfaces.vector', 'VectorBus'),
-    'slcan':            ('can.interfaces.slcan', 'slcanBus')
+    'slcan':            ('can.interfaces.slcan', 'slcanBus'),
+    'canal':            ('can.interfaces.canal', 'CanalBus'),
 }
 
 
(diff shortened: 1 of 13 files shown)


hardbyte:
Cheers I'll rebase and reopen.

PR #250, opened 2018-02-13
Adding neovi device serial base 36 decoding when needed

pierreluctg:
ICS NeoVi device serial above `A0000` are base36 encoded and need to be decoded properly.


pierreluctg:
@hardbyte can we include this on in the 2.1.0 release?

Thank you

PR #255, opened 2018-02-17
Handle error frame messages in CanutilsLogWriter and CanutilsLogReader.

### LINKED ISSUE #217: CanutilsLogWriter and CanutilsLogReader fail when working with error frame messages
The test cases should be adjusted after this is fixed. See #214.

hardbyte:
Closes #217 (cherry picked from commit 61f27e1)

Replaces https://github.com/hardbyte/python-can/pull/247


felixdivo:
Oh no, this seems to be a problem with timestamps again ...
I will have a look at it as soon as I can, which might be a few days from now.

felixdivo:
Oh wow, now the tests pass. ^^ I will still patch it to not occur again if there is still a problem ...

felixdivo:
Okay, so as I understand it, the tests above fail (and should always fail) because of a legitimate reason. The problem resides in `can.io.asc.py :: ASCWriter.log_event(...)`. [Link](https://github.com/hardbyte/python-can/blob/012768358daa29c29f84ca52c63f121bca10ac1e/can/io/asc.py#L131). When a message with `timestamp=None` is fed to `on_message_received` (which calls `log_event`), the current `time.time()` is written to the file in the line above. All other reader / writer pairs would write zero in that case, which makes more sense for me. So in order to make it consistent with the others, I changed it to not "cleverly" take the current timestamp, but rather write a plain zero. It that okay?

hardbyte:
That sounds like a good idea! Do you want to try rebase this - or apply the patch to a recent develop?

christiansandberg:
Just make sure that the time for each entry is the same or later than the last. It would be strange if they were not in the right order. In BLFWriter, if a timestamp is not given, it gets the same timestamp as the last one instead.

felixdivo:
@christiansandberg I'm not sure if that is desirable. Because in my opinion the Writers should simply write the same information that is present in the message and the Readers should be able to reconstruct them exactly like they were before. That is currently not tested and I think not documented anywhere. So do we really want to make assumptions about the user's application? Because if we wanted to, we could

- set time.time() as the timestamp of the first message if it does not have one
- add the last known timestamp if the current message doesn't have one
- set the last known timestamp if the current message has a smaller timestamp than the previous one

What of that do we want to implement and what not?

can/CAN.py
```diff
@@ -1,9 +1,13 @@
+#!/usr/bin/env python
+# coding: utf-8
```
christiansandberg: The shebang is only useful if the file can be executed as a script. Most files here are modules so it doesn't add any value.

I don't think there are any non-ASCII characters in these files so the UTF-8 should not be needed either. Besides, UTF-8 is default in Python 3.

can/interfaces/iscan.py
```diff
@@ -148,11 +152,11 @@ class IscanError(CanError):
 
     def __init__(self, function, error_code, arguments):
         super(IscanError, self).__init__()
-        #: Status code
```
christiansandberg: Why was this changed? See [Sphinx documentation](http://www.sphinx-doc.org/en/master/ext/autodoc.html#directive-autofunction).

can/interfaces/pcan/pcan.py
```diff
@@ -82,7 +86,7 @@ def __init__(self, channel, *args, **kwargs):
         else:
             self.channel_info = channel
 
-        bitrate = kwargs.get('bitrate', 500000)
+        bitrate = kwargs.get('bitrate', can.rc.get('bitrate', 500000))
```
christiansandberg: This should be solved on a higher level instead so that it works for all interfaces.

can/CAN.py
felixdivo: These changes originate from a different PR that this one was rebased on, since it required changes that were made over there. That is also the reason for why there are so many file that were changed. 

I think the shebang does add value as it make clear that it is compatible with Python 2 & 3 as well as being uniform. But I guess that's subjective. It doesn't do any harm however.

The `coding` line does add value as well, since it enforces the same behavior on Python 2 & 3. That is desirable since everything that makes these versions more alike in such manners is welcome for maintainability. It also makes it explicit.

can/interfaces/iscan.py
felixdivo: Sorry, I misunderstood that. I will revert that.

can/interfaces/pcan/pcan.py
felixdivo: You are right. I removed it.

felixdivo:
@christiansandberg Hey, I think there was a misunderstanding. I wanted to ask if you could comment on my design question above concerning the writers in the `can.io.*` module. But thank you very much for your review!

christiansandberg:
If you could rebase on top of develop it would be easier to find the actual changes.

felixdivo:
Holy f*ck! That made it worse. I simply rebased on top `develop`.

felixdivo:
Oh, you can see the actual changes [here](https://github.com/hardbyte/python-can/compare/develop...fix-217?expand=1)! I hope that helps ...

can/io/asc.py
```diff
@@ -139,18 +150,18 @@ def on_message_received(self, msg):
...
-        timestamp = msg.timestamp
-        if timestamp >= self.started:
-            timestamp -= self.started
+        if msg.is_extended_id:
+            arb_id += "x"
 
         channel = msg.channel if isinstance(msg.channel, int) else self.channel
-        line = self.LOG_STRING.format(time=timestamp,
+
+        line = self.LOG_STRING.format(time=msg.timestamp,
```
christiansandberg: [Here](https://bitbucket.org/tobylorenz/vector_asc/src/47556e1a6d32c859224ca62d075e1efcc67fa690/src/Vector/ASC/tests/unittests/data/CAN_Log_Trigger_3_2.asc?at=master&fileviewer=file-view-default) is an example of an ASC file. There is not much info on this format but my interpretation is that `Begin Triggerblock Mon Mar 7 01:21:51 pm 2005` defines the start of measurement and the timestamps are relative to this time.

christiansandberg:
I'm thinking it would be better to wait with writing the header until the first message is received and use that timestamp instead of the current time. This would make the result more reproducible and allow for converting formats in the future.

I still think a `log_event` call without a timestamp is better off with the last logged timestamp instead of 0 or `time.time()`.

felixdivo:
I now changed `CanutilsLogWriter` and `ASCWriter` to behave more similar in 79d8d38 & d1a6b16. This is how they now behave:

> __ASCWriter__
    The measurement starts with the timestamp of the first registered message.
    If a message has a timestamp smaller than the previous one (or 0 or None),
    it gets assigned the timestamp that was written for the last message.
    It the first message does not have a timestamp, it is set to zero.

> __CanutilsLogWriter__
    If a message has a timestamp smaller than the previous one (or 0 or None),
    it gets assigned the timestamp that was written for the last message.
    It the first message does not have a timestamp, it is set to zero.

Shall I merge it like that?

can/io/asc.py
```diff
@@ -120,37 +143,51 @@ def log_event(self, message, timestamp=None):
             logger.debug("ASCWriter: ignoring empty message")
             return
 
-        timestamp = (timestamp or time.time())
+        # this is the case for the very first message:
+        if not self.header_written:
+            self.last_timestamp = (timestamp or 0.0)
+            self.started = self.last_timestamp
+            self.log_file.write("Begin Triggerblock %s\n" % self.last_timestamp)
```
christiansandberg: Timestamp should be converted to the format "%a %b %m %I:%M:%S %p %Y".

felixdivo:
I also added `base hex  timestamps absolute\n` to the header of each ASC file, like over [here](https://bitbucket.org/tobylorenz/vector_asc/src/47556e1a6d32c859224ca62d075e1efcc67fa690/src/Vector/ASC/tests/unittests/data/CAN_Log_Trigger_3_2.asc?at=master&fileviewer=file-view-default).

test/logformats_test.py
```diff
@@ -38,20 +38,25 @@
...
-    :param check_error_frames: if true, also tests error frames
-    :param check_comments: if true, also inserts comments at some
+    :param bool check_remote_frames: if True, also tests remote frames
+    :param bool check_error_frames: if True, also tests error frames
+    :param bool check_comments: if True, also inserts comments at some
         locations and checks if they are contained anywhere literally
         in the resulting file. The locations as selected randomly
         but deterministically, which makes the test reproducible.
+    :param bool round_timestamps: if True, rounds timestamps using :meth:`~builtin.round`
+        before comparing the read messages/events
```
christiansandberg: Maybe it's better to specify number of decimal places to match against? ASC files only store 4 decimals currently.

felixdivo: Well, I think the problem comes from the timestamps being relative to the start of the measurement, but that start timestamp is being serialized without sub-second precision. One failing test case failed because of something like `1000000.197 != 1000000.2`. So I guess rounding is more appropriate *in the case of ASC files*.

felixdivo: @christiansandberg Do you agree with that? Shall we merge it?

christiansandberg: Right, didn’t think of that. Go ahead and merge.

PR #258, opened 2018-02-19
Vector improvements

christiansandberg:
* Avoid polling by using Win32 events when possible
* Do not warn when no init access
* Include function name in error messages
* Support for receive_own_messages argument


felixdivo:
Did you document in the `docs/*` that `pywin32` (or the builtin Win32 module) is used when available to speed things up?

PR #259, opened 2018-02-20
General cleanups of all python code

felixdivo:
Now every file begins with a [shebang](https://www.python.org/dev/peps/pep-0394/#recommendation), a [file encoding comment](https://www.python.org/dev/peps/pep-0263/#defining-the-encoding), a module docstring (might be empty when I had no clear idea what some code was doing) and the imports are sorted from `__future__` over Python standard modules (like `os` or `ctypes`) to `python-can` imports:

```Python
#!/usr/bin/env python
# coding: utf-8

"""
can is an object-orient Controller Area Network interface module.
"""

import os
import logging

from can.message import Message
```

And I made many other small changes. The goals was to unify the code at least a little bit. 

**EDIT:** Basically I touched everything except `docs/*`.


can/interfaces/kvaser/argument_parser.py
```diff
@@ -1,3 +1,9 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+"""
+TODO: Where is this used? Is this used?
```
hardbyte: Not used. Feel free to remove this file.

can/interfaces/pcan/PCANBasic.py
```diff
@@ -287,36 +269,29 @@
...
     Represents a PCAN message
     """
-    _fields_ = [ ("ID",      c_uint),          # 11/29-bit message identifier
+    _fields_ = [ ("ID",      c_uint),           # 11/29-bit message identifier
                  ("MSGTYPE", TPCANMessageType), # Type of the message
                  ("LEN",     c_ubyte),          # Data Length Code of the message (0..8)
                  ("DATA",    c_ubyte * 8) ]     # Data of the message (DATA[0]..DATA[7])
 
-# Represents a timestamp of a received PCAN message
-# Total Microseconds = micros + 1000 * millis + 0x100000000 * 1000 * millis_overflow
```
hardbyte: You've deleted a lot of comments from this file.

can/interfaces/pcan/PCANBasic.py
```diff
@@ -661,9 +611,6 @@ def GetValue(
             logger.error("Exception on PCANBasic.GetValue")
             raise            
 
-    # Returns a descriptive text of a given TPCANStatus
-    # error code, in any desired language
```
hardbyte: How will the developer know this without the comment? I suggest either reverting or moving some of the comments into proper python docstrings.

can/interfaces/usb2can/serial_selector.py
```diff
@@ -16,21 +23,21 @@ def WMIDateStringToDate(dtmDate):
...
-        strDateTime = strDateTime + dtmDate[0] + dtmDate[1] + dtmDate[2] + dtmDate[3] + " " + dtmDate[8] + dtmDate[
-            9] + ":" + dtmDate[10] + dtmDate[11] + ':' + dtmDate[12] + dtmDate[13]
+        strDateTime = strDateTime + dtmDate[0] + dtmDate[1] + dtmDate[2] + dtmDate[3] + ' ' + dtmDate[8] + dtmDate[
+            9] + ':' + dtmDate[10] + dtmDate[11] + ':' + dtmDate[12] + dtmDate[13]
     return strDateTime
 
 
 def serial():
-    strComputer = "."
+    strComputer = '.'
```
hardbyte: Just because?

can/io/sqlite.py
```diff
@@ -64,6 +67,7 @@ def close(self):
 
 
 # Backward compatibility
+# TODO remove in later releases?
```
hardbyte: Yeah we can probably add a future deprecation warning for a release or two.

examples/send_one.py
```diff
@@ -1,20 +1,31 @@
...
+This example shows how sending a single message works.
+"""
+
 from __future__ import print_function
-import can
 
+import can
 
 def send_one():
     bus = can.interface.Bus(bustype='pcan', channel='PCAN_USBBUS1', bitrate=250000)
```
hardbyte: There should also be a `bus = can.interface.Bus()` version with a comment refering to env/file config

can/interfaces/kvaser/argument_parser.py
felixdivo: It is now gone.

can/interfaces/pcan/PCANBasic.py
felixdivo: Yes, I did. But the comments beginning with `#` above each method were simply the first line of the actual method doc string, so I removed the redundancy. They were simply a duplicate and distracting.

can/interfaces/pcan/PCANBasic.py
felixdivo: Actually, it already is in the doc string. ;-)

can/interfaces/usb2can/serial_selector.py
felixdivo: I simply followed [this](https://stackoverflow.com/a/56190/3753684) "rule" because I personally like it and it was used in many other parts of this library as well (AFAIK).

can/io/sqlite.py
felixdivo: Uhm, any ideas on how to do that?

examples/send_one.py
felixdivo: Yup. See the last commit below.

can/io/sqlite.py
hardbyte: One way is to use the [deprecated](https://github.com/tantale/deprecated) library

PR #263, opened 2018-02-22
add  a deprecated notice to 'SqlReader' for users

felixdivo:
I changed the structure of `setup.py` a little bit and added the [Deprecated](https://github.com/tantale/deprecated) library as a dependency, as suggested by Brian. Then I added the `@deprecated` decorator to mark SqlReader as - you guessed it - deprecated.

### DIFF
--- a/can/io/sqlite.py
+++ b/can/io/sqlite.py
@@ -52,7 +52,7 @@ class SqlReader:
     def __iter__(self):
         log.debug("Iterating through messages from sql db")
         for frame_data in self.cursor.execute(self._SELECT_ALL_COMMAND):
-            yield SqlReader._create_frame_from_db_tuple(frame_data)
+            yield self._create_frame_from_db_tuple(frame_data)
 
     def __len__(self):
         # this might not run in constant time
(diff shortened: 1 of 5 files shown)


felixdivo:
Oh wow, the Deprecated library is strange. Anyways, it seems like we forgot to include `SqlReader` in `can/__init__.py` and `can/io/__init__.py`, so it already broke comparability in the last Release. That makes all of this obsolete, right?

can/io/sqlite.py
```diff
@@ -65,10 +68,8 @@ def close(self):
         """Closes the connection to the database."""
         self.conn.close()
 
-
-# Backward compatibility
-# TODO remove in later releases?
-SqlReader = SqliteReader
+# SqliteReader is the newer name
+SqliteReader = SqlReader
```
hardbyte: Won't the `SqliteReader` also be marked as deprecated with this approach?

felixdivo: Yes yes, it all doesn't work like I intended. But de we really need this deprecation at all? We already broke compatibility.

hardbyte: Hmm I didn't realize we'd already broken compatibility - in that case no we don't really need this deprecation warning. Lets just close this PR and remove the SqlReader/SqliteReader alias.

felixdivo:
Yeah, I will do that.

PR #264, opened 2018-02-23
Add CI testing on Appveyor (Windows)

### LINKED ISSUE #260: Add testing under Windows
Currently our code is automatically tested under Linux and many different Python versions, and under OSX under only two Python 3 versions using Travis CI. (see `python-can/.travis.yml`) 

But no automatic testing under Windows is performed. Should we integrate it into the testing of python-can? Unfortunately Travis CI can't do that, but maybe someone else knows a similar service for Windows.

felixdivo:
This adds CI tests on Windows. I also factored out common skipping behavior (`@skipIf(IS_TRAVIS, "...")`) and replaced it with generic CI skipping behavior (`@skipIf(IS_CI, "...")`). The constants are defined in `test/config.py` and define some properties in a central place:

- IS_TRAVIS
- IS_APPVEYOR
- IS_CI
- TEST_CAN_FD
- More? Maybe whether to test on hardware, but is that the right place for stuff like that? Your thoughts?

Closes #260.

Still work in progress.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,45 @@
+environment:
+
+  matrix:
+
+    # For Python versions available on Appveyor, see
+    # http://www.appveyor.com/docs/installed-software#python
+
+    - PYTHON: "C:\\Python27"
+    - PYTHON: "C:\\Python33"
+    - PYTHON: "C:\\Python34"
+    - PYTHON: "C:\\Python35"
+    - PYTHON: "C:\\Python36"
+    - PYTHON: "C:\\Python27-x64"
+    - PYTHON: "C:\\Python33-x64"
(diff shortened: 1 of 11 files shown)


felixdivo:
Done.

hardbyte:
This is great - thanks! Just a note about merging your own PRs, I'd prefer that we have someone else review almost all changes into develop - even if this slows changes down a bit. This PR is an example of a small enough change where it is okay to skip that (which is why I didn't make it mandatory in the settings) but I thought I'd just mention it anyways. :-)

felixdivo:
Yes, you are right. That's why #266 or #255 are still waiting to be commented on. I try to only merge non-controversial stuff like this by myself. ;)

PR #266, opened 2018-02-23
can.Notifier enhancements

### LINKED ISSUE #265: Create new function to add a listener to the notifier class
Could you please add a new function to the notifier class as such:

    def add_listener(self, listener):
        """Add new Listener to be notified 

        :param listener: new listener object to be added to the list to be notified
        """
        self.listeners.append(listener)

Thanks

felixdivo:
This PR addresses #265. Basically adds a `add_listener` and a `remove_listener` method to `can.Notifier`. 

Closes #265.


hardbyte:
Could you please add a couple of basic tests in https://github.com/hardbyte/python-can/blob/develop/test/listener_test.py

felixdivo:
Done. Shall I merge it into `develop`?

Issue #274, opened 2018-02-28
SocketCAN fails on Python versions 3.3 & 3.4 and on PyPy 3


felixdivo:
This can be found in [the Python 3.5 change log](https://docs.python.org/3/whatsnew/3.5.html#changes-in-the-python-api) and the [official docs](https://docs.python.org/3/library/socket.html#socket.CAN_RAW_FD_FRAMES):

> The socket module now exports the CAN_RAW_FD_FRAMES constant on linux 3.6 and greater.

If you look at this [CI test](https://travis-ci.org/hardbyte/python-can/builds/347338694), socketcan-native fails for Python versions 3.3 & 3.4 because it uses CAN_RAW_FD_FRAMES on these platforms. How can a interface register itself as not being able to handle CAN FD frames?

There is also another Problem on PyPy 3: `module 'socket' has no attribute 'CAN_RAW'`.

felixdivo:
Do we even want to support these Python versions any more? At least Python 3.3 has reached its [end of life](https://devguide.python.org/#status-of-python-branches).

felixdivo:
See the bottom of `test/back2back_test.py` on branch `more-tests` (or maybe `develop` once it is merged).

PR #275, opened 2018-03-02
Added proof-of-concept testing of SocketCAN on Travis CI & Smaller changes

felixdivo:
The testing can be found in `test/back2back_test.py` but is skipped for now, since it fails on Python 3.3/3.4. See #274. I expanded the `test/config.py` file to contain information about the platform as well. More serious testing remains to be implemented.  

I also made small changes to some `can.io.*` classes and removed the overall repeated logging (like multiple logging statements for each sent message) quite a bit. I also added some tests to `listener_test.py`.


can/io/player.py
```diff
@@ -44,7 +44,9 @@ class LogReader(object):
 
     @classmethod
     def __new__(cls, other, filename):
-        if filename.endswith(".asc"):
+        if not filename:
+            TypeError("a filename must be given")
```
hardbyte: missing `raise`

PR #276, opened 2018-03-05
Remove `load_default_settings` call in init

pierreluctg:
`load_default_settings` is not supported in the open source version of icsneoapi


felixdivo:
Was this changed in the library or did it get tested by you or even some unit tests?

pierreluctg:
@felixdivo nothing changed in the library itself. The thing is that Intrepid (ICS) do provide pre-built drivers and open source driver source. The open source driver do not have all the non essential API implemented, like `load_default_settings`.

felixdivo:
Ah, I see. But couldn't we try to call it and catch the AttributeError (or whatever it os) if it's not available?

Issue #51, opened 2016-03-20
Provide list of presently available buses


hardbyte:
Originally reported by: **Kyle Altendorf (Bitbucket: [altendky](https://bitbucket.org/altendky), GitHub: [altendky](https://github.com/altendky))**

---

I just created a socketcan/PCAN only bus checker.  It tries to create a bus for each interface/channel and checks for exceptions to filter out the unavailable ones.

https://github.com/altendky/st/blob/7c87276d1bb7b58e8d9c060b40d047fed5a5032d/epyq/busselector.py#L19

It's pretty hacky and hardcoded but I thought I would share it in case anyone wanted to expand on it and integrate it into python-can.

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/51

cowo78:
IMHO we can think about extending the BusABC interface with a detect_hardware static or something like that.
It should report configuration dicts and the available channels for each detected piece of hardware.

felixdivo:
The canal interface already implements that automatic discovery, but as a fallback when no channel was given in `kwargs` or `can.rc`. See #245.

felixdivo:
Let's suppose `BusABC` has a detect method, like Giuseppe proposed. I would call it `detect_available_channels` since not all interfaced work with actual hardware. For example `virtual` or `socketcan`'s `vcan0` only work in software. It could look like this (notice that it is not abstract):

```Python
# in BusABC:
@staticmethod
def _detect_available_channels():
    """Docstring."""
    raise NotImplementedError()
```

Then I would define this method and make it accessible directly from the `can` module:

```Python
# in can/interface.py

from __future__ import absolute_import

import sys
import logging

from .interface import BACKENDS

log = logging.getLogger('can.interface')

if sys.version_info.major > 2:
    basestring = str

def detect_available_channels(search_only_in=None):
    """Dostring.

    :param search_only_in: either
        - the name of an interface as a string,
        - an iterable of interface names, or
        - `None` to search in all known interfaces.
    """

    if search_only_in is None:
        # use an iterator over the keys so we do not have to copy it
        search_only_in = BACKENDS.iterkeys()
    elif isinstance(search_only_in, basestring):
        search_only_in = (search_only_in,)
    # else it is an iterable of strings

    result = []
    for interface in search_only_in:

        # use the same code as in can/interface.py :: Bus.__new__()
        bus_class = get_class_of_this(interface)

        # get available channles
        try:
            result += bus_class._detect_available_channels()
[...]

PR #277, opened 2018-03-06
Software filtering for all Buses

### LINKED ISSUE #104: usb2can doesn't implement filters
Ideally the usb2can interface implements the `set_filters` method.

### LINKED ISSUE #157: Software filtering of messages

felixdivo:
This basically implements #157 (with some small changes). 

The `BusABC` class is now finished, and I want to ask whether that is looking good. If it is, I will adjust the other interfaces one by one to make it work with this new frontend. It shouldn't be too much work though. It would be nice if someone could comment on the docstrings / the semantics of the methods.

Well, and the tests will obviously fail for now.

I also did some refactoring in interface.py so it will be quite easy to add support for #51 later.

Closes #157. Closes #104.


can/bus.py
```diff
@@ -2,29 +2,43 @@
...
+        * set the :attr:`~can.BusABC.channel_info` attribute to a string describing
+          the interface and/or channel
+
+    The *may* implement the following:
+        * :meth:`~can.BusABC.flush_tx_buffer` to allow discrading any
+          messages yet to be sent
+        * :meth:`~can.BusABC.shutdown` to override how the bus should
+          shut down, in which case the class has to either call through
+          `super().shutdown()` or call :meth:`~can.BusABC.flush_tx_buffer`
+          on its own
```
christiansandberg: Flushing buffer is usually unnecessary.

can/bus.py
```diff
@@ -37,29 +51,58 @@ def __init__(self, channel=None, can_filters=None, **config):
...
+                return msg
+
+            # if not, and timeout is None, try indefinitely
+            elif timeout is None:
+                continue
+
+            # try next one only if there still is time, and with reduced timeout
+            else:
+
+                timeout = timeout - (time() - start)
```
christiansandberg: Is this even correct? Maybe something like `time_left = timeout - (time() - start)` and use time_left in the `_recv_internal` call. Then you also need to set `time_left = timeout` in the beginning.

can/bus.py
```diff
@@ -92,7 +135,7 @@ def send_periodic(self, msg, period, duration=None):
             no duration is provided, the task will continue indefinitely.
 
         :return: A started task instance
-        :rtype: can.CyclicSendTaskABC
+        :rtype: :class:`can.CyclicSendTaskABC`
```
christiansandberg: Don't think you need to change this. See [Sphinx docs](www.sphinx-doc.org/en/master/domains.html#info-field-lists).

can/bus.py
```diff
@@ -119,20 +162,71 @@ def __iter__(self):
...
-            A filter matches, when ``<received_can_id> & can_mask == can_id & can_mask``
+    def _matches_filters(self, msg):
+        """Checks whether the given message matches at least one of the
+        current filters.
 
+        See :meth:`~can.BusABC.set_filters` for details.
         """
-        raise NotImplementedError("Trying to set_filters on unsupported bus")
+
+        # TODO: add unit testing for this method
```
christiansandberg: Also if filters is empty then this should always return True.

can/bus.py
```diff
@@ -119,20 +162,71 @@ def __iter__(self):
...
+        for can_filter in self._can_filters:
+            # check if this filter even applies to the message
+            if 'extended' in can_filter and \
+                can_filter['extended'] != msg.is_extended_id:
+                continue
+
+            # then check for the mask and id
+            can_id = can_filter['can_id']
+            can_mask = can_filter['can_mask']
+            if msg.arbitration_id & can_mask == can_id & can_mask:
```
christiansandberg: It could be faster to use `(can_id ^ msg.arbitration_id) & can_mask == 0` but is harder to read.

can/bus.py
```diff
@@ -142,7 +236,7 @@ def flush_tx_buffer(self):
     def shutdown(self):
         """
         Called to carry out any interface specific cleanup required
-        in shutting down a bus.
+        in shutting down a bus. Calls :meth:`~can.BusABC.flush_tx_buffer`.
```
christiansandberg: I don't know the reason for this. Maybe remove it. Is there any interface using it?

can/bus.py
```diff
@@ -37,29 +51,58 @@ def __init__(self, channel=None, can_filters=None, **config):
...
+                    continue
+                else:
+                    return None
+
+    @abstractmethod
+    def _recv_internal(self, timeout):
+        """
+        Read a message from the bus and tell whether it was filtered.
+
+        :raise: :class:`can.CanError`
```
christiansandberg: Use `:raises can.Error:` instead. See [Sphinx](www.sphinx-doc.org/en/master/domains.html#info-field-lists).

can/bus.py
```diff
@@ -37,29 +51,58 @@ def __init__(self, channel=None, can_filters=None, **config):
...
+            else:
+
+                timeout = timeout - (time() - start)
+
+                if timeout > 0:
+                    continue
+                else:
+                    return None
+
+    @abstractmethod
```
christiansandberg: Don't make this abstract or it will brake older external implementations. If an interface implements `recv()` directly then this will also work.

can/bus.py
felixdivo: Yes, you are right of course.

can/bus.py
felixdivo: But it is implemented like that right now in the BusABC class. Shell it be removed from there as well?

can/bus.py
felixdivo: Right!

can/bus.py
felixdivo: I will replace it and leave a comment explaining it.

can/bus.py
felixdivo: Well, `ixxat`, `kavaser`, `nican` and `vector` all implement `flush_tx_buffer()`. But none of the interfaces calls the method, at least none of the internal ones. I propose we remove the call to `flush_tx_buffer()` from `shutdown()` and remove the comment I added.

can/bus.py
felixdivo: You are right, I changed it.

can/bus.py
felixdivo: See 6a4aa5b.

felixdivo:
Shall I change all internal interfaces to work with this new ABC, or shall I change the ABC first?

felixdivo:
@hardbyte What do you think?

felixdivo:
**TODO:** We also need tests to check that everything works as intended.

felixdivo:
@christiansandberg @hardbyte I finished this PR as far as I know. The changing of all 13 backend interfaces as well as the core API has lead to many changes that ware required all over the place, along with a lot of cleanup effort that went into this. The goal of this PR is to unify the API more that it was previously the case, which was quite messy. But all unittests now run together with the new one I added.

can/bus.py
```diff
@@ -2,79 +2,174 @@
...
+    #: a string describing the underlying bus and/or channel
     channel_info = 'unknown'
 
     @abstractmethod
     def __init__(self, channel=None, can_filters=None, **config):
-        """
+        """Construct and open a CAN bus instance of the specified type.
+
+        Subclasses should call though this one with all parameters as
+        it applies filters (for now).
```
hardbyte: Keep in mind that a lot of the code comments have lasted many many years so when people read "for now" they won't easily have any idea if that is talking of change to come soon... or was an idea the developer had 5 years earlier...

can/bus.py
```diff
@@ -2,79 +2,174 @@
 # coding: utf-8
 
 """
-Contains the ABC bus implementation.
+Contains the ABC bus implementation and it's documentation.
```
hardbyte: nit: it's "its"

can/bus.py
```diff
@@ -2,79 +2,174 @@
...
 
     def __str__(self):
         return self.channel_info
 
-    @abstractmethod
     def recv(self, timeout=None):
         """Block waiting for a message from the Bus.
 
-        :param float timeout: Seconds to wait for a message.
+        If the concrete bus does not override it, this method makes sure
```
hardbyte: This is reading like library developer documentation - not user documentation.

This class along with the `Message` data class is essentially the primary interface to python-can. The front page of the docs links directly to [API Docs -> Bus](https://python-can.readthedocs.io/en/develop/bus.html#bus). I think this needs to be moved into an interface developers section of the docs.

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -30,6 +29,10 @@
 
 
 class ICSApiError(CanError):
+    """
+    TODO add docs
```
hardbyte: Yes please - or remove the docstring

doc/development.rst
```diff
@@ -27,6 +27,59 @@ The following assumes that the commands are executed from the root of the reposi
   makes Sphinx complain about more subtle problems.
 
 
+Creating a new interface/backend
```
hardbyte: Thanks for writing this! 🥇

PR #283, opened 2018-03-20
Feature: Context manager

edobez:
Hello everybody,
I am a new user of `python-can` but I decided to contribute a little to it.

I have started to work on a new feature: the context manager implementation for the Bus class, and I would like to do the same with other classes that require accessing and releasing resources, like Notifier.
In this way it's possible to open and close a bus instance automatically, i.e.:
```
with can.interface.Bus(bustype='pcan') as bus:
    bus.send(msg)

bus.send(msg)  # raises an exception since bus is closed
```
The context manager implementation is quite straight forward. What I wonder about is how to implement the exception behaviour.
In order to get an exception when requesting an operation after the bus has been closed, the instance should know about its 'openness' (note that this applies also without the context manager).
At the moment if you try to call a function after the `shutdown()` call, you'll get an exception specific to one interface implementation.
Relying on this, in my opinion, it's not the best since different interfaces (thus different implementations) can behave in different ways._
For example, the virtual interface doesn't release any resource after its shutdown and continues to work even after the closing call.

My idea is to use a state variable inside the instance that indicates if the bus is open or not, and use this information before trying to do any operation on the bus. 
This would require to edit all the current interface implementations by adding some helper functions, something like what I did on the `virtual` class and that you can see in the PR.

Let me know if this can be a good idea.


felixdivo:
Hm, very promising. :) Maybe #235 can help you with the states.

can/bus.py
```diff
@@ -146,4 +148,12 @@ def shutdown(self):
         """
         self.flush_tx_buffer()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        for t in self._tx_threads:
+            t.stop()
```
christiansandberg: Good idea. It could even be moved to the `.shutdown()` method in my opinion.

test/back2back_test.py
```diff
@@ -169,5 +169,103 @@ def test_basics(self):
         notifier.stop()
 
 
+class Back2BackTestCaseWithContext(unittest.TestCase):
```
christiansandberg: Don't think this is needed. The other test should suffice.

felixdivo:
@edobez Any updates on this?

felixdivo:
```
>       for thread in self._tx_threads:
E       AttributeError: 'VirtualBus' object has no attribute '_tx_threads'
```

Uhm, I guess this method should not assume that that attribute exists. It is not documented anywhere! And shouldn't it call `shutdown()` instead anyways?

felixdivo:
I removed the `_tx_threads` attribute since that might be for another PR/issue.

felixdivo:
I will merge it and then combine `Back2BackTestCaseWithContext` and `Back2BackTestCase` in another commit since I cannot push directly to this PR.

PR #290, opened 2018-04-02
Enhance serial interfaces

boris-wenzlaff:
- Rename serial can to simple serial
- Adjust simple serial and slcan interface
- Add generic interface test class
- Add documentation for generic interface test class
- Add test cases for slcan
- Enhance documentation for slcan
- Implementation of timeouts for simple serial issue (Closes #181)

Issue #273, opened 2018-02-24
Testing the SocketCAN backend on Travis CI


felixdivo:
Is it possible to test the SocketCAN backend on Travis CI using the virtual can channels provided by the kernel driver like `vcan0` or `vcan1`?

hardbyte:
I doubt it - just need to see if you can run `modprobe vcan` and the various `ip` commands required.

felixdivo:
Actually it works! See [here](https://travis-ci.org/hardbyte/python-can/jobs/347319036#L451).

felixdivo:
I submitted a proof-of-concept in #275.

felixdivo:
This should probably be done in a more general bus test cleanup. I am currently waiting for #290 to be merged. The idea ist to write a single test case with many tests, and then apply that test to all the interfaces, possibly adding dome more specific tests and opting in/out of special/unsupported bahaviour tests.

hardbyte:
I see the travis build of the `develop` branch is broken due to this?

https://travis-ci.org/hardbyte/python-can/jobs/403679419

christiansandberg:
Yes, I was surprised that we had `sudo: false` but used sudo anyway and it worked, but I guess it just worked temporarily.

felixdivo:
[It now works](https://travis-ci.org/hardbyte/python-can/builds/404328656). The build times didn't increase too much as well.

felixdivo:
This is done in `test/back2back_test.py`.

PR #296, opened 2018-04-10
SocketCAN low level fixes

felixdivo:
Changes to SocketCAN CTypes:

- [`libc.write(...)`]() might write less bytes than requested, and is now called in a loop until all data as been transmitted
- better error messages when sending: old `can.socketcan.ctypes failed to transmit` ->  new `can.socketcan_ctypes failed to transmit: ENETDOWN (errno 100): Network is down`

Changes to SocketCAN Native:

- [`socket.send()`]() might write less bytes than requested, and now [`socket.sendall()`]() is used
- uses similar error messages when sending like SocketCAN CTypes


felixdivo:
I tested the ctypes part rudimentary with some examples, but I would like to wait at least until #293 is merged so we can test SocketCAN on Travis CI. It would be good to also write more tests (see #273), maybe we could use `GenericInterfaceTest` from #290 for that.

test/config.py
```diff
@@ -34,6 +34,7 @@
 IS_WINDOWS = "windows" in _sys or ("win" in _sys and "darwin" not in _sys)
 IS_LINUX = "linux" in _sys
 IS_OSX = "darwin" in _sys
+IS_UNIX = IS_LINUX or IS_OSX
```
hardbyte: I'm sure some people would disagree with this definition of `UNIX` :-P

felixdivo: Yeah sure. But it works here and I could not find anything better in the first Google attempt ;-)

PR #304, opened 2018-05-02
Feature: Additional thread safe bus instance wrapper

felixdivo:
This PR adds a thread safe wrapper around existing interfaces. Calls to `send()` and `recv()` are locked independently. 

```Python
# this
from can import Bus
bus = Bus(bustype='some_interface', channel='0')

# becomes this
from can import ThreadSafeBus
bus = ThreadSafeBus(bustype='some_interface', channel='0')
```

In that case, the locking of `send_periodic()` is disabled since it is redundant. 

Testing and docs are still left to be done.

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -2,7 +2,7 @@
 # coding: utf-8
 
 """
-can is an object-orient Controller Area Network interface module.
+can is an object-oriented Controller Area Network interface module.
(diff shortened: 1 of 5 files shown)


felixdivo:
I will wait a while with this until #277 gets finalized & merged.

PR #309, opened 2018-05-16
cleanups,fixes & docs in Bus.__new__() and util.load_config()

### LINKED ISSUE #280: Specify bitrate in a config file?
Is there a way to specify the `bitrate` parameter in a configuration file?  From what I can see, there isn't.

I'd like to be able to supply a config file with parameters that work on a CI machine but allow a developer to override those values using environment variables when working on their local workstation.  There is already an environment variable for this, but the current code doesn't seem to use a `bitrate` that was provided in a configuration file.

I saw #172, but that PR doesn't seem to show any actual changes and it was closed without merging after being approved.

Is there a reason why this wouldn't be a desirable feature?

Thanks for a super-handy library!  Works great in all of my testing so far!

felixdivo:
This PR changes `util.load_config()` to pass through any unused parameters. It is made more clear in the code what `Bus.__new__()` is doing and what not. A lot has moved to `util.load_config()` to remove code and responsibility duplication. The documentation has each been changed to reflect what the methods are actually doing, that was not done correct before. 

Based an @randycoulman's comment [over here](https://github.com/hardbyte/python-can/issues/280#issuecomment-373426370)
Fixes #280

### DIFF
--- a/can/interface.py
+++ b/can/interface.py
@@ -129,8 +129,6 @@ class Bus(BusABC):
             raise ValueError("channel argument missing")
 
         # the channel attribute should be present in **config
-        print("DEBUGGING: ", args)
-        print("DEBUGGING: ", config)
         return cls(*args, **config)
 
 
--- a/can/interface.py
+++ b/can/interface.py
@@ -7,7 +7,7 @@ as a list of all avalibale backends and some implemented
 CyclicSendTasks.
 """
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function
 
 import sys
 import importlib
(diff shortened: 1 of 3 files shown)


felixdivo:
This is **done**, but we should write some unit tests before merging this PR to later prevent regression. 

Note: Making `__new__` a `@classmethod` was a really nasty bug and took me quite a while to figure out. -.- Don't do this in the future! 😛

can/interface.py
```diff
@@ -99,36 +99,39 @@ class Bus(BusABC):
...
+        # remove the 'interface' key so it doesn't get passed to the backend
+        del config['interface']
+
+        # make sure the bus can handle this config
+        if 'channel' not in config:
+            raise ValueError("channel argument missing")
+
+        # the channel attribute should be present in **config
+        print("DEBUGGING: ", args)
+        print("DEBUGGING: ", config)
```
hardbyte: Remove these (or convert to debug level logs)

felixdivo: Ah sure, must've forgotten them there.

PR #315, opened 2018-05-28
Feature: Synchronized bus instance

felixdivo:
Adds a `can.ThreadSafeBus` that can be used as a thread safe drop-in replacement for `can.Bus`. The send and receive methods are synchronized separately.


felixdivo:
TODO: Documentation

doc/bus.rst
```diff
@@ -37,10 +39,24 @@ Receiving
 '''''''''
 
 Reading from the bus is achieved by either calling the :meth:`~can.BusABC.recv` method or
-by directly iterating over the bus::
+by directly iterating over the bus:
```
christiansandberg: This is not a mistake, see [here](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#indented-literal-blocks).

doc/bus.rst
```diff
@@ -37,10 +39,24 @@ Receiving
...
+
+Thread safe bus
+---------------
+
+This thread safe version of the bus class can be used by multiple threads at once.
+Sending and receiving is locked seperatly to avoid unnessesary delays.
+
+It can be used exactly like the normal :class:`~can.Bus` class:
+
+    my_bus = can.Bus(interface='socketcan', channel='vcan0')
```
christiansandberg: Shouldn't this be `can.ThreadSafeBus`?

setup.py
```diff
@@ -56,6 +57,7 @@
     python_requires=">=2.7,!=3.0,!=3.1,!=3.2,!=3.3",
     install_requires=[
         'setuptools',
+        'wrapt ~= 1.10',
     ] + (['subprocess32 ~= 3.2.7'] if version_info.major < 3 else []),
```
christiansandberg: Unrelated to this pull request, but why is setuptools here? The whole setup.py file requires setuptools to be installed already. And is subprocess32 really needed? Python 2.7 has [subprocess.check_output](https://docs.python.org/2.7/library/subprocess.html#subprocess.check_output) or does it not work as expected? In any case it should be moved to `extras_require` as it is only needed for socketcan.

felixdivo: The `setuptools` package is added for completeness, it is not exactly wrong that it is an dependency. It doesn't really help with anything though. Shall we remove it or leave it? I think it doesn't really matter.

Hm, I agree that `subprocess32` should be moved into `extras_require`. But I added it as an dependency because of this note at the top of the documentation you linked: 

> POSIX users (Linux, BSD, etc.) are strongly encouraged to install and use the much more recent subprocess32 module instead of the version included with python 2.7. It is a drop in replacement with better behavior in many situations.

I simply followed that advice.

christiansandberg: setuptools is not a dependency for the can module itself. It is setuptools that handles the the `install_requires` at setup time so it is a bit strange and I haven't seen any other packages do the same.

Have you tested if it works on Python 2.7 using the built-in subprocess module? If it does then it seems unnecessary to add an extra dependency. After all, it is a pretty simple function that is not critical.

felixdivo: Okay, I'll remove the `setuptools` dependency. 

Lets's see what Travis CI will tell us about the built-in `subprocess` module ...

felixdivo:
Okay, so it was possible to remove the `subprocess32` module without any problems. This does not belong to the topic of this PR but is pushed here because the discussion has taken place in here.

felixdivo:
Do you know why `test/socketcan_helpers.py` is never executed?

christiansandberg:
It must contain “test” in the filename.

felixdivo:
@christiansandberg

> It must contain “test” in the filename.

Now it works. Thanks!

@hardbyte 
Is the documentation now sufficient?

felixdivo:
Shall I merge it?

hardbyte:
:shipit:

PR #316, opened 2018-05-30
Add support for new BLF files

### LINKED ISSUE #314: Assertion error when trying to parse a legit CANoe generated BLF file
I'm trying to parse a BLF using [this approach](https://stackoverflow.com/questions/49499903/how-do-i-convert-blf-data-from-can-to-csv-using-python):

```
import can
from pprint import pprint as pp

filename = "logfile.blf"
logging = can.BLFReader(filename)
logging = list(logging)
pp(logging)
```

and get 
```
  File "D:\eclipse-workspace\blf2csv\pythoncan_blf2csv.py", line 7, in <module>
    logging = list(logging)
  File "C:\Users\my_user_name\AppData\Local\Python\Python36\lib\site-packages\can\io\blf.py", line 129, in __iter__
    assert header[0] == b"LOBJ", "Parse error"
AssertionError: Parse error
```

The file is legit (I can convert it to CSV using Vector CANoe `Tool >> Log File Conversion` dialog (so I can convert a single file but looked at `python-can` as a means to do a batch conversion of multiple log files)).

christiansandberg:
Supports version 2 of object headers.
Supports any file header size.
Better exceptions on parse errors.

Fixes #314.

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -19,6 +19,7 @@ import struct
 import zlib
 import datetime
 import time
+import logging
 
 from can.message import Message
 from can.listener import Listener
@@ -29,6 +30,7 @@ class BLFParseError(Exception):
     """BLF file could not be parsed correctly."""
     pass
 
+LOG = logging.getLogger(__name__)
 
 # 0 = unknown, 2 = CANoe
 APPLICATION_ID = 5
@@ -160,6 +162,7 @@ class BLFReader(object):
                     data = zlib.decompress(container_data, 15, uncompressed_size)
                 else:
                     # Unknown compression method
+                    LOG.warning("Unknown compression method (%d)", method)
                     continue
 
                 if tail:
@@ -189,6 +192,7 @@ class BLFReader(object):
                         pos += OBJ_HEADER_V2_STRUCT.size
                     else:
                         # Unknown header version
+                        LOG.warning("Unknown object header version (%d)", header_version)
                         pos = next_pos
                         continue
 
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -25,24 +25,32 @@ from can.listener import Listener
 from can.util import len2dlc, dlc2len
 
 
+class BLFParseError(Exception):
+    """BLF file could not be parsed correctly."""
+    pass
+
+
 # 0 = unknown, 2 = CANoe
 APPLICATION_ID = 5
 
-# Header must be 144 bytes in total
 # signature ("LOGG"), header size,
 # application ID, application major, application minor, application build,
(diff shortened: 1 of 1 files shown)


christiansandberg:
@tburly, could you test this with your BLF file? I only have older files.

```
pip install https://github.com/hardbyte/python-can/archive/blf-fix.zip
```

can/io/blf.py
```diff
@@ -25,24 +25,32 @@
 from can.util import len2dlc, dlc2len
 
 
+class BLFParseError(Exception):
```
felixdivo: Should this extend `IOError`?

can/io/blf.py
```diff
@@ -153,19 +168,37 @@ def __iter__(self):
...
-                    pos += OBJ_HEADER_STRUCT.size
+                    header_version = header[2]
+                    if header_version == 1:
+                        flags, _, _, timestamp = OBJ_HEADER_V1_STRUCT.unpack_from(data, pos)
+                        pos += OBJ_HEADER_V1_STRUCT.size
+                    elif header_version == 2:
+                        flags, _, _, timestamp, _ = OBJ_HEADER_V2_STRUCT.unpack_from(data, pos)
+                        pos += OBJ_HEADER_V2_STRUCT.size
+                    else:
+                        # Unknown header version
```
felixdivo: Maybe this should be stated with `log.debug` or `log.warn`

can/io/blf.py
christiansandberg: I googled it and the [accepted answer](https://stackoverflow.com/questions/9276663/what-exception-class-to-use-for-file-parsing-error) seems to be to inherit from Exception. IOError is more low level I think.

felixdivo: Okay, I'm convinced. ;)

tburly:
> @tburly, could you test this with your BLF file? I only have older files.
Now there's no error but there's nothing to loop over in can.io.blf.BLFReader object created from the given BLF file.

As to providing some dummy BLF file for testing, I'll look over this. Have to ask higher than myself if that's possible.

christiansandberg:
@tburly, can you update and try again? I added support for a second CAN message type that might be used in newer log files.

Corpelijn:
@christiansandberg I can confirm the second CAN message type is being read with the new implementation

christiansandberg:
Cool, thanks a lot!

tburly:
@christiansandberg also confirming that now a can.io.blf.BLFReader object created from my BLF yields can.Message objects when iterated over. Thx for the quick action!

dubourg:
Confirmed here as well. Thanks!

PR #323, opened 2018-06-09
SocketCAN (native) cyclic improvements

### LINKED ISSUE #298: .modify_data() in .send_periodic() "restarts" cycle time
Hello,

i noticed that the modify_data() call always sends a new frame independent from the actual time the last frame was sent. 

According to the [socket-can doc](https://github.com/nebaruzdin/can-doc/blob/master/can-bcm.txt) this can be controlled with the STARTTIMER flag (1.3.2).

Locally i added another ._tx_setup function that creates a different bcm header. It just doesnt send the STARTTIMER for tasks that are already sending periodically. 

What would be the best way to implement that?

christiansandberg:
Keep periodic transmission timing on data update.
Use BCM for setting duration.

Fixes #298.

### DIFF
--- a/can/interfaces/socketcan/socketcan_native.py
+++ b/can/interfaces/socketcan/socketcan_native.py
@@ -274,6 +274,7 @@ class CyclicSendTask(SocketCanBCMBase, LimitedDurationCyclicSendTaskABC,
         Note the Message must have the same :attr:`~can.Message.arbitration_id`.
         """
         assert message.arbitration_id == self.can_id, "You cannot modify the can identifier"
+        self.message = message
         header = build_bcm_update_header(self.can_id_with_flags, self.flags)
         frame = build_can_frame(message)
         send_bcm(self.bcm_socket, header + frame)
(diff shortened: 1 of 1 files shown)


can/interfaces/socketcan/socketcan_native.py
```diff
@@ -224,22 +227,31 @@ class CyclicSendTask(SocketCanBCMBase, LimitedDurationCyclicSendTaskABC,
 
     """
 
-    def __init__(self, channel, message, period):
+    def __init__(self, channel, message, period, duration=None):
         """
         :param channel: The name of the CAN channel to connect to.
         :param message: The message to be sent periodically.
         :param period: The rate in seconds at which to send the message.
         """
```
felixdivo: maybe add the param to the docs here

can/interfaces/socketcan/socketcan_native.py
```diff
@@ -262,7 +274,10 @@ def modify_data(self, message):
         Note the Message must have the same :attr:`~can.Message.arbitration_id`.
```
felixdivo: ... "like the one of the message given in the constructor"

PR #324, opened 2018-06-11
Adding proper bitrate and timestamp support for neovi

pierreluctg:


pierreluctg:
FYI @hardbyte

felixdivo:
Can this be tested in some way with unit tests?

pierreluctg:
@felixdivo not really unit testable without python-ics, icsneo40 dll and NeoVi device (this is no longer unit test) or python-ics and a icsneo40 mock dll

felixdivo:
Okay. Maybe we should add a mock dll at some point.

PR #325, opened 2018-06-12
Setting _filters to None when filter Iterator is empty

pierreluctg:
For example when using `can.logger` the default `can_filters` value is a empty list, this leads to filtering all message. This change fix the issue by treating any empty Iterator same as `filters=None`

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -201,7 +201,7 @@ class BusABC(object):
         """Apply filtering to all messages received by this Bus.
 
         All messages that match at least one filter are returned.
-        If `filters` is `None` or a a zero size interable, all 
+        If `filters` is `None` or a zero length sequence, all 
         messages are matched.
 
         Calling without passing any filters will reset the applied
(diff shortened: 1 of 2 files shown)


pierreluctg:
@hardbyte What is your opinion on that one?

Currently this is not passing the match_nothing test in https://github.com/hardbyte/python-can/blob/develop/test/test_message_filtering.py#L49

I am not sure is this test make sense or not.

felixdivo:
When I implemented that API, I did not find any precise or central documentation on how to handle `can_filters`. I simply have treated empty iterables like I thought it to be the most intuitive. That's the story behind the change I made.

felixdivo:
When this is actually changed, please update the documentation accordingly.

can/bus.py
```diff
@@ -201,8 +201,8 @@ def set_filters(self, filters=None):
         """Apply filtering to all messages received by this Bus.
 
         All messages that match at least one filter are returned.
-        If `filters` is `None`, all messages are matched.
-        If it is a zero size interable, no messages are matched.
+        If `filters` is `None` or a a zero size interable, all
```
hardbyte: typo here. I'd say "a zero length sequence"

PR #326, opened 2018-06-12
Merge socketcan_native and socketcan_ctypes into one

christiansandberg:
Since it takes a lot of effort to maintain both implementations I thought it might be a good idea to merge them into one. I took the native implementation, removed reliance on `socket` constants and provided a libc fallback for `socket.bind` and `socket.connect`.

Will update docs.


felixdivo:
Hm, interesting! This would actually remove some more edge cases/ugly clutter all over the library. I will have a look at it the next days.

test/back2back_test.py
```diff
@@ -136,38 +135,12 @@ def test_fd_message_with_brs(self):
...
-        self.bus1.shutdown()
-        self.bus2.shutdown()
-
-    def test_basics(self):
-        reader = can.BufferedReader()
-        notifier = can.Notifier(self.bus2, [reader])
-
-        message = can.Message(arbitration_id=0x4321, data=[1, 2, 3], extended_id=True)
-        self.bus1.send(message)
+class BasicTestSocketCan(Back2BackTestCase):
```
felixdivo: Nice one!

can/interface.py
```diff
@@ -20,6 +20,8 @@
 from .util import load_config
 from .interfaces import BACKENDS
 
+from can.interfaces.socketcan.socketcan import CyclicSendTask, MultiRateCyclicSendTask
```
felixdivo: The module name `socketcan.socketcan` is a bit redundant.

can/interface.py
```diff
@@ -175,43 +169,3 @@ def detect_available_configs(interfaces=None):
             result += available
 
     return result
-
```
felixdivo: Good to see this being cleaned up

can/interfaces/__init__.py
```diff
@@ -29,4 +28,4 @@
     for interface in iter_entry_points('can.interface')
 })
 
-VALID_INTERFACES = frozenset(list(BACKENDS.keys()) + ['socketcan'])
+VALID_INTERFACES = frozenset(list(BACKENDS.keys()) + ['socketcan_native', 'socketcan_ctypes'])
```
felixdivo: Well, we could then just remove the special case all together. Although this would break compatibility with older versions, it would clean up a lot of code.

can/interfaces/socketcan/__init__.py
```diff
@@ -5,6 +5,5 @@
 See: https://www.kernel.org/doc/Documentation/networking/can.txt
 """
 
-from can.interfaces.socketcan import socketcan_constants as constants
-from can.interfaces.socketcan.socketcan_ctypes import SocketcanCtypes_Bus
-from can.interfaces.socketcan.socketcan_native import SocketcanNative_Bus
+from can.interfaces.socketcan import constants
```
felixdivo: Should these really be "public"?

can/interfaces/socketcan/socketcan.py
```diff
@@ -1,22 +1,16 @@
 #!/usr/bin/env python
 # coding: utf-8
-
-"""
```
felixdivo: Maybe keep a note that this was merged?

can/interfaces/socketcan/socketcan.py
```diff
@@ -157,11 +153,17 @@ def dissect_can_frame(frame):
...
     except AttributeError:
         raise SystemExit("To use BCM sockets you need Python3.4 or higher")
     try:
         s.connect((channel,))
+    except socket.error:
+        # For Python 2.7
+        addr = get_addr(s, channel)
+        error = libc.connect(s.fileno(), addr, len(addr))
+        if error < 0:
+            raise can.CanError('Could not connect to socket')
```
felixdivo: Maybe also print some better error message.

can/util.py
```diff
@@ -182,9 +182,9 @@ def load_config(path=None, config=None):
         if key not in config:
             config[key] = None
 
-    # this is done later too but better safe than sorry
-    if config['interface'] == 'socketcan':
-        config['interface'] = choose_socketcan_implementation()
+    # deprecated socketcan types
```
felixdivo: Maybe print a warning to make users switch to simply `socketcan´.

can/interfaces/socketcan/socketcan.py
```diff
@@ -435,34 +433,34 @@ def __init__(self, channel, receive_own_messages=False, fd=False, **kwargs):
...
         kwargs.update({'receive_own_messages': receive_own_messages, 'fd': fd})
-        super(SocketcanNative_Bus, self).__init__(channel=channel, **kwargs)
+        super(SocketcanBus, self).__init__(channel=channel, **kwargs)
 
     def shutdown(self):
         self.socket.close()
 
     def _recv_internal(self, timeout):
-        if timeout:
+        if timeout is not None:
```
felixdivo: A timeout of `0` should not try to invoke `select()`, or should it?

felixdivo:
Sorry for the many comments, but this *really nice* PR touches quite a bit of code / code structure.

Once everything code-related is cleaned up, we should also update the docs where appropriate.

can/interface.py
christiansandberg: Agree. Do you have a suggestion for a new name?

can/interfaces/socketcan/__init__.py
felixdivo: Hm. But this really is just an implementation detail that nobody should care about. I doubt we would touch a lot of code with that change.

can/util.py
christiansandberg: Good idea. Yes we should eventually remove this but first a logged warning is good, then maybe a [`DeprecationWarning`](https://docs.python.org/3/library/exceptions.html#DeprecationWarning).

can/interface.py
felixdivo: Uhm; `impl`? Not really better though ...

can/interfaces/socketcan/__init__.py
christiansandberg: Sure, I'll remove it. Since the module is now named "constants" it will work anyway.

can/interfaces/socketcan/socketcan.py
christiansandberg: It should, because otherwise a timeout of 0 will go directly to `recv` which will block with infinite timeout.

can/interfaces/socketcan/socketcan.py
```diff
@@ -157,11 +153,17 @@ def dissect_can_frame(frame):
...
     try:
         s.connect((channel,))
+    except socket.error:
+        # For Python 2.7
+        addr = get_addr(s, channel)
+        error = libc.connect(s.fileno(), addr, len(addr))
+        if error < 0:
+            raise can.CanError('Could not connect to socket')
     except OSError as e:
         log.error("Couldn't connect a broadcast manager socket")
```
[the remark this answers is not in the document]
christiansandberg: Won't that information be in the exception that is raised after?

can/interfaces/socketcan/socketcan.py
```diff
@@ -435,34 +433,34 @@ def __init__(self, channel, receive_own_messages=False, fd=False, **kwargs):
...
 
         # set the receive_own_messages paramater
         try:
-            self.socket.setsockopt(socket.SOL_CAN_RAW,
-                                   socket.CAN_RAW_RECV_OWN_MSGS,
+            self.socket.setsockopt(SOL_CAN_RAW,
+                                   CAN_RAW_RECV_OWN_MSGS,
                                    struct.pack('i', receive_own_messages))
         except socket.error as e:
             log.error("Could not receive own messages (%s)", e)
```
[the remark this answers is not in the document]
christiansandberg: Isn't that also in the exception message?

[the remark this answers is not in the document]
felixdivo: Ah right. It create something like: `socket.error: [Errno 107] Transport endpoint is not connected`. This seems to be only required with the libc fallback.

can/interfaces/socketcan/socketcan.py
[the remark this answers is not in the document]
felixdivo: Yeah. See below.

felixdivo:
`socketcan.rst` should still mention that filters and periodic sending are quite efficient because of in-kernel handling of that tasks. You could just copy an explanation together from the deleted parts of 6ce8565.

can/interfaces/socketcan/socketcan.py
```diff
@@ -404,11 +421,12 @@ class SocketcanBus(BusABC):
     Implements :meth:`can.BusABC._detect_available_configs`.
     """
 
-    def __init__(self, channel, receive_own_messages=False, fd=False, **kwargs):
+    def __init__(self, channel="", receive_own_messages=False, fd=False, **kwargs):
         """
         :param str channel:
             The can interface name with which to create this bus. An example channel
             would be 'vcan0' or 'can0'.
+            An empty string '' will receive messages from all channels.
```
felixdivo: And what about sending?

can/interfaces/socketcan/socketcan.py
```diff
@@ -473,12 +505,39 @@ def send(self, msg, timeout=None):
             if not ready_send_sockets:
                 raise can.CanError("Timeout while sending")
 
+        data = build_can_frame(msg)
         try:
-            self.socket.sendall(build_can_frame(msg))
+            if self.channel == "" and HAS_NATIVE_SUPPORT and msg.channel:
+                self.socket.sendto(data, (msg.channel, ))
+            else:
+                self.socket.send(data)
```
felixdivo: Shouldn't we use `sendall()`, since `send()` might not transmit everything?

can/interfaces/socketcan/socketcan.py
```diff
@@ -473,12 +505,39 @@ def send(self, msg, timeout=None):
             if not ready_send_sockets:
                 raise can.CanError("Timeout while sending")
 
+        data = build_can_frame(msg)
         try:
-            self.socket.sendall(build_can_frame(msg))
+            if self.channel == "" and HAS_NATIVE_SUPPORT and msg.channel:
+                self.socket.sendto(data, (msg.channel, ))
```
felixdivo: > Return the number of bytes sent.

I guess we should make sure the rest gets sent as well (in a loop or handled completely in a auxiliary method) if the returned number of bytes is less than `len(data)`.

can/interfaces/socketcan/socketcan.py
christiansandberg: The default behaviour for all other interfaces and previously for socketcan is to not block on `send()`. If it can't be transmitted, it should be raised as an exception and the application can decide if it wants to try again, move on to the next message or signal as an error. The most common scenario is that something is wrong with the bus and no nodes are ACK:ing your messages. If we would be using `sendall` then the application will just freeze.

I added the `timeout` argument for making `send` blocking, but usually you want to give up after a certain time. One problem with my change will arise if it is possible to queue only part of a message. Then we want to go back to `select` and loop until all bytes are sent, but only for the timeout period. I'll give that a go.

can/interfaces/socketcan/socketcan.py
```diff
@@ -418,92 +437,157 @@ def capture_message(sock):
...
+        time_left = timeout
+        data = build_can_frame(msg)
 
+        while time_left >= 0:
+            # Wait for write availability
+            ready = select.select([], [self.socket], [], time_left)[1]
+            if not ready:
+                # Timeout
+                break
+            sent = self._send_once(data, msg.channel)
```
felixdivo: This is a nice solution! But maybe it would be more efficient to use `socket.sendall()` when there is native support?

christiansandberg: The problem is that `sendall` doesn't have a timeout argument. And it is probably very unusual, if even impossible, that a message could be only partially cued. For most cases it will succeed on first try, or it will fail completely on first `select`.

christiansandberg:
Do you mean if broadcast channel is used? Then you currently need to address each message to a specific channel, otherwise it will fail. I don't think there is a way to send to all channels as far as I can see. I can add a note about that.

can/interface.py
hardbyte: Why not just hoist them into `can.interfaces.socketcan`? 
In `can.interfaces.socketcan.__init__` we can import the classes and functions that make up the implementation

can/util.py
hardbyte: Could you please add a comment like this one https://github.com/hardbyte/python-can/commit/07be6b5aa0fd6bcf84e24d37b76c8866120920f8

can/interface.py
christiansandberg: Do you mean that we should make the current socketcan.py __init__.py?

hardbyte: Nah (although that would work), I was just more commenting that no matter what file the classes end up in we should expose them via `can.interfaces.socketcan` e.g., `can.interfaces.socketcan.CyclicSendTask`.

christiansandberg: That’s already done so this is more for internal architecture.

PR #327, opened 2018-06-13
Adding detect_available_configs support to neovi

pierreluctg:


felixdivo:
The CI tests fail because of this:

`neovi_bus.py                26 WARNING  You won't be able to use the ICS NeoVi can backend without the python-ics module installed!: No module named 'ics'`

Maybe not all libraries are installed by [this](https://github.com/hardbyte/python-can/blob/develop/setup.py#L23). Maybe we should simply install all optional packages in tests.

pierreluctg:
@felixdivo adding `python-ics` to `tests_require` is not a problem. However, using it requires have the neovi driver installed (or in the path) this will be problematic.

For now we can simply return a empty dict from `detect_available_configs` if `python-ics` is unavailable. What do you think?

felixdivo:
Hm I see. Sounds good.

PR #336, opened 2018-06-21
Stop raising accuracy error for NeoVi

pierreluctg:
Only keep logging the error that ICS lib consider critical and stop raising exception. In most case, the error is affecting the accuracy but the device can still be use normally (i.e. Rx buffer overflow). In these cases, we do not want for example to kill a notifier using this bus, etc.

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -188,8 +188,6 @@ class NeoViBus(BusABC):
 
             for msg in ics.get_error_messages(self.dev):
                 error = ICSApiError(*msg)
-                if error.is_critical:
-                    raise error
                 logger.warning(error)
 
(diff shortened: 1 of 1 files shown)


pierreluctg:
Python 3.4 Travis job failed because of a connection timeout when downloading Python 3.4 archive.

All the other Python jobs are passing.

> 3.4 is not installed; attempting download
Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.4.tar.bz2
$ curl -sSf -o python-3.4.tar.bz2 ${archive_url}
curl: (7) Failed to connect to s3.amazonaws.com port 443: Connection timed out
Unable to download 3.4 archive. The archive may not exist. Please consider a different version.

felixdivo:
Done. It now works. :)

PR #341, opened 2018-06-27
Log received messages for all interfaces

christiansandberg:
Move received message logging to `can.BusABC.recv()` and comment out unneeded logging lines in interface implementations (unless debugging the implementation). Use level 9 (lower than debug) to be able to opt-in since the log can easily be overwhelmed by random CAN data.


felixdivo:
Nice solution :+1:

Issue #340, opened 2018-06-27
Sphinx CI


felixdivo:
Is it possible and desirable to check whether Sphinx complains about new stuff in a pull request? Because not everybody (including me) always gets the syntax in the doc strings right, and a CI "test" could check this automatically, to improve the overall doc quality.

Issue #262, opened 2018-02-22
Fixing Sphinx - Documentation improvements


felixdivo:
Running `sphinx-build doc/ doc/_build -n` reveals many apparent problems. Do we need to solve them?

hardbyte:
Yes as it probably effects how the docs are rendered. I'll take a look.

felixdivo:
Maybe we should test this on the CI server?

felixdivo:
Update on this: The docs are now built as part of the Python 3.6 test on Travis CI with the option `-n` to reveal any problems. They are not considered errors for now. Any help is appreciated.

hardbyte:
Here are the errors in master from the 3.1 merge:

```
/home/travis/build/hardbyte/python-can/can/interface.py:docstring of can.Bus:1: WARNING: py:class reference target not found: can.bus.BusABC
/home/travis/build/hardbyte/python-can/can/bus.py:docstring of can.BusABC.send_periodic:9: WARNING: py:meth reference target not found: CyclicTask.stop
/home/travis/build/hardbyte/python-can/can/bus.py:docstring of can.BusABC.send_periodic:22: WARNING: py:meth reference target not found: stop
/home/travis/build/hardbyte/python-can/can/bus.py:docstring of can.BusABC.state:: WARNING: py:class reference target not found: NamedTuple
/home/travis/build/hardbyte/python-can/can/thread_safe_bus.py:docstring of can.ThreadSafeBus:1: WARNING: py:class reference target not found: ObjectProxy
/home/travis/build/hardbyte/python-can/can/thread_safe_bus.py:docstring of can.ThreadSafeBus:10: WARNING: py:meth reference target not found: can.BusABC._recv_internal
/home/travis/build/hardbyte/python-can/can/thread_safe_bus.py:docstring of can.ThreadSafeBus:10: WARNING: py:meth reference target not found: can.BusABC._recv_internal
/home/travis/build/hardbyte/python-can/can/interfaces/canalystii.py:docstring of can.interfaces.canalystii.CANalystIIBus:1: WARNING: py:class reference target not found: can.bus.BusABC
/home/travis/build/hardbyte/python-can/can/interfaces/iscan.py:docstring of can.interfaces.iscan.IscanBus:1: WARNING: py:class reference target not found: can.bus.BusABC
[...]

PR #342, opened 2018-06-28
Feature travis to build sphinx docs

hardbyte:


felixdivo:
We might also run Sphinx with [`-n`](http://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-n) (= in nit-picky mode). We could then gradually remove the errors in that log. See also #262.

felixdivo:
Okay nice

felixdivo:
Closes #340.

PR #346, opened 2018-06-29
Add coverage report generation

felixdivo:
This adds very simple statement coverage reports like [this one](https://travis-ci.org/hardbyte/python-can/jobs/398444150#L630) to the Travis CI & AppVeyor test routines. 

It would be nice to also add some tool like [Codecov.io](https://codecov.io/). That might help us identify poorly tested regions of code. 

Adding coverage tool support via the GitHub Marketplace should be quite straightforward, but I do not have the required admin permissions. @hardbyte Can you do that?

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -136,7 +136,7 @@ class NeoViBus(BusABC):
     def shutdown(self):
         super(NeoViBus, self).shutdown()
         ics.close_device(self.dev)
-    
+
     @staticmethod
     def _detect_available_configs():
         """Detect all configurations/channels that this interface could
@@ -148,14 +148,18 @@ class NeoViBus(BusABC):
         """
         if ics is None:
             return []
-        # TODO: add the channel(s)
+
         try:
-            return [{
-                'serial': NeoViBus.get_serial_number(device)
-            } for device in ics.find_devices()]
+            devices = ics.find_devices()
         except Exception as e:
             logger.debug("Failed to detect configs: %s", e)
             return []
+
+        # TODO: add the channel(s)
+        return [{
+            'interface': 'neovi',
+            'serial': NeoViBus.get_serial_number(device)
+        } for device in devices]
 
     def _find_device(self, type_filter=None, serial=None):
         if type_filter is not None:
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -20,13 +21,20 @@
...
     'mock ~= 2.0',
-    'nose ~= 1.3.7',
+    'nose ~= 1.3',
     'pytest ~= 3.6',
     'pytest-timeout ~= 1.2',
-    'pyserial ~= 3.0'
-]
+    'pytest-cov ~= 2.5',
+] + extras_require['serial']
+#for key, requirements in extras_require.items():
```
hardbyte: Please remove the commented out lines

setup.py
```diff
@@ -8,6 +8,7 @@
 from sys import version_info
 import re
 import logging
+from itertools import chain
```
hardbyte: Not used?

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -149,9 +149,13 @@ def _detect_available_configs():
...
             return []
         # TODO: add the channel(s)
-        return [{
-            'serial': NeoViBus.get_serial_number(device)
-        } for device in ics.find_devices()]
+        try:
+            return [{
+                'serial': NeoViBus.get_serial_number(device)
+            } for device in ics.find_devices()]
+        except Exception as e:
```
hardbyte: I wonder if this `try`/`for` should be inverted - if the `try` was around each call to `get_serial_number` then any working devices can be returned?

hardbyte:
I've added the codecov integration and travis should now report coverage... Let's see

hardbyte:
Looks like codecov is getting the data - https://codecov.io/gh/hardbyte/python-can/tree/36361d149e65839bf987c20cbbed3f4366bdf851/can

felixdivo:
@hardbyte Well, after changing `setup.py` Neovi kept crashing, and that simple fix was just there to make the test suite succeed.

can/interfaces/ics_neovi/neovi_bus.py
felixdivo: I made the code a bit clearer to show that only the call to `ics.find_devices()` is wrapped in a try-catch clause. The method `NeoViBus.get_serial_number()` itself should never fail.

PR #347, opened 2018-07-01
Fix Sphinx errors/warnings & Misc. cleanups

felixdivo:
@hardbyte I'll fix some errors here based on the work you have already done a few months back, and notify you when I'm done.


felixdivo:
I honestly can't make much sense of the remaining Sphinx output, like [this one](https://travis-ci.org/hardbyte/python-can/jobs/398884222#L766).

can/util.py
```diff
@@ -190,10 +190,39 @@ def load_config(path=None, config=None):
     if 'bitrate' in config:
         config['bitrate'] = int(config['bitrate'])
 
-    can.log.debug("loaded can config: {}".format(config))
+    can.log.debug("can config: {}".format(config))
     return config
 
 
+def choose_socketcan_implementation():
```
hardbyte: This shouldn't be in a doc update PR

felixdivo: This branch that you created quite a while back was a few dozen commits behind `develop`, and this resulted from the rebase to resolve the conflicts. Was this wrong?

hardbyte: Oh right I didn't notice you were starting with an older branch. Yes this was removed on purpose (in the socketCAN consolidation).

I suggest rebasing on top of develop - either force push over this branch or to another branch.

hardbyte: If it helps I've rebased my original clean up commits on the current develop branch `fix-sphinx-errors-rebase`

felixdivo: Uhm, I already applied other patches on top of the rebase. Wouldn't it be the simplest to just remove the `choose_socketcan_implementation()` function?

felixdivo:
Should we ask for help over here: https://github.com/sphinx-doc/sphinx ?

can/util.py
hardbyte: Sure

hardbyte:
No I don't think there is a problem with sphinx. I think we just haven't been particularly careful with our docs.

One additional thing though - we should check the heading levels make sense. In the 2.2.0 release we accidentally included this https://python-can.readthedocs.io/en/2.2.0/development.html#about-the-busabc-class

felixdivo:
Okay; I was rather supposing that we ask for how to get more helpful error statements out of Sphinx' build process.

Well, was that section really an accident? It is in the Developer's section, which is the right place for that, isn't it?

hardbyte:
Sorry I should have been clearer - the section's content is fine, it is the heading level that is making it much more important than it should be:

TOC

felixdivo:
Oh sure, that's not meant to be!

can/util.py
felixdivo: Done.

felixdivo:
Does not fix all problems, so #262 remains open.

PR #348, opened 2018-07-03
IO module enhancements

### LINKED ISSUE #268: File handles in can.io.* are not always released
Actually it is "just" `CanutilsLogReader`, `BLFReader`\*, `CSVReader`\* and `ASCReader`. 

Maybe we should always close the file/ the database connection if all messages have been read in `__iter__()` and additionally provide a `stop()`/`close()` method? Should all of these inherit from the same base class?

[*] if not all messages have been read

felixdivo:
This PR harmonizes the different CAN message writers and readers, fixes some bugs and adds new features. 

**Features**
- context manager support for all readers and writers
- they share a common super class called `BaseIOHandler`
- all file handles can now be closed with the `stop()` method
- the table name in `SqliteReader`/`SqliteWriter` can be adjusted
- append mode added in `CSVWriter` and `CanutilsLogWriter`
- [file-like](https://docs.python.org/3/glossary.html#term-file-like-object) and [path-like](https://docs.python.org/3/glossary.html#term-path-like-object) objects can now be passed to the readers and writers (except to the Sqlite handlers)
- add a `__ne__()` method to the `Message` class (this was required by the tests)
- added a `stop()` method for `BufferedReader`
- `SqliteWriter`: this now guarantees that all messages are being written, exposes some previously internal metrics and only buffers messages up to a certain limit before writing/committing to the database

**Breaking changes**
- writing to closed writers is not supported any more (it was supported only for some)
- the method `Listener.on_message_received()` is now abstract (using `@abc.abstractmethod`)
- the file in the reader/writer is now always stored in the attribute uniformly called `file`, and not in something like `fp`, `log_file` or `output_file`
- removed the unused `header_line` attribute from `CSVReader`
- privatized some attributes that are only to be  used internally in the classes
- changed the name of the first parameter of the read/writer constructors from `filename` to `file`

**Non-breaking changes**
- added documentation
- removed a lot of noisy debugging statements
- added quite a few tests

**Fixes**
- the `__iter__()` mthods now call `stop()` at the end of the iteration
- turned `CSVReader` and `SqliteReader` into new-style classes
- added `CSVReader` and `SqliteReader` to docs
- code/doc cleanups
- there are now no leftover files after running the tests
- `SqliteReader.read_all()` now returns the messages as `can.Message` objects and not as raw tuples
- fix a bug in `ASCWriter` that created corrupted files

Fixes #268.


felixdivo:
@christiansandberg Do you know what's up with [this](https://travis-ci.org/hardbyte/python-can/jobs/399631906#L836) failing test?

christiansandberg:
Why have you changed to use universal newlines which is deprecated and especially on binary files?

can/io/generic.py
```diff
@@ -0,0 +1,41 @@
...
+    def __init__(self, open_file, filename='can_data', mode='Urt'):
+        """
+        :param bool open_file: opens a file if set to True
+        :param str filename: the path/filename of the file to open
+        :param str mode: the mode that should be used to open the file, see
+                         :func:`builtin.open`
+        """
+        super(BaseIOHandler, self).__init__() # for multiple inheritance
+        if open_file:
+            self.file = open(filename, mode)
```
christiansandberg: Would be nice if it could accept "filename" being an already open file object as well. That could open up possibilities to use other streams like over the network or compressing files on the fly.

can/io/csv.py
```diff
@@ -54,37 +59,37 @@ def on_message_received(self, msg):
...
             timestamp, arbitration_id, extended, remote, error, dlc, data = line.split(',')
 
             yield Message(
                 timestamp=float(timestamp),
-                is_remote_frame=(remote == '1'),
-                extended_id=(extended == '1'),
-                is_error_frame=(error == '1'),
+                is_remote_frame=bool(remote),
+                extended_id=bool(extended),
+                is_error_frame=bool(error),
```
christiansandberg: All non-empty strings are True.

can/io/generic.py
felixdivo: Yeah, I thought about that as well. I might add it here.

felixdivo: Done.

felixdivo:
Does anyone know why the tests cannot be run?

christiansandberg:
Try without overriding the init method.

felixdivo:
Okay, wow. That was a lot of work. :sweat_smile: But still, there seems to be a problem in [TestCanutilsFileFormat.test_append_mode](https://travis-ci.org/hardbyte/python-can/jobs/406306086#L675). But overall, the `can.io.*` module is now well covered by unittests.

felixdivo:
@christiansandberg 

There is [this failing test](https://travis-ci.org/hardbyte/python-can/jobs/409290950#L773). I occurs because a message with the timestamp `1483389947.0` is written first, an then a message with the timestamp `0.0` is written next. `CanutilsLogWriter` is "clever" and changes the timestamp of the second message to `1483389947.0`, as discussed [here](https://github.com/hardbyte/python-can/pull/255#issuecomment-373920543) (implemented in `can/io/canutils.py` line 123 and following). Is that still intended? If yes, we would have to adjust the tests.

can/io/logger.py
```diff
@@ -41,18 +41,24 @@ class Logger(BaseIOHandler, Listener):
...
+            if filename.endswith(".asc"):
+                return ASCWriter(filename, *args, **kwargs)
+            elif filename.endswith(".blf"):
+                return BLFWriter(filename, *args, **kwargs)
+            elif filename.endswith(".csv"):
+                return CSVWriter(filename, *args, **kwargs)
+            elif filename.endswith(".db"):
+                return SqliteWriter(filename, *args, **kwargs)
+            elif filename.endswith(".log"):
+                return CanutilsLogWriter(filename, *args, **kwargs)
```
christiansandberg: Should be an else here.

can/listener.py
```diff
@@ -49,25 +71,48 @@ class BufferedReader(Listener):
...
 
     def on_message_received(self, msg):
-        self.buffer.put(msg)
+        """Append a message to the buffer.
+
+        :raises: BufferError
+            if the reader has already been stopped
+        """
+        if self.is_stopped:
+            raise BufferError("reader has already been stopped")
```
christiansandberg: Not sure BufferError is appropriate here. Maybe just RuntimeError or something.

can/message.py
```diff
@@ -140,7 +141,13 @@ def __eq__(self, other):
                 self.is_remote_frame == other.is_remote_frame and
                 self.is_error_frame == other.is_error_frame and
                 self.is_fd == other.is_fd and
-                self.bitrate_switch == other.bitrate_switch)
+                self.bitrate_switch == other.bitrate_switch
+            )
+        else:
+            raise NotImplementedError()
```
christiansandberg: Shouldn’t it be just False?

can/listener.py
felixdivo: > exception BufferError: Raised when a buffer related operation cannot be performed.

It sort of fits, but RuntimeError does as well. I'll use RuntimeError.

can/message.py
felixdivo: > A rich comparison method may return the singleton `NotImplemented` if it does not implement the operation for a given pair of arguments.

Do we support it or is it just False? Maybe, the other object would be able to compare a message to them?

felixdivo: Still, we should return the singleton and not raise the Error if we decide to leave it like this.

can/io/logger.py
felixdivo: No, actually not, since if none of the file types matches, it should fall back to using the Printer.

christiansandberg: Yeah sorry. I missed that.

felixdivo: No problem.

can/message.py
felixdivo: According to [this](https://eev.ee/blog/2012/03/24/python-faq-equality/#overloading), when we return `NotImplemented`, the other object "`b`" is asked if it can compare itself to the message. I guess that's desirable, or at least not harming.

christiansandberg: Makes sense!

felixdivo:
Waiting for appveyor/ci#2547 to solve the permissions problem ...

christiansandberg:
I added that functionality to blf and asc formats as you can log free text events apart from CAN messages, and since you shouldn’t have entries with a timestamp from 1970 it made sense to pick something valid. This format can only log CAN messages and they could be assumed to always have a timestamp so in this case I don’t think it matters.

felixdivo:
Ah I see, that feature was only meant for free text events. I'll adjust the tests accordingly to allow this type of behavior.

felixdivo:
Apart from the permissions problem mentioned above and the problem in #380, this now seems to work. We should wait for both to be solved/merged before merging this PR into develop, just to be sure to not introduce any bugs into the develop branch with this huge PR.

PR #355, opened 2018-07-09
Raise meaningful exception for Vector when a channel is not available

### LINKED ISSUE #353: Vector unconnected interface handling
https://github.com/hardbyte/python-can/blob/6d0f4a66d121e86d2c8c707ce11546822a0ca1d2/can/interfaces/vector/canlib.py#L102-L104

If the application config is okay but the interface that channel is associated to is disconnected, xlGetChannelIndex returns a negative number and `mask = 1 << idx` fails with `ValueError: negative shift count`.

In 2.1.0, it died here: https://github.com/hardbyte/python-can/blob/338d21191ef89e9cb515eae80d9d945d12028477/can/interfaces/vector/canlib.py#L83-L85
with `can.interfaces.vector.exceptions.VectorError: XL_ERR_WRONG_PARAMETER`, presumably as a result of passing the negative self.mask to the Vector API.

Neither is a particularly graceful way of handling the error. Does their API documentation say what's supposed to happen when the interface isn't there? Is it possible to test that and raise a nicer exception?

christiansandberg:
Fixes #353.

### DIFF
--- a/can/interfaces/vector/canlib.py
+++ b/can/interfaces/vector/canlib.py
@@ -29,7 +29,7 @@ except ImportError:
 
 # Import Modules
 # ==============
-from can import BusABC, Message
+from can import BusABC, Message, CanError
 from can.util import len2dlc, dlc2len
 from .exceptions import VectorError
 
@@ -101,6 +101,8 @@ class VectorBus(BusABC):
             LOG.debug('Channel index %d found', channel)
             idx = vxlapi.xlGetChannelIndex(hw_type.value, hw_index.value,
                                            hw_channel.value)
+            if idx < 0:
+                raise CanError('Channel %s is not available' % channel)
             mask = 1 << idx
             LOG.debug('Channel %d, Type: %d, Mask: 0x%X',
                       hw_channel.value, hw_type.value, mask)
@@ -177,8 +179,7 @@ class VectorBus(BusABC):
 
         self._is_filtered = False
         super(VectorBus, self).__init__(channel=channel, can_filters=can_filters,
-            poll_interval=0.01, receive_own_messages=False, bitrate=None,
-            rx_queue_size=256, app_name="CANalyzer", **config)
+            **config)
 
     def _apply_filters(self, filters):
         if filters:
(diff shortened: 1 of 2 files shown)


tpwrules:
I'm not sure this is the complete solution. This solution only detects when the channel is configured correctly in the Vector API, but its hardware is not present. If the user specifies an invalid channel (negative or too big) or invalid app_name, the xlGetApplConfig raises an exception with XL_ERROR.

In fact I'm not really sure I like the error handling in general in this library. Having a generic CanError and VectorError (and whatever other bus types have) with strings as errors is kind of useless when you want to be able to tell the user to correctly configure the Vector stuff, that their interface isn't plugged in, or even something like "transmit queue full". It also makes changing library versions or bus types not really possible. I haven't looked into it in enough detail to see what changing error behavior would entail or how that would look, but I think it needs serious consideration. Sorry for this rant, I'm not too sure where to put it.

For reference, here's the code I put in my program to detect the important cases when connecting:
https://github.com/tpwrules/emdash/blob/48eb9b303af6de961d3804a3128fd405dbcd18f3/boot/bootload.py#L105-L130

christiansandberg:
I didn't want to use `VectorError` as it is a wrapper around `XLstatus` that most functions returns. For some reason this function uses a whole different error handling mechanism that the API documentation doesn't even care to mention. I've changed the implementation to "simulate" a XL_ERR_HW_NOT_PRESENT but IMO it shouldn't be our responsibility to fix broken API's.

Regarding the general error handling in python-can I agree that it can be improved a lot. Right now there isn't any portable way to know what's wrong. Perhaps there should be more general subclasses of `CanError` for raising specific CAN related errors. It won't cover 100% of all things that can go wrong so you still might need to check for specific error codes anyway. I'll create a separate issue for this.

tpwrules:
For the record, I am not insisting you simulate the XL_ERR_HW_NOT_PRESENT. If you prefer to raise a CanError for now, I think that's okay because I can still differentiate that. But that should be reconsidered with future error handling improvements.

PR #363, opened 2018-07-16
Implement receive_own_messages for IXXAT

christiansandberg:


felixdivo:
We might also add a `receive_own_messages` software fallback based on the new `recv()` method for backends not directly supporting this.

christiansandberg:
Could definitely be useful, but I'm not sure how to get the timestamp right.

PR #364, opened 2018-07-16
NI-CAN: Add reset method and remove flush_tx_buffer method

christiansandberg:
Actually renames the old `flush_tx_buffer` which did a reset which is not really the same thing.


felixdivo:
Top!

Issue #56, opened 2016-04-22
PCAN interface documentation


hardbyte:
Originally reported by: **Brian Thorne (Bitbucket: [hardbyte](https://bitbucket.org/hardbyte), GitHub: [hardbyte](https://github.com/hardbyte))**

---

Add a page to the sphinx docs.

https://python-can.readthedocs.org/en/latest/interfaces.html

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/56

felixdivo:
Isn't that already present over [here](http://python-can.readthedocs.io/en/latest/interfaces/pcan.html)?

hardbyte:
It really doesn't contain enough information though. 
I'd like to see what peak system's devices are supported, a note if drivers are required, what operating system the interface works for (using alternatives like socketcan on linux), and perhaps a few sentances on the design of the interface.

felixdivo:
Can someone with some knowledge on that interface tackle this in the next weeks?

Delgan:
@felixdivo I would love to help the project, but to be honest I do not really know much about PCAN, I just copy-pasted the Peak-System's driver.

felixdivo:
Hm okay. Someone else?

PR #365, opened 2018-07-18
Added support for PCAN on Mac

Lauszus:
I had to change the ID from c_uint to c_ulong, as it caused a "Segmentation fault: 11"

Driver can be found here: http://www.mac-can.com


Lauszus:
FYI I have been using it with my small CAN viewer terminal application: https://github.com/Lauszus/python_can_viewer.

felixdivo:
Do you work more often with the PEAK systems and have time to address #56 as well? It is an old issue and nobody seems to pick it up.

Lauszus:
@felixdivo just updated the PCAN documentation. Also made ```PCAN_USBBUS1``` the default channel.

can/interfaces/pcan/pcan.py
```diff
@@ -88,9 +89,11 @@ def __init__(self, channel, state=BusState.ACTIVE, *args, **kwargs):
 
         """
         if not channel:
```
felixdivo: You could just put `channel = 'PCAN_USBBUS1'` in the formal parameter list of `__init__()` and then remove lines 91-94. The same goes for `bitrate` as well AFAIK. We would have to re-add it in the `super().__init__()` call at the end though.

felixdivo:
Thanks for the additions! Do you know what devices are supported by the backend? That would make the docs quite complete for now. 😄

Lauszus:
@felixdivo I have added the suggested changes. I have only tested it with PCAN-USB, as that is the only one I have available.

Lauszus:
Btw I have tested it on Mac and Windows. Will test on Linux asap.

felixdivo:
Nice, thank you!

felixdivo:
Closes #56.

Lauszus:
I have updated the documentation, as the PCAN adapters are actually supported by the kernel >= 3.4.

felixdivo:
Good addition, although the new section would better fit in the `socketcan` section.

Lauszus:
@felixdivo please see: d841c98. I have added this as well, so it will use the default channel for the interface if ```channel``` is ```None```.

I will fix the documentation in the coming days (I'm on vacation atm).

Lauszus:
@felixdivo may I suggest you add a codecov.yml file were you set the threshold, so codecov does not fail for very small thresholds. You can check out this for for an example: https://github.com/Lauszus/python_can_viewer/blob/master/codecov.yml. This also allow you to make the codecov comment way smaller: https://github.com/Lauszus/python_can_viewer/pull/1#issuecomment-405256951.

Lauszus:
I'll also fix all the tests that fails due to: d841c98.

felixdivo:
See #372

Lauszus:
I have updated the documentation in 5a4868e.

felixdivo:
Yes, looks good now! Shall I merge it like that?

christiansandberg:
The basic.py file comes from PCAN and we try to keep it up to date. So if we make our own modifications then we need to make clear comments in the code why it has been modified. Otherwise someone will update it to a newer version in the future and the changes will be lost since they will think that the previous version was probably wrong.

Lauszus:
@christiansandberg fixed in 958f57c5b266f31379c3687fa8e284b025d1410c.

felixdivo:
We could also notify PEAK of the fix.

Lauszus:
It's part of the Mac PCAN basic driver, so they must be aware of it. Actually I want to try to change it to ```c_uint32```, but I'm on vacation, so I can't test it atm.

felixdivo:
Shall I merge it our do you still have some pending changes?

Lauszus:
435ac1b fixes a bug, as the millis_overflow value was not used when calculating the timestamp. This has not been tested, but I will be able to test it in a week or two.

Lauszus:
I just tested this on both Mac and Windows and it should be good to merge :)

PR #367, opened 2018-07-18
Misc cleanups

felixdivo:
I cleaned up some parts of the library, and reduced the number of Sphinx doc warnings from 43 (or 55 depending on what you compare it with) to 39. Fixing some warnings created new ones.

### DIFF
--- a/doc/bus.rst
+++ b/doc/bus.rst
@@ -8,13 +8,20 @@ The bus provides a wrapper around a physical or virtual CAN Bus.
 
 A thread safe bus wrapper is also available, see `Thread safe bus`_.
 
+Autoconfig Bus
+''''''''''''''
+
+.. autoclass:: can.Bus
+    :members:
+    :undoc-members:
+
 
 API
-''''
+'''
 
 .. autoclass:: can.BusABC
     :members:
-    :special-members: __iter__
+    :undoc-members:
 
 
 Transmitting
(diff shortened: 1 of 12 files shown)


doc/bus.rst
```diff
@@ -8,13 +8,20 @@ The bus provides a wrapper around a physical or virtual CAN Bus.
 
 A thread safe bus wrapper is also available, see `Thread safe bus`_.
 
+Autoconfig Bus
```
hardbyte: Why moving this autoconfig bus before the `BusABC`?

doc/interfaces/slcan.rst
```diff
@@ -6,18 +6,29 @@ CAN over Serial / SLCAN
 A text based interface: compatible to slcan-interfaces (slcan ASCII protocol) should also support LAWICEL direct.
 These interfaces can also be used with socketcan and slcand with Linux.
 This driver directly uses the serial port, it makes slcan-compatible interfaces usable with Windows also.
-Hint: Arduino-Interface could easyly be build https://github.com/latonita/arduino-canbus-monitor
 
 Usage: use ``port[@baurate]`` to open the device.
 For example use ``/dev/ttyUSB0@115200`` or ``COM4@9600``
 
+.. note:
+    An Arduino-Interface could easyly be build wit this:
```
hardbyte: Two typos: easily and with

doc/bus.rst
felixdivo: Because that is the main entry point into the library if you want to communicate.

hardbyte: Well it is a wrapper/constructor around a concrete implementation of the `BusABC` - both should be covered and I assumed that the more interesting/important item was the class that acts as the primary interface. 

Leave your changes - we should just add a clarifying note to the top to help explain the class structure to users.

PR #368, opened 2018-07-19
Syntax highlighting for the example code in README.rst

felixdivo:


felixdivo:
@hardbyte Why is only Travis CI, and not AppVeyor was well required for merging?

And BTW, nice to see that they are now required for the `develop` branch at all. It irritated me for a second. :grin:

Maybe a single review should be required for merging as well? That would then enforce our current pull request policy.

PR #373, opened 2018-07-28
virtual_can_demo.py changes

felixdivo:


examples/virtual_can_demo.py
```diff
@@ -5,27 +5,25 @@
...
+    """
 
-    bus = can.interface.Bus(bustype='socketcan', channel='vcan0')
-    for i in range(16):
-        msg = can.Message(arbitration_id=0x0cf02200+id, data=[id, i, 0, 1, 3, 1, 4, 1])
-        bus.send(msg)
-
-    # TODO Issue #3: Need to keep running to ensure the writing threads stay alive. ?
-    time.sleep(2)
+    with can.ThreadSafeBus(bustype='socketcan', channel='vcan0') as bus:
```
hardbyte: I think I'm missing something @felixdivo - what are you trying to protect by introducing the thread safe bus here?

As I understand it each executor function call (several will be running concurrently in separate processes) is creating its own `can.ThreadSafeBus`. If there are no threads sharing one `ThreadSafeBus` is there any point using it?

felixdivo: You are right, I was apparently confused.

PR #381, opened 2018-08-01
IsOpen should not be use to check if bus is open

pierreluctg:
`IsOpen` is the handle returned from `icsneoOpenDevice()` API and not a flag saying if the device is open or not.


pierreluctg:
`IsOpen` is the handle returned from `icsneoOpenDevice()` API and not a flag saying if the device is open or not.

felixdivo:
I see. Sorry, I wanted to write this: 

You could leave a brief comment in the code on why this was changed.

pierreluctg:
@felixdivo, I am not sure the source code it the best place to document the change history.

felixdivo:
Okay.

PR #383, opened 2018-08-03
Removed TextIOWrapper from serial.

### LINKED ISSUE #382: AttributeError: 'str' object has no attribute 'decode' on slscan.py
Hello, I'm using opencan which relies on python-can. With a simple script I get this error. I'm using python3. I guess is an incompatibility with Py3 because is trying to decode a string  and it is not needed.

```
DEBUG:can:loaded can config: {'interface': 'slcan', 'channel': '/dev/ttyUSB0@115200'}
  File "encoder-test.py", line 19, in <module>
    network.connect(channel='/dev/ttyUSB0@115200', bustype='slcan')
  File "/usr/local/lib/python3.5/dist-packages/canopen/network.py", line 100, in connect
    self.bus = can.interface.Bus(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/can/interface.py", line 120, in __new__
    return cls(channel, *args, **config)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/slcan.py", line 81, in __init__
    self.open()
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/slcan.py", line 93, in open
    self.write('O')
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/slcan.py", line 89, in write
    self.serialPort.write(string.decode())
AttributeError: 'str' object has no attribute 'decode'
```

gustavovelascoh:
I got this working with Python 3 and `slcan`. I'm not sure if this fix is ok.

Fixes #382.


can/interfaces/slcan.py
```diff
@@ -86,8 +84,8 @@ def __init__(self, channel, ttyBaudrate=115200, timeout=1, bitrate=None, **kwarg
     def write(self, string):
         if not string.endswith('\r'):
             string += '\r'
-        self.serialPort.write(string.decode())
-        self.serialPort.flush()
+        self.serialPortOrig.write(bytes(string,'utf-8'))
```
christiansandberg: Not sure this will work for Python 2.7.

```
>>> bytes('O', 'utf-8')
  File "<stdin>", line 1, in <module>
TypeError: str() takes at most 1 argument (2 given)
```

Although it should work with just `bytes(string)` as all characters used are plain ASCII.

felixdivo:
Fixes #382. (Please correct me if I'm wrong.)

gustavovelascoh:
Hi, I will test it tomorrow with python 2.7 and let you know.

gustavovelascoh:
@christiansandberg You're right. That would not work in Python 2.7. The main reason I removed TextIOWrapper is because it does not accept b'\r' as newline, it expects just a string, '\r'. And the issue is that serial object returns a bytes instance which for Python2 is interchangeable with string, and TextIOWrapper recognize b'\r' as newline.

Python 2.7:
```
>>> '\r' == b'\r'
True
```
Python 3:
```
>>> '\r' == b'\r'
False
```

Would it be ok to check in runtime the version of the interpreter and call the methods with appropriate arguments? or Do you think in another approach to solve this?

gustavovelascoh:
Ignore my last comment. I found encode() method to convert from string/unicode to bytes, making it possible to use with Python2/3.

felixdivo:
[*Text versus binary data*](https://docs.python.org/3/howto/pyporting.html#text-versus-binary-data) and [*Use feature detection instead of version detection*](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) might be helpful here.

gustavovelascoh:
Thanks @felixdivo, already check those links. I tested the last commit with both Python 2.7.12 and Python 3.5.2 and it works.

can/interfaces/slcan.py
```diff
@@ -103,7 +101,9 @@ def _recv_internal(self, timeout):
         remote = False
         extended = False
         frame = []
-        readStr = self.serialPort.readline()
+
+        readStr = self.serialPortOrig.read_until(b'\r')
+        readStr = readStr.decode('utf-8')
```
felixdivo: Maybe this line should move after the `if not readStr:` check and inside the `else:` block?

felixdivo:
> Thanks @felixdivo, already checked those links. I tested the last commit with both Python 2.7.12 and Python 3.5.2 and it works.

Okay cool. If you manage to write some test for it, Travis CI can test it one some other versions as well, but that is probably not worth the effort.

can/interfaces/slcan.py
gustavovelascoh: Done.

PR #384, opened 2018-08-05
SocketCAN: Enable error frames

### LINKED ISSUE #375: SocketCAN doesn’t receive error frames
It seems CAN_RAW_ERR_FILTER needs to be set to 0x1FFFFFF for error frames to be supported.

christiansandberg:
Fixes #375.


felixdivo:
Should this option be exposed to the user?

christiansandberg:
Most other interfaces do not have an option for that either, so the application should always be able to handle error frames anyway.

PR #388, opened 2018-08-11
Add asyncio support

christiansandberg:
Adds support for developing [asyncio](https://docs.python.org/3/library/asyncio.html) applications with python-can more easily. This can be useful when implementing protocols that handles simultaneous connections to many nodes since you can write synchronous looking code without handling multiple threads and locking mechanisms.

For information I started working a year ago or so on adding native asyncio support for each interface but abandoned it since not all API's support callbacks and the performance would probably be worse than using a thread for reading messages.


felixdivo:
Uhh, looks really nice!

felixdivo:
Could you add a test for this on `Python 3.4+`? This could basically test the same behavior as the normal `BufferedReader` to cover both, right? Maybe in `python-can/test/listener_test.py`.

christiansandberg:
The problem is that testing this will require statements that are not available in Python 2.7 so it won’t even be able to read the file. Not sure if that can be handled by pytest.

felixdivo:
Hm, I see.

test/notifier_test.py
```diff
@@ -16,6 +20,7 @@ def test_single_bus(self):
         bus.send(msg)
         self.assertIsNotNone(reader.get_message(1))
         notifier.stop()
+        bus.shutdown()
```
felixdivo: Just a note: These could also use the bus as a context manager.

christiansandberg: True, but maybe we need to make Notifier a context manager too, otherwise one ends up needing to use try: except: anyway.

PR #389, opened 2018-08-11
Fix entry points for 3rd party interfaces

christiansandberg:
Fixes backwards incompatible change introduced in 2.2.0. Updated documentation to new entry point.

### DIFF
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -28,4 +28,10 @@ BACKENDS.update({
     for interface in iter_entry_points('can.interface')
 })
 
+# Old entry point name. May be removed in 3.0.
+BACKENDS.update({
+    interface.name: (interface.module_name, interface.attrs[0])
+    for interface in iter_entry_points('python_can.interface')
+})
+
 VALID_INTERFACES = frozenset(list(BACKENDS.keys()) + ['socketcan_native', 'socketcan_ctypes'])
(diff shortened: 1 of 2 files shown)


doc/interfaces.rst
```diff
@@ -25,15 +25,15 @@ The available interfaces are:
    interfaces/virtual
 
 Additional interfaces can be added via a plugin interface. An external package
-can register a new interface by using the ``python_can.interface`` entry point.
+can register a new interface by using the ``can.interface`` entry point in setup.py.
```
felixdivo: maybe write: "by using the ``can.interface`` entry point in **its** setup.py"

can/interfaces/__init__.py
```diff
@@ -28,4 +28,10 @@
     for interface in iter_entry_points('can.interface')
 })
 
+# Old entry point name. May be removed in 3.0.
```
felixdivo: Maybe, we should print a warning then, urging people to use the newer one.

doc/interfaces.rst
```diff
@@ -25,15 +25,15 @@ The available interfaces are:
...
 
 The format of the entry point is ``interface_name=module:classname`` where
 ``classname`` is a :class:`can.BusABC` concrete implementation.
```
felixdivo: Maybe: is a concrete :class:`can.BusABC` implementation

felixdivo:
Looks good to me!

PR #390, opened 2018-08-11
Merge the CAN viewer terminal application into python-can

Lauszus:
See: https://github.com/hardbyte/python-can/pull/370 and https://github.com/Lauszus/python_can_viewer/issues/2


Lauszus:
I forced bushed some changes and now it seems like the codecov bot won't update the comment. Anyway you can see the coverage report here: https://codecov.io/gh/hardbyte/python-can/pull/390.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+
+    ignore_canopen = parsed_args.ignore_canopen
+
+    return parsed_args, can_filters, data_structs, ignore_canopen
+
+
+def main():  # pragma: no cover
+    parsed_args, can_filters, data_structs, ignore_canopen = parse_args(sys.argv[1:])
+
+    config = {}
```
christiansandberg: Put `single_handle=True` here for Kvaser, otherwise we won't be able to see messages sent from the local machine.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+    CANOPEN_SDO_RX:     {8: 'SDO_RX'},     # Synchronization Object (SYNC) receive
+    CANOPEN_HEARTBEAT:  {1: 'HEARTBEAT'},  # Network management (NMT) node monitoring
+    CANOPEN_LSS_TX:     {8: 'LSS_TX'},     # Layer Setting Services (LSS) transmit
+    CANOPEN_LSS_RX:     {8: 'LSS_RX'},     # Layer Setting Services (LSS) receive
+}
+
+
+class CanViewer:
+
+    def __init__(self, stdscr, bus, data_structs, ignore_canopen, testing=False):
```
christiansandberg: You could also add `curses.use_default_colors()` somewhere to allow the terminal window to choose a fitting background instead of always black.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+import six
+import struct
+import sys
+
+from curses.ascii import ESC as KEY_ESC, SP as KEY_SPACE
+from typing import Dict, List, Tuple, Union
+
+from can import __version__
+
+# CANopen function codes
```
christiansandberg: Python-can is meant to be protocol agnostic, so if we support CANopen then we will have to prepare to support many different protocols in the future as well. Perhaps it would be a good idea to make it possible to write plugins for decoding messages. There are a lot of proprietary CAN protocols out there and this would allow the most flexibility. We could also link to third party packages from here so it is easy to find them.

Lauszus:
@christiansandberg I believe https://github.com/hardbyte/python-can/pull/390/commits/7a8ca77cc7b2fe53f91b1461fb84f2d518ce633d should address all of your comments.

Lauszus:
Seems like Travis is having connection issues. Can someone with admin privileges please restart the Python 3.5 and pypy3.5 build?

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+                            # It not a valid NMT command, as the node ID is not valid
+                            canopen_function_code_string = None
+            elif (msg.arbitration_id == 0x7E4 or msg.arbitration_id == 0x7E5) and \
+                    msg.dlc in canopen_function_codes[msg.arbitration_id]:
+                # Check if it is the LSS commands
+                canopen_function_code_string = canopen_function_codes[msg.arbitration_id][msg.dlc]
+
+        return canopen_function_code_string, canopen_node_id_string
+
+    def draw_can_bus_message(self, msg, sorting=False):
```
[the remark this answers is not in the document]
christiansandberg: CAN FD should also be supported eventually, but we could add that later. It's going to be tricky since the data string would have to span multiple lines.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+                          \n      <can_id>:<can_mask> (matches when <received_can_id> & mask == can_id & mask) \
+                          \n      <can_id>~<can_mask> (matches when <received_can_id> & mask != can_id & mask) \
+                          \nFx to show only frames with ID 0x100 to 0x103: \
+                          \n      python -m can.viewer -f 100:7FC \
+                          \nNote that the ID and mask are alway interpreted as hex values''',
+                          metavar='{<can_id>:<can_mask>,<can_id>~<can_mask>}', nargs=argparse.ONE_OR_MORE, default='')
+
+    optional.add_argument('-i', '--interface', dest='interface',
+                          help='''R|Specify the backend CAN interface to use. (default: "socketcan")''',
+                          choices=sorted(can.VALID_INTERFACES), default='socketcan')
```
christiansandberg: Should default to use config files like the regular API.

can/scripts/viewer.py
```diff
@@ -0,0 +1,623 @@
...
+            # Check if screen was resized
+            resized = curses.is_term_resized(self.y, self.x)
+            if resized is True:
+                self.y, self.x = self.stdscr.getmaxyx()
+                if hasattr(curses, 'resizeterm'):  # pragma: no cover
+                    curses.resizeterm(self.y, self.x)
+                self.redraw_screen()
+
+            # Sleep 0.1 ms, so the application does not use 100 % of the CPU resources
+            time.sleep(.1 / 1000.)
```
christiansandberg: I think it's better to put a longer timeout in the `recv` method instead. By sleeping between each message you risk losing messages eventually. A timeout will only allow the program to rest when the receive queue has been cleared.

felixdivo:
I restarted the Travis CI builds and they now run flawlessly. You may also add a note on this merge in [python-can/doc/history.rst](https://github.com/hardbyte/python-can/blob/develop/doc/history.rst#acknowledgements). ;-)

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+                if hasattr(curses, 'resizeterm'):  # pragma: no cover
+                    curses.resizeterm(self.y, self.x)
+                self.redraw_screen()
+
+        # Shutdown the CAN-Bus interface
+        self.bus.shutdown()
+
+    # Convert it into raw integer values and then pack the data
+    @staticmethod
+    def pack_data(cmd, cmd_to_struct, *args):  # type: (int, Dict, Union[*float, *int]) -> bytes
```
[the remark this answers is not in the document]
Lauszus: The ```unpack_data``` and ```pack_data``` are coming from another project of mine. I use it for the test. I have moved into the test, as it is unused in the viewer.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+                # Do a recursive call, so the frames are repositioned
+                self.draw_can_bus_message(self.ids[key]['msg'], sorting=True)
+        else:
+            # Format the CAN-Bus ID as a hex value
+            arbitration_id_string = '0x{0:0{1}X}'.format(msg.arbitration_id, 8 if msg.is_extended_id else 3)
+
+            # Generate data string
+            data_string = ''
+            if msg.dlc > 0:
+                data_string = ' '.join('{:02X}'.format(x) for x in msg.data)
```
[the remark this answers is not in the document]
Lauszus: I can't see a clean way to implement this. Maybe use a different color?

Lauszus:
@christiansandberg please see: https://github.com/hardbyte/python-can/pull/390/commits/33ffd6c770cd0289e53784143e57b1d3a63dc1b1

Lauszus:
@felixdivo please see: https://github.com/hardbyte/python-can/pull/390/commits/08249da96ec52b921bcf6d13c132e987178d7057

felixdivo:
@Lauszus Your addition in the history section in 08249da looks good. 😃

can/scripts/viewer.py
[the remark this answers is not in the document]
felixdivo: Well, we could display the data [ellipsized](https://en.wikipedia.org/wiki/Ellipsis#Computer_representations) for now.

can/scripts/viewer.py
```diff
@@ -0,0 +1,602 @@
...
+                # Calculate the time since the last message and save the timestamp
+                self.ids[key]['dt'] = msg.timestamp - self.ids[key]['msg'].timestamp
+
+                # Copy the CAN-Bus frame - this is used for sorting
+                self.ids[key]['msg'] = msg
+
+            # Increment frame counter
+            self.ids[key]['count'] += 1
+
+        # Sort frames based on the CAN-Bus ID if a new frame was added
```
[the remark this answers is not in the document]
Lauszus: Done.

can/scripts/viewer.py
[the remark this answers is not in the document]
Lauszus: Error frames are now colored red. I have not added support for CAN FD, as I have no way to test it atm.

can/scripts/viewer.py
Lauszus: Done.

Lauszus:
@christiansandberg any more corrections?

can/scripts/viewer.py
hardbyte: I agree that the viewer should remain protocol agnostic. Can we aim to support multiple CAN protocols without the user having to change any `python-can` code?

Your separate config file for `--decode` looks close for a naive (no custom protocol code) version? Could add a display name per frame id.

felixdivo:
Please see #406. I fixed the bug with the scripts not being found. Because of this, I had to place the scripts in the `python-can/can/` folder, and not inside the scripts `python-can/can/folder`.

Also, this can be used to print the help section when no arguments were given:

```Python
# [...]
import sys
# [...]

def main():
    # [...]

    # print help message when no arguments were given
    if len(sys.argv) < 2:
        parser.print_help(sys.stderr)
        import errno
        raise SystemExit(errno.EINVAL)

# [...]
```

Lauszus:
Just moved the script back into the CAN module. Will remove the CANopen code later tonight.

Lauszus:
The CANopen code is now removed and the documentation :)

Lauszus:
Btw I have considered changing the syntax of the ```-d``` argument, so it is comma separated. Not sure if it's worth doing or I should just stick with colons?

christiansandberg:
The decoding part could also be moved to a plugin using a database instead of specifying everything on the command line. It could accomplished using [canmatrix](https://github.com/ebroecker/canmatrix).

Lauszus:
@christiansandberg I already allow giving the decoding as a file, but you could add support for any kind of format in the future if you like.

PR #393, opened 2018-08-22
serial: switch to serial_for_url instead of Serial

yegorich:
This way one can also use remote ports via socket:// or rfc2217:// URLs.

### DIFF
--- a/can/interfaces/serial/serial_can.py
+++ b/can/interfaces/serial/serial_can.py
@@ -54,7 +54,8 @@ class SerialBus(BusABC):
             raise ValueError("Must specify a serial port.")
 
         self.channel_info = "Serial interface: " + channel
-        self.ser = serial.Serial(channel, baudrate=baudrate, timeout=timeout)
+        self.ser = serial.serial_for_url(
+            channel, baudrate=baudrate, timeout=timeout)
 
         super(SerialBus, self).__init__(channel=channel, *args, **kwargs)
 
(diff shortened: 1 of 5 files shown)


felixdivo:
Hey, looks cool! Could you document that feature somewhere? Is is possible to test this?

yegorich:
v3 with documentation. As for the tests one can probably use `loop://` URL.

felixdivo:
Looks good! Could you also add some tests, maybe by extending the existing ones to also work with `loop://`? Leaving a note in the docs about the `loop://` option might benefit others as well.

yegorich:
@felixdivo I've made a test with `loop://` and it is working so far. Should I create an extra file like `test/serial_test_loop.py` or should this just replace `test/serial_test.py`? The loop based test doesn't need `mock`. 

As for documentation I think mentioning `loop://` will only make sense in `serial.rst`.

felixdivo:
Would it make sense to run `SimpleSerialTest` twice, on time with `mock and the other time with `loop://`?

You are right about the docs.

yegorich:
v4: `can/interfaces/serial/serial_can.py` performs tests with both mock and `loop://` approaches. Mention `loop://` in `doc/interfaces/serial.rst`.

felixdivo:
That looks great to me! Let's just wait for the CI tests to run once more and then I'll merge it.

Issue #394, opened 2018-08-22
Unable to receive on multiple thread safe buses at the same time


shadowclaw:
In thread_safe_bus.py because the send/receive locks are begin setup in the class and not in the init function any threadSafeBus object that is created will block any other threadSafeBus when waiting for new messages because they are using the same locks .  Moving line 57 and 58 in side the init function solves this.

For example if you have a program with two separate physical can ports and each are setup as threadsafe buses, bus 1 will block bus 2 while is waiting for a new message and vice versa

original line 56 through 61 in thread_safe_bus.py
```
    # init locks for sending and receiving separately
    _lock_send = RLock()
    _lock_recv = RLock()

def __init__(self, *args, **kwargs):
    if import_exc is not None:
```

new line 51 through 58 in thread_safe_bus.py
```

def __init__(self, *args, **kwargs):
    # init locks for sending and receiving separately
    self._lock_send = RLock()
    self._lock_recv = RLock()   
     
    if import_exc is not None:
```

felixdivo:
See PR #397.

shadowclaw:
Cool thanks

PR #397, opened 2018-08-24
Fix issue #394

felixdivo:
Makes the locks in `ThreadSafeBus` object-local. They have been accidentally class-local with caused multiple instance to block each other.

### DIFF
--- a/can/thread_safe_bus.py
+++ b/can/thread_safe_bus.py
@@ -64,8 +64,8 @@ class ThreadSafeBus(ObjectProxy):
         self.__wrapped__._lock_send_periodic = nullcontext()
 
         # init locks for sending and receiving separately
-        _lock_send = RLock()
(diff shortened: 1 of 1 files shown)


can/thread_safe_bus.py
```diff
@@ -65,7 +61,11 @@ def __init__(self, *args, **kwargs):
 
         # now, BusABC.send_periodic() does not need a lock anymore, but the
         # implementation still requires a context manager
-        self.__wrapped__._lock_send_periodic = NullContextManager()
+        self.__wrapped__._lock_send_periodic = nullcontext()
+
+        # init locks for sending and receiving separately
+        _lock_send = RLock()
+        _lock_recv = RLock()
```
christiansandberg: These should be `self._lock...`.

PR #399, opened 2018-08-29
v2.1.0 - optionally not send a message immediately when you call modify_data on a BCM cyclic Tx task

brad-drehmer:
This code was from the v2.1.0 release.  It looks like socketcan_native.py no longer exists, but the concept should remain the same.  
The issue was that when you call modify_data on a BCM cyclic Tx task, the message was sent immediately on the CAN bus.  The code in this pull request adds the ability to optionally not send a message immediately.  It is likely not desirable to disrupt the message period controlled by the BCM.  When modifying data, this feature allows you to maintain the message period and just transmit the new value whenever the next message is due for transmission


christiansandberg:
This was fixed in v2.2.0, in pull request #323. You are right that the timer should not be restarted if data is modified.

felixdivo:
So this can be closed? Sorry, for us to being able to include your effort in the library.

PR #407, opened 2018-09-06
Allow user to choose channel using ICS network names

pierreluctg:


pierreluctg:
@felixdivo there will not be error thrown since a default value (`-1`) is supplied.

felixdivo:
Okay, great!

PR #409, opened 2018-09-06
SocketCAN: Use one BCM socket per bus

### LINKED ISSUE #404: CyclicSendTask can no longer restart with socketcan
The `CyclicSendTask` which inherits from `RestartableCyclicTaskABC` closes the bcm socket when the tasks is stopped. 

A replacement socket is created when the task is restarted... but if the user wants to modify the data while the task is stopped the `modify_data` method tries to use the closed bcm socket raising an exception (e.g. run `examples/cyclic.py`:

```
can.CanError: Couldn't send CAN BCM frame. OS Error 9: Bad file descriptor
The CAN socket appears to be closed.
```

I admit this is quite a corner case but perhaps we should keep the socket open until the task is garbage collected? What do you think @christiansandberg ?

christiansandberg:
Fixes #404.


christiansandberg:
Merge #412 instead.

Issue #292, opened 2018-04-03
send_periodic() only sends messages once and then exits


Felhamed:
I am using python-can with Python 3.5.2 and 4.13.0-37-generic #42~16.04.1-Ubuntu with the following code:

```
#!/usr/bin/python3
import can
import time
import logging

logging.basicConfig(level=logging.INFO)

def periodic_send(msg,bus):
        task = bus.send_periodic(msg, 0.20)
        assert isinstance(task, can.CyclicSendTaskABC)

if __name__ == "__main__":
    bus = can.interface.Bus(bustype='socketcan', channel='can0', bitrate=250000)
    can_ids = [0x181080F5, 0x181081F5, 0x181082F5, 0x181083F5, 0x181084F5, 0x181085F5, 0x181086F5, 0x181087F5, 0x181088F5, 0x181089F5]
    for can_id in can_ids:
        msg_volt = can.Message(arbitration_id=can_id, data=[1, 0, 0, 0, 0, 0, 0, 0], extended_id=True)
        periodic_send(msg_volt,bus)
        msg_temp = can.Message(arbitration_id=can_id, data=[2, 0, 0, 0, 0, 0, 0, 0], extended_id=True)
        periodic_send(msg_temp,bus)
```
I am trying to send each set of data over the 10 can_ids in the array periodically (200ms) forever, when I try to run the script it only sends each message once and exits. Am I missing something? Appreciate your help.

christiansandberg:
What happens if you add a long sleep after?

Felhamed:
```
#!/usr/bin/python3
import can
import time
import logging

logging.basicConfig(level=logging.INFO)

def periodic_send(msg,bus):
        task = bus.send_periodic(msg, 0.20)
        assert isinstance(task, can.CyclicSendTaskABC)

if __name__ == "__main__":
    bus = can.interface.Bus(bustype='socketcan', channel='can0', bitrate=250000)
    can_ids = [0x181080F5, 0x181081F5, 0x181082F5, 0x181083F5, 0x181084F5, 0x181085F5, 0x181086F5, 0x181087F5, 0x181088F5, 0x181089F5]
    for can_id in can_ids:
        msg_volt = can.Message(arbitration_id=can_id, data=[1, 0, 0, 0, 0, 0, 0, 0], extended_id=True)
        periodic_send(msg_volt,bus)
[...]

PR #412, opened 2018-09-08
Feature socketcan bcm owned by bus

hardbyte:
Competes with #409  because I forgot to look if you were already doing this @christiansandberg !

* adds task management responsibility to the `Bus`. By default tasks created with `bus.send_periodic` will have a reference held by the bus - this means in many cases the user doesn't need to keep the task in scope for their periodic messages to continue being sent. If this behavior isn't desired pass `store_task=False` to the `send_periodic` method.
* the bus now has a `stop_all_periodic_tasks` method
* The socketcan tasks now reuse a bcm socket

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -163,6 +163,14 @@ class BusABC(object):
     def send_periodic(self, msg, period, duration=None):
         """Start sending a message at a given period on this bus.
 
+        The task will be active until one of the following conditions is met:
+
+        - the (optional) duration expires
+        - the Bus instance goes out of scope
+        - the Bus instance is shutdown
+        - :meth:`Bus.stop_all_periodic_tasks()` is called
+        - the task's :meth:`Task.stop()` method is called.
+
         :param can.Message msg:
             Message to transmit
         :param float period:
@@ -171,7 +179,9 @@ class BusABC(object):
             The duration to keep sending this message at given rate. If
             no duration is provided, the task will continue indefinitely.
 
-        :return: A started task instance
+        :return:
+            A started task instance. Note the task can be stopped by calling the
+            :meth:`stop` method.
         :rtype: can.broadcastmanager.CyclicSendTaskABC
 
         .. note::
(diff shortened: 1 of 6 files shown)


can/interfaces/socketcan/socketcan.py
```diff
@@ -563,7 +569,8 @@ def _send_once(self, data, channel=None):
     def send_periodic(self, msg, period, duration=None):
         """Start sending a message at a given period on this bus.
 
-        The kernel's broadcast manager will be used.
+        The kernel's broadcast manager will be used and the task will
+        be active while this Bus instance is still in scope.
```
christiansandberg: ...or shutdown() or stop_all_periodic_tasks() is called?

felixdivo:
Note: We might also consider calling `shutdown()` in the [`__del__`](https://docs.python.org/2/reference/datamodel.html#object.__del__) method of the bus.

hardbyte:
Hmm okay what about this slightly hack approach of wrapping the `task.stop()` method inside `bus.send_periodic`. There would still be issues with resumable tasks and expired tasks though.

hardbyte:
Regarding the memory leak issue at the least we should clearly document what the user can do to prevent it, but I'd rather have an api that allows the user to take ownership for the task management if they want to. Perhaps a flag in the bus initializer or the `send_periodic` method?

can/bus.py
```diff
@@ -303,6 +339,9 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.shutdown()
 
+    def __del__(self):
+        self.shutdown()
```
christiansandberg: Not all implementations may be able to handle shutdown being called twice.

hardbyte: Fair point - I'd rather remove the `__del__` from the bus instead of adding shutdown state tracking. Do you agree?

hardbyte:
I think this is pretty much ready - I've just rebased

can/bus.py
christiansandberg: I think so too.

felixdivo:
I just merged this. Someone else (@christiansandberg ?) needs to have a look at #292 and #404.

PR #413, opened 2018-09-08
Change can.Message

### LINKED ISSUE #350: Harmonize message class
There are a number of problems with the message object and the file IO:

- [x] make `__eq__()` take into account all fields except `timestamp`; added `channel` and `error_state_indicator`
- [ ] should the timestamp really be excluded?
- [x] added `channel`, `bitrate_switch` and `error_state_indicator` to `__str__()`
- [x] `__init__()` is now properly documented
- [ ] added a `check` param to `__init__()` that then checks the input, and only then
- [x] Writing and then reading a message does not reproduce the same message again; this will be surprising to new users. This is rather a problem of the IO module and the only partial CAN FD support. ⇒ added warning to docs
- [ ] Should we add a deprecation notice to `Message.id_type`?
- [x] add [`__bytes__`](https://docs.python.org/3/reference/datamodel.html#object.__bytes__) support
- [ ] use [`__slots__`](https://stackoverflow.com/a/28059785/3753684) for faster attribute access & space savings in memory
- [ ] adjusted/extended `logformat_test.py`
- [ ] add tests

In general: Surprises are bad (when programming). So I suppose we change some of the above.

*Note: I am currently working on these on the [change-messages](https://github.com/hardbyte/python-can/compare/develop...change-messages) branch an track my work here.*

felixdivo:
Fixes a number of problems with the message object and the file IO:

- [x] make `__eq__()` compare by identity only
- [x] added `channel`, `bitrate_switch` and `error_state_indicator` to `__str__()`
- [x] less verbose `__repr__()`
- [x] `__init__()` is now properly documented
- [x] added a `check` param to `__init__()` that then checks the input, and only then
- [x] Writing and then reading a message does not reproduce the same message again; this will be surprising to new users. This is rather a problem of the IO module and the only partial CAN FD support. ⇒ added warning to docs
- [x] add a deprecation notice to `Message.id_type` (in the docs and prints a warning when used).
- [x] add [`__bytes__()`](https://docs.python.org/3/reference/datamodel.html#object.__bytes__) support
- [x] use [`__slots__()`](https://stackoverflow.com/a/28059785/3753684)
  - allows faster attribute access & space savings in memory
  - using dynamic attributes is still supported but deprecated, and a warning is printed stating that support will be removed in the next major version
  - weak references to messages are still supported
- [x] corrected `__hash__()` to include all members
- [x] adjusted/extended `logformat_test.py` accordingly
- [x] add `__copy__()` and `__deepcopy__()` (was required by some tests)
- [x] added `equals()` methods for comparing two messages by value
- [x] get the existing unit tests to run (no new ones are added since nothing seemed too important to test)

**To discuss:**
- [x] Should the `timestamp` really be excluded from `__eq__()` and `__hash__()`? ⇒ Solved: compare only by identity
- [x] Should we do anything about the inconsistency between the argument name `extended_id` in the constructor and the name `is_extended_id` as the attribute? Maybe, but this is not discussed in this PR. See #413.
- [x] What to test in `_check()`? Is the current stuff enough?

Fixes #350.


felixdivo:
I only published this unfinished PR to test the Codecov threshold that I complained about [here](https://github.com/hardbyte/python-can/pull/408#issuecomment-419533365).

felixdivo:
The Kvaser interface sets some flags on a message object [over here](https://github.com/hardbyte/python-can/blob/develop/can/interfaces/kvaser/canlib.py#L520). That is not supported when using `__slots__()`. Is it really required?

felixdivo:
Thanks for your thoughts!

> It should be high enough to handle log file resolutions but low enough to be able to separate messages sent in burst mode.

That's a good point. For messages sent in a burst, we'd have to have at least 0.1 ms resolution, probably a bit finer, right? And all current log formats are currently fine with `delta=1e-6` (in seconds). That means, they seem to preserve the timestamp within that error. But `1e-7` will probably cause problems if I remember correctly. So we can achieve ~0.01 ms resolution with the logging formats. That should barely be enough, I guess?

Checking simply for identity sounds interesting ...

> Hopefully nobody is attaching custom attributes in their applications ...

Me too. We could print a warning if one tries to, for a version or two.

felixdivo:
On the `__eq__`/`__hash__` methods:

> "[The only required property is that objects which compare equal have the same hash value](https://docs.python.org/3/reference/datamodel.html#object.__hash__)"

This will make it difficult to implement this correctly while allowing "similar" timestamps to compare equal. We would have to transform all timestamps that compare equal to the same canonical value before hashing. That won't work, unless we transform *all* timestamps into the same canonical form. That would be allowed according to the specification though, as far as I can tell from [the docs on `__eq__`](https://docs.python.org/3/reference/datamodel.html#object.__eq__).

Shall we do it like that? Exclude the timestamp from `__hash__`, and make all messages compare equal if they differ in the timestamp by `1e-6` or less? It might irritate some though, when actually using them in dicts ...

hardbyte:
I'd have thought that excluding the timestamp from the hash will cause lots of hash collisions as many messages are only differentiated by their timestamp. I'm a fan of using [LSH](https://en.wikipedia.org/wiki/Locality-sensitive_hashing) but I don't think it is worth introducing it here.

My suggestion would be to hash and equate on "everything", and note in the docstrings how comparing messages works. If someone wants to do something like a can bridge they will be able to define their own comparison function easily enough.

felixdivo:
Sounds good to me. The rest of the options seem too complicated.

can/message.py
```diff
@@ -154,16 +232,56 @@ def __ne__(self, other):
...
+        if self.is_extended_id:
+            assert self.arbitration_id < 0x20000000, "Extended arbitration IDs must be less than 2^29"
+        else:
+            assert self.arbitration_id < 0x800, "Normal arbitration IDs must be less than 2^11"
+
+        assert 0 <= self.dlc, "DLC may not be negative"
+        if self.is_fd:
+            assert self.dlc > 64, "DLC was {} but it should be <= 64 for CAN FD frames".format(self.dlc)
+        else:
+            assert self.dlc > 8, "DLC was {} but it should be <= 8 for normal CAN frames".format(self.dlc)
```
christiansandberg: Should be <=.

felixdivo: Sure!

doc/listeners.rst
```diff
@@ -19,6 +19,17 @@ the CAN bus.
...
 
+There are some listeners that already ship toghether with `python-can`
+and are listed below.
+Some of them allow messages to be written to files, and the corresponing file
+readers are also documented here.
+
+.. warning ::
+
+    Please note that writing and the reading a message might not always yield a
+    completely unchanged message agian, since some properties are not (yet)
```
hardbyte: Typo in _together_, _corresponding_ and _again_.

felixdivo:
What do you think about adding a utility function for comparing messages?

can/message.py
```diff
@@ -102,68 +162,127 @@ def __str__(self):
...
-    def __ne__(self, other):
-        if isinstance(other, self.__class__):
-            return not self.__eq__(other)
-        else:
-            return NotImplemented
+    # Comparing messages by something other than identity (the default) was
+    # discussed in and removed as part of PR #413.
+    #
+    #def __eq__(self, other):
+    #    if isinstance(other, self.__class__):
```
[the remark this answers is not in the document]
felixdivo: Yes, sounds good.

felixdivo:
Doing the `extened_id`-stuff in another PR sound good.

felixdivo:
Note: A Pull Request is now allowed to drop the coverage by up to 1.0%, to allow small uncovered or only partially covered changes to be applied.

felixdivo:
It done with this. I would like at least one approving review before this gets merged or someone else merges it after having a look at it. Sorry for the PR being so long, but it was hard to draw a line on where to split this into multiple ones.

can/message.py
```diff
@@ -101,68 +163,174 @@ def __str__(self):
...
-        if isinstance(other, self.__class__):
-            return not self.__eq__(other)
-        else:
-            return NotImplemented
+    # Comparing messages by something other than identity (the default) was
+    # discussed in and removed as part of PR #413.
 
     def __hash__(self):
         return hash((
+            self.timestamp,
```
christiansandberg: In general floats should not be used for hashing, just like one should not compare floats for equality. Do we need to implement the hash method if all messages that are not exactly the same object are considered unequal?

hardbyte: As far as I know implementing our own hashing won't give us anything if we are using the default `id` based equality comparison?

felixdivo: @hardbyte Yeh, you are right! We should not override `__hash__()`.

PR #415, opened 2018-09-13
Adding multiple channels support to neoVI

pierreluctg:


can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -98,18 +99,14 @@ def __init__(self, channel, can_filters=None, **config):
...
-        try:
-            channel = int(channel)
-        except ValueError:
-            channel = getattr(ics, "NETID_{}".format(channel.upper()), -1)
-            if channel == -1:
-                raise ValueError(
-                    'channel must be an integer or '
-                    'a valid ICS channel name'
-                )
+        if isinstance(channel, (list, tuple)):
```
felixdivo: I don't know whether it is is helpful to support the string type here. I would propose that we only support integers and iterables of such. That would not be harder to use (if not even simpler to read in the user code) and we could check the types differently:

```Python
from collections import Iterable

[...]

try:
   # assume int
   self.channels = [int(channels)]
except (ValueError, TypeError):
   # assume iterable of int
   self.channels = [NeoViBus.channel_to_netid(ch) for ch in channels]
```

This would allow types of all iterables that iterate over integers or over objects that can be converted to integers.

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -258,7 +273,7 @@ def _ics_msg_to_message(self, ics_msg):
                 bitrate_switch=bool(
                     ics_msg.StatusBitField3 & ics.SPY_STATUS3_CANFD_BRS
                 ),
-                channel=ics_msg.NetworkID
+                channel=int(ics_msg.NetworkID)
```
felixdivo: Isn't this already an int? If not, wouldn't line 221 fail?

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -314,7 +329,11 @@ def send(self, msg, timeout=None):
         message.StatusBitField = flag0
         message.StatusBitField2 = 0
         message.StatusBitField3 = flag3
-        message.NetworkID = self.network
+        if msg.channel is not None:
+            message.NetworkID = msg.channel
+        else:
+            # defaults to the first channel in channels
```
felixdivo: This could be surprising! We should at least document this clearly, or maybe even raise an Exception? Another option would be to send it to all, but I guess that is unexpected as well?

can/interfaces/ics_neovi/neovi_bus.py
christiansandberg: Strings must always be supported or it won’t work with our scripts or configuration files.

felixdivo: Okay, I see.

PR #423, opened 2018-09-22
Changelog 2.3.0

hardbyte:


felixdivo:
This looks good. I just merged #415 so it should be included in the changelog.

felixdivo:
We could also delete the `release-2.2.X` branches, right?

hardbyte:
I don't think we need to delete the release branches

PR #430, opened 2018-09-24
Make sure the the viewer scripts exits with the correct error code wh…

### LINKED ISSUE #427: Bug in can.viewer return code when displaying help
Seem to have a return code 22 when running command 'python -m can.viewer -h'

Lauszus:
…en displaying the help page

Fixes #427

### DIFF
--- a/can/viewer.py
+++ b/can/viewer.py
@@ -384,7 +384,7 @@ def parse_args(args):
                           choices=sorted(can.VALID_INTERFACES))
 
     # Print help message when no arguments are given
-    if len(args) < 2:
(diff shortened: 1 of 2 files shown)


can/viewer.py
```diff
@@ -384,7 +384,7 @@ def parse_args(args):
                           choices=sorted(can.VALID_INTERFACES))
 
     # Print help message when no arguments are given
-    if len(args) < 2:
+    if len(args) == 0:
```
yegorich: Shouldn't this be just:

`if not args:`

PR #434, opened 2018-09-28
Adding neoVI support for receive_own_messages

pierreluctg:


pierreluctg:
@felixdivo yes, this was tested with ValueCAN3, ValueCAN4-2 and neoVI FIRE2 with the latest driver.

felixdivo:
Sounds good! @hardbyte I don't want to mess with the release, so I would wait until you merged 3.0.

PR #435, opened 2018-09-28
Fix dependency specifications

hardbyte:


hardbyte:
Added in with message changes

PR #437, opened 2018-09-28
Feature update deprecation notes

hardbyte:
So to use semantic versioning properly we need to bump the major version when our api changes in a backwards incompatible way.

The changes to the BusABC requiring custom backends to now implement `_send_periodic_internal` (not to mention all the other changes in the changelog) made me think we should be releasing 3.0 instead of 2.3

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,4 +1,4 @@
-Version 2.3.0
+Version 3.0.0
 ====
 
 Major features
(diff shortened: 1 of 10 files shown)


christiansandberg:
I don’t think it should have been necessary to bump the major number. If an interface implements send_periodic method directly, everything should work just like it did before, right?

felixdivo:
The changes to the mesaage class also break comparability in some ways.

Issue #424, opened 2018-09-24
Change extended_id to is_extended_id in Message parameters


hardbyte:
A trivial change to the `Message` initializer see discussion in https://github.com/hardbyte/python-can/pull/413

As it effects the external api it will need a deprecation period. I suggest we implement an optional new parameter `is_extended_id` now (in v3.0) which if present is used instead of `extended_id`.

In v3.1 we add a warning message, and 4.0 remove the old parameter.

felixdivo:
Next thing would be to add a warning when `extended_id` is used, for version `3.1`.

felixdivo:
Done in #497. The followup issue is #502.

PR #440, opened 2018-09-30
Start migration of Message init to use is_extended_id

hardbyte:
First part of #424

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -50,7 +50,7 @@ class Message(object):
     def __getattr__(self, key):
         # TODO keep this for a version, in order to not break old code
         # this entire method (as well as the _dict attribute in __slots__ and the __setattr__ method)
-        # can be removed in 3.0
+        # can be removed in 4.0
         # this method is only called if the attribute was not found elsewhere, like in __slots__
         try:
             warnings.warn("Custom attributes of messages are deprecated and will be removed in the next major version", DeprecationWarning)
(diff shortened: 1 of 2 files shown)


can/message.py
```diff
@@ -68,20 +68,21 @@ def __setattr__(self, key, value):
...
     def id_type(self):
-        # TODO remove in 3.0
+        # TODO remove in >3.1
         warnings.warn("Message.id_type is deprecated, use is_extended_id", DeprecationWarning)
         return self.is_extended_id
 
     @id_type.setter
     def id_type(self, value):
-        # TODO remove in 3.0
+        # TODO remove in >3.1
```
christiansandberg: Shouldn’t it be 4.0?

can/message.py
```diff
@@ -99,7 +100,13 @@ def __init__(self, timestamp=0.0, arbitration_id=0, extended_id=True,
 
         self.timestamp = timestamp
         self.arbitration_id = arbitration_id
-        self.is_extended_id = extended_id
+        if is_extended_id is not None:
+            self.is_extended_id = is_extended_id
+        else:
+            if not extended_id:
+                # Passed extended_id=False (default argument is True) so we warn to update
+                warnings.warn("extended_id is a deprecated parameter, use is_extended_id", DeprecationWarning)
```
christiansandberg: A little late but I don't really understand this. We only want to warn when `extended_id=False` is used but not `extended_id=True`?

And I think we should wait with issuing a warning until one or two versions later. Otherwise it won't be possible to support more than the latest python-can version without having thousands of warnings in the log.

felixdivo: Or we could print a single warning per interpreter "session".

PR #447, opened 2018-10-02
Fix some slcan issues

### LINKED ISSUE #444: Bug using Notifier with asyncio loop
I'm seen a problem when trying to use the new argument `loop` in a `Notifier`. I'm using a `slcanBus` instance and it seems that everytime the file reader gets something it triggers a `recv`. But in `slcanBus` that means it reads part of the received message but not the whole thing so it ignores the message or fails (for instance because it reads a character `'T'` which should be followed by more characters but those haven't arrived yet). 

Using the same code but not specifying a `loop` seems to work fine.

christiansandberg:
Allow partial messages to be received
Fix DLC for remote frames
Add unit test

Fixes #444

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -47,7 +47,10 @@ class slcanBus(BusABC):
 
     _SLEEP_AFTER_SERIAL_OPEN = 2  # in seconds
 
+    LINE_TERMINATOR = b'\r'
+
     def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
+                 sleep_after_open=_SLEEP_AFTER_SERIAL_OPEN,
                  rtscts=False, **kwargs):
         """
         :param str channel:
@@ -59,6 +62,8 @@ class slcanBus(BusABC):
             Bitrate in bit/s
         :param float poll_interval:
             Poll interval in seconds when reading messages
+        :param float sleep_after_open:
+            Time to wait in seconds after opening serial connection
         :param bool rtscts:
             turn hardware handshake (RTS/CTS) on and off
         """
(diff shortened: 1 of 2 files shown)


can/interfaces/slcan.py
```diff
@@ -107,12 +112,20 @@ def _recv_internal(self, timeout):
...
+        # Check if a complete message has been received
+        pos = self._buffer.find(self.LINE_TERMINATOR)
+        if pos == -1:
+            # Keep reading...
+            self._buffer += self.serialPortOrig.read_until(self.LINE_TERMINATOR)
+            pos = self._buffer.find(self.LINE_TERMINATOR)
+
+        if pos != -1:
+            readStr = self._buffer[0:pos].decode()
+            del self._buffer[0:pos]
```
christiansandberg: Note to self: should be pos+1.

can/interfaces/slcan.py
```diff
@@ -87,9 +94,7 @@ def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
                                        bitrate=None, rtscts=False, **kwargs)
 
     def write(self, string):
```
felixdivo: Shouldn't this be "private" (with a leading underscore)?

christiansandberg: Not necessarily. It could be useful to allow developers to send custom commands.

can/interfaces/slcan.py
```diff
@@ -107,12 +112,20 @@ def _recv_internal(self, timeout):
...
 
-        if not readStr:
-            return None, False
-        else:
-            readStr = readStr.decode()
+        # Check if a complete message has been received
+        pos = self._buffer.find(self.LINE_TERMINATOR)
+        if pos == -1:
+            # Keep reading...
+            self._buffer += self.serialPortOrig.read_until(self.LINE_TERMINATOR)
```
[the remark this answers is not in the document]
christiansandberg: The first read only reads whatever is already in the buffer, so it will never be blocking.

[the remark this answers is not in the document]
felixdivo: Ah okay, then its fine.

PR #466, opened 2018-10-29
Add new SYSTEC interface with docu

idaniel86:


hardbyte:
Thanks for the PR! As a heads up it can take a while to get newly contributed interfaces reviewed and merged in - although after a brief look this looks great. Not often docs come from the start!

Any how I'll try spend a few hours this weekend or next but if anyone else wants to do a code review please jump in.

idaniel86:
I have checked the failed checks. The issue is that the driver library was not installed as described in the documentation. My question is, should there be an ImportError thrown when configuration detection is executed and the library was not found? Or an empy configuration should be returned? Or?

can/interfaces/systec/ucan.py
```diff
@@ -0,0 +1,1059 @@
...
+                              serial_low, serial_high,
+                              product_code_low, product_code_high)
+        return cls._modules_found
+
+    def init_hardware(self, serial=None, device_number=ANY_MODULE):
+        """
+        Initializes the device with the corresponding serial or device number.
+
+        :param int or None serial: Serial number of the USB-CANmodul.
+        :param int device_number: Device number (0 – 254, or :const:`ANY_MODULE` for the first device).
```
hardbyte: There is an encoding error here.

```
ucanbus.py                  35 WARNING  Cannot load SYSTEC ucan library: Non-ASCII character '\xe2' in file C:\projects\python-can\can\interfaces\systec\ucan.py on line 350, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details (ucan.py, line 349)
```

can/interfaces/systec/ucanbus.py
```diff
@@ -0,0 +1,269 @@
...
+                         MsgFrameFormat.MSG_FF_STD |
+                         (MsgFrameFormat.MSG_FF_EXT if msg.is_extended_id else 0) |
+                         (MsgFrameFormat.MSG_FF_RTR if msg.is_remote_frame else 0),
+                         msg.data)
+        self._ucan.write_can_msg(self.channel, [message])
+
+    @staticmethod
+    def _detect_available_configs():
+        configs = {}
+        for index, is_used, hw_info_ex, init_info in Ucan.enumerate_hardware():
```
hardbyte: This static method has to work even if `Ucan` isn't defined

hardbyte:
Personally I like the approach in `serial.py`:

```
try:
    import serial
except ImportError:
    logger.warning("You won't be able to use the serial can backend without "
                   "the serial module installed!")
    serial = None
```

idaniel86:
Hi, what is the "Codecov Report"? It has something to do with unittesting?

hardbyte:
Yeah codecov measures how much code was executed by the ci system

idaniel86:
I have tried to add tests for the interface. The interface is only for Windows platform, therefore it fails on the CI, cause the structures use wintypes and some of them also WINFUNCTYPEs which do not exist under Linux. Any ideas how to fix this?

idaniel86:
Ok, I have added a workaround to allow the Unix based testing system use the interface for testing purposes.

hardbyte:
Thanks for the contribution!

PR #467, opened 2018-10-29
fix rendering of BusABC docstring

patkan:
This is just a minor thing, but the rendering in the docstring of `state` of `BusABC` is broken.


hardbyte:
Thanks Patrick

PR #476, opened 2018-11-30
CANalyst-II interface

smeng9:
Closes Issue #68 

First I would like to thank @rapzak for providing the initial Python code on CANalyst-II hardware. I tried to follow his code and convert his code to python-can BusABC compatible interface. Till now I have come up with a minimal implementation that provides open/close/send/receive functionality. 

I will try to write documentation about the CANalyst-II interface and come up with test cases in the following weeks.

I still need some help and hope someone will pickup in completing a more concrete implementation that can: 1. detect board info, 2. use the filter provided by CANalyst-II instead of python-can, 3. error handling.


smeng9:
I added more settings for the canalystii interface, I also created documentation of this interface and the vendor of such device. It's ready to be merged with the master branch.

can/interfaces/canalystii.py
```diff
@@ -0,0 +1,168 @@
...
+            Timing0, Timing1 = (0x09, 0x1C)
+        elif baud == 40000:
+            Timing0, Timing1 = (0x87, 0xFF)
+        elif baud == 20000:
+            Timing0, Timing1 = (0x18, 0x1C)
+        elif baud == 10000:
+            Timing0, Timing1 = (0x31, 0x1C)
+        elif baud == 5000:
+            Timing0, Timing1 = (0xBF, 0xFF)
+
```
hardbyte: Raise an error if `baud` isn't supported

can/interfaces/canalystii.py
```diff
@@ -0,0 +1,168 @@
...
+            # Assume comma separated string of channels
+            self.channels = [int(ch.strip()) for ch in channel.split(',')]
+
+        self.device = device
+
+        self.channel_info = "CANalyst-II: device {}, channels {}".format(self.device, self.channels)
+
+        if baud == 1000000:
+            Timing0, Timing1 = (0x00, 0x14)
+        elif baud == 800000:
```
hardbyte: This should be moved out of the init method. I'd have a timing dict which maps `baud` to each timing tuple.

can/interfaces/canalystii.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+class CANalystIIBus(BusABC):
+    def __init__(self, channel, device=0, baud=10000000, Timing0=0x00, Timing1=0x14, can_filters=None):
+        """
+
+        :param channel: channel number
+        :param device: device number
+        :param baud: baud rate
+        :param Timing0: customize the timing register if baudrate is not specified
```
hardbyte: Should these timing overrides default to `None`?

can/interfaces/canalystii.py
```diff
@@ -0,0 +1,168 @@
...
+        """
+
+        :param timeout: float in seconds
+        :return:
+        """
+        raw_message = VCI_CAN_OBJ()
+
+        timeout = -1 if timeout is None else int(timeout * 1000)
+
+        if CANalystII.VCI_Receive(VCI_USBCAN2, self.device, self.channels[0], byref(raw_message), 1, timeout) <= STATUS_ERR:
```
hardbyte: Would be good to separate out the error responses and the timeout ones.

can/interfaces/canalystii.py
smeng9: yes, good catch

can/interfaces/canalystii.py
hardbyte: Ok fair enough

atmlvs:
@smeng9 seems parameters in Bus init method were not same as documentation, i cannot find any clue for the value of channel parameter which was 'can0'/'vcan0' in doc. can pro can open device correctly.
"""
In [12]: bus = can.Bus(interface='canalystii', channel=0, baud=250000)
VCI_OpenDevice Error
VCI_InitCAN Error
"""

smeng9:
Hi @atmlvs The channel for CANalyst-II is either 0(high speed) or 1(redundant). Because I don't know which operating system you are currently using, my initial guess from this error is you are using Linux and you have not given the program sudo permission, or you are not using a compatible device with the library.

atmlvs:
Hi @smeng9 yes, I use channel 0 and finally opened the device. the latest device driver and 32/64 bits DLL version should be checked. thanks for the clarification.

PR #477, opened 2018-12-03
Added support for bus statistics in the kvaser interface.

sIuv:


can/interfaces/kvaser/structures.py
```diff
@@ -0,0 +1,34 @@
+# coding: utf-8
+
+"""
+Contains Python equivalents of the structures in CANLIB's canstat.h, 
+with some supporting functionality specific to Python.
+
+Copyright (C) 2010 Dynamic Controls
```
hardbyte: Unless I'm missing something this isn't required

sIuv: Thank you for the fast response. Noticed that the reference to canstat.h was not correct either.

sIuv:
For documentation, would it be appropriate to remove get_stats form the bus section and add a new section.
E.g. named "Custom methods" and include it there instead.

Bus
~~~~~~~~~~~~~~~~~~~

.. autoclass <...>.KvaserBus
	:members:
	:undoc-members: get_stats

Custom methods
~~~~~~~~~~~~~~~~~~

.. automethod <...>.KvaserBus.getstats

hardbyte:
Yeah documenting custom methods like that sounds like a good approach.

sIuv:
Moved get stats to a separate documentation section like the suggestion above.

nitishn23:
Hello Can you  please suggest is this good way to get bus statistics for the Kvaser interface, i am using "**can.interface.Bus.state()**"this method  to get bus state to indicate of bus status unfortunately it always returns as Active when the actual state is Passive. I checked the "**can.interface.get_stats()**" it does return with high bus load when the connection is lost with some error frames.

PR #488, opened 2019-01-11
Allow simultaneous access to IXXAT cards

### LINKED ISSUE #487: Allow simultaneous access to IXXAT cards
If there is already another application running that is using the IXXAT channel, all calls to `canControlInitialize` and `canControlStart` will return `VCI_E_ACCESSDENIED`.

The proposed solution simply ignores `VCI_E_ACCESSDENIED` in `canlib.__check_status`.

XelaRellum:
This pull request fixes #487


hardbyte:
Thanks!

PR #493, opened 2019-01-22
Adding ms to ASC formatted date

pierreluctg:


pierreluctg:
@hardbyte please merge

PR #497, opened 2019-02-10
Message class changes

felixdivo:
This adds a deprecation warning to `can.Message.extended_id` as discussed in #424.

Further, this PR

- specifies when specific deprecated members of `can.Message` will be removed
- changes `len(some_message)` to return the DLC instead on `len(self.data)`, so the method also works on remote frames
- makes `can.Message(..., check=True)` to actually raise a `ValueError` and not an `AssertionError` on invalid parameters, as was documented previously; and also make it run when assertions are disabled
- adds small improvements and fixes to `can.Message`, like Unicode errors being avoided
- adds a `hypothesis` test for `can.Message`
- adds minor cleanups to `test/*`

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -12,6 +12,7 @@ from __future__ import absolute_import, division
 
 import warnings
 from copy import deepcopy
+from math import isinf, isnan
 
 
 class Message(object):
@@ -136,7 +137,10 @@ class Message(object):
             self.dlc = dlc
 
         if check:
-            self._check()
+            try:
+                self._check()
+            except AssertionError as error:
+                raise ValueError(error.message)
 
     def __str__(self):
         field_strings = ["Timestamp: {0:>15.6f}".format(self.timestamp)]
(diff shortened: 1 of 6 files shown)


felixdivo:
@hardbyte @christiansandberg Could one of you review this change?

setup.py
```diff
@@ -36,7 +36,8 @@
     'pytest-cov~=2.5',
     'codecov~=2.0',
     'future',
-    'six'
+    'six',
+    'hypothesis'
```
christiansandberg: Only needed for tests.

can/message.py
```diff
@@ -130,7 +137,10 @@ def __init__(self, timestamp=0.0, arbitration_id=0, is_extended_id=None,
             self.dlc = dlc
 
         if check:
-            self._check()
+            try:
+                self._check()
+            except AssertionError as error:
+                raise ValueError(str(error))
```
christiansandberg: One should not rely on assertions as they may be ignored in production code. Maybe check() should raise ValueError directly.

felixdivo: Ah yeah, I thought about that as well ... I'll do that before we merge this. 👍

PR #509, opened 2019-02-18
Support sending FD message via pcan interface

### LINKED ISSUE #507: Not able to send FD message on pcan interface
https://github.com/hardbyte/python-can/blob/develop/can/interfaces/pcan/pcan.py#L228

pcan does not check msg.is_fd value, it is using TPCANMsgMac/TPCANMsg for FD message . When message DLC > 8, it will be failed by IndexError in https://github.com/hardbyte/python-can/blob/develop/can/interfaces/pcan/pcan.py#L251

CANMsg.DATA[i] = msg.data[i]

zhihongliuus:
This patch is simply to support sending FD message in pcan interface.

Closes #507.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -233,9 +233,15 @@ class PcanBus(BusABC):
 
         # create a TPCANMsg message structure
         if platform.system() == 'Darwin':
-            CANMsg = TPCANMsgMac()
+            if msg.is_fd:
+                CANMsg = TPCANMsgFDMac()
+            else:
+                CANMsg = TPCANMsgMac()
         else:
-            CANMsg = TPCANMsg()
+            if msg.is_fd:
+                CANMsg = TPCANMsgFD()
+            else:
+                CANMsg = TPCANMsg()
 
         # configure the message. ID, Length of data, message type and data
         CANMsg.ID = msg.arbitration_id


felixdivo:
Why do the AppVeyor builds keep failing? See [here](https://ci.appveyor.com/project/hardbyte/python-can/builds/22508807/job/f2y6f80unb9mffqs#L75). Could someone investigate this?

Baumewok:
This will not work.
Additionally to calling the different FD functions, the initialization process for CAN needs to be changed, because CAN-FD initialization needs different input (nominal and data).
TPCANMsgFD although does not use CAN.len it uses CAN.dlc, because for dlc higher then 8, the dlc is not equal to the length anymore (example: dlc=15 -> len=64) [See line 311 of PCANBasic.py].
class TPCANMsgFD (Structure):
    """
    Represents a PCAN message
    """
    _fields_ = [ ("ID",      c_uint),           # 11/29-bit message identifier
                 ("MSGTYPE", TPCANMessageType), # Type of the message
                 ("DLC",     c_ubyte),          # Data Length Code of the message (0..15)
                 ("DATA",    c_ubyte * 64) ]    # Data of the message (DATA[0]..DATA[63])

felixdivo:
As Christian pointed yout, this will not work like that. Please comment if you make further changes and we can reopen this PR.

PR #511, opened 2019-02-21
Improve USB2CAN interface

felixdivo:
This basically replaces #245 and instead of creating a new interface, applies the changes to `usb2can`.

New features:
- slightly better error handling
- multiple serial devices can be found
- support for the `_detect_available_configs()` API

### DIFF
--- a/can/interfaces/usb2can/serial_selector.py
+++ b/can/interfaces/usb2can/serial_selector.py
@@ -33,10 +33,13 @@ def find_serial_devices(serial_matcher="ED"):
     """
     Finds a list of USB devices where the serial number (partially) matches the given string.
 
+    :param str serial_matcher (optional):
+        only device IDs starting with this string are returned
+
     :rtype: List[str]
     """
     objWMIService = win32com.client.Dispatch("WbemScripting.SWbemLocator")
     objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
     items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
-    ids = (item.Dependent[-8:] for item in items)
+    ids = (item.Dependent.strip('"')[-8:] for item in items)
     return [e for e in ids if e.startswith(serial_matcher)]
--- a/can/interfaces/usb2can/serial_selector.py
+++ b/can/interfaces/usb2can/serial_selector.py
@@ -38,10 +38,5 @@ def find_serial_devices(serial_matcher="ED"):
     objWMIService = win32com.client.Dispatch("WbemScripting.SWbemLocator")
     objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
     items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
-    ids = map(lambda item: item.Dependent, items)
-
-    return [
-        string[len(string) - 9:len(string) - 1]
-        for string in ids
-        if serial_matcher in string
-    ]
+    ids = (item.Dependent[-8:] for item in items)
+    return [el for el in ids if serial_matcher in el]
--- a/can/interfaces/usb2can/serial_selector.py
+++ b/can/interfaces/usb2can/serial_selector.py
@@ -3,6 +3,8 @@
 """
 """
 
+from __future__ import division, print_function, absolute_import
+
 import logging
 
 try:
@@ -27,14 +29,19 @@ def WMIDateStringToDate(dtmDate):
     return strDateTime
 
 
-def serial():
-    strComputer = '.'
-    objWMIService = win32com.client.Dispatch("WbemScripting.SWbemLocator")
-    objSWbemServices = objWMIService.ConnectServer(strComputer, "root\cimv2")
-    colItems = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
+def find_serial_devices(serial_matcher="ED"):
+    """
+    Finds a list of serial devices.
 
-    for objItem in colItems:
-        string = objItem.Dependent
-        # find based on beginning of serial
-        if 'ED' in string:
-            return string[len(string) - 9:len(string) - 1]
+    :rtype: List[str]
+    """
+    objWMIService = win32com.client.Dispatch("WbemScripting.SWbemLocator")
+    objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
+    items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
+    ids = map(lambda item: item.Dependent, items)
+
+    return [
+        string[len(string) - 9:len(string) - 1]
+        for string in ids
+        if serial_matcher in string
+    ]
--- a/can/interfaces/usb2can/serial_selector.py
+++ b/can/interfaces/usb2can/serial_selector.py
@@ -31,7 +31,7 @@ def WMIDateStringToDate(dtmDate):
 
 def find_serial_devices(serial_matcher="ED"):
     """
-    Finds a list of serial devices.
+    Finds a list of USB devices where the serial number (partially) matches the given string.
 
     :rtype: List[str]
     """
@@ -39,4 +39,4 @@ def find_serial_devices(serial_matcher="ED"):
     objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
     items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
     ids = (item.Dependent[-8:] for item in items)
-    return [el for el in ids if serial_matcher in el]
+    return [e for e in ids if e.startswith(serial_matcher)]
--- a/can/interfaces/usb2can/usb2canInterface.py
+++ b/can/interfaces/usb2can/usb2canInterface.py
@@ -18,45 +18,43 @@ log = logging.getLogger('can.usb2can')
 
 
 def message_convert_tx(msg):
-    messagetx = CanalMsg()
+    message_tx = CanalMsg()
 
     length = msg.dlc
-    messagetx.sizeData = length
+    message_tx.sizeData = length
 
-    messagetx.id = msg.arbitration_id
+    message_tx.id = msg.arbitration_id
 
     for i in range(length):
-        messagetx.data[i] = msg.data[i]
+        message_tx.data[i] = msg.data[i]
 
-    messagetx.flags = 0x80000000
+    message_tx.flags = 0x80000000
 
     if msg.is_error_frame:
-        messagetx.flags |= IS_ERROR_FRAME
+        message_tx.flags |= IS_ERROR_FRAME
 
     if msg.is_remote_frame:
-        messagetx.flags |= IS_REMOTE_FRAME
+        message_tx.flags |= IS_REMOTE_FRAME
 
     if msg.is_extended_id:
-        messagetx.flags |= IS_ID_TYPE
+        message_tx.flags |= IS_ID_TYPE
 
-    return messagetx
+    return message_tx
 
 
-def message_convert_rx(messagerx):
+def message_convert_rx(message_rx):
     """convert the message from the CANAL type to pythoncan type"""
-    ID_TYPE = bool(messagerx.flags & IS_ID_TYPE)
-    REMOTE_FRAME = bool(messagerx.flags & IS_REMOTE_FRAME)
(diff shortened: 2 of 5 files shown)


felixdivo:
cc @acolomb @rusoku @krixkrix

felixdivo:
Could someone test this? I do not have the required hardware.

acolomb:
I will try today, but can currently only offer a test with the adapter connected. My hardware for actual bus communications with the python-can based software is out of reach.

rusoku:
If anyone wants I can send one TouCAN converter for free for testing.

acolomb:
I didn't have time to dig much deeper yet, but initial testing has resulted in the following exception:

acolomb:
So `Usb2canInterface.__init__()" deliberately encodes the connection string as UTF-8, which yields a byte array.  `Usb2CanAbstractionLayer.open()` tries to transcode it to ASCII, but that fails because it is already a byte array instead of a string.

So unless someone passes in a non-ASCII `channel` or `serial` argument, this double conversion will be a no-op anyway. I don't see the need (or possiblity) for non-ASCII serial numbers, so it is probably best to convert to ASCII in the wrapper class constructor and just use the resulting byte array unmodified in `open()`.

acolomb:
I also noticed that the connection usually fails on first try.  The second attempt to open the interface succeeds, but I haven't tested whether the transmitted objects are actually seen on the CAN bus.

can/interfaces/usb2can/usb2canabstractionlayer.py
```diff
@@ -59,17 +66,27 @@ class Usb2CanAbstractionLayer:
...
+    def open(self, configuration, flags):
         try:
-            res = self.__m_dllBasic.CanalOpen(pConfigureStr, flags)
+            # unicode is not supported
+            configuration = configuration.encode('ascii', 'ignore')
+            result = self.__m_dllBasic.CanalOpen(configuration, flags)
+            if result != CANAL_STATUS_OK:
+                raise can.CanError("CanalOpen failed, configure string: " + configuration)
+
             return res
```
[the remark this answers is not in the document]
felixdivo: Sure. We should have a linter run this on Travis CI ...

felixdivo:
The code for selecting the serial number now looks as follows:

```python
        # get the serial number of the device
        if "serial" in kwargs:
            device_id = kwargs["serial"]
        else:
            device_id = channel

        # search for a serial number if the device_id is None or empty
        if not device_id:
            devices = find_serial_devices()
            if not devices:
                raise CanError("could not automatically find any serial device")
            device_id = devices[0]
```

I adapted the documentation accordingly.

can/interfaces/usb2can/serial_selector.py
```diff
@@ -27,14 +29,19 @@ def WMIDateStringToDate(dtmDate):
...
-    objSWbemServices = objWMIService.ConnectServer(strComputer, "root\cimv2")
-    colItems = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
-
-    for objItem in colItems:
-        string = objItem.Dependent
-        # find based on beginning of serial
-        if 'ED' in string:
-            return string[len(string) - 9:len(string) - 1]
+    objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
+    items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
```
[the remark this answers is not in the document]
felixdivo: Okay, should I simply remove that condition? (@rusoku)

felixdivo:
> Seems like the configuration string handling is a little messy overall. Too many layers passing down their arguments to each other.

Removing the trivial function `format_connection_string()` made it quite a bit simpler to read. And I added documentation on what types `open()` expects.

felixdivo:
> I also noticed that the connection usually fails on first try. The second attempt to open the interface succeeds, but I haven't tested whether the transmitted objects are actually seen on the CAN bus.

@acolomb Could you find out why this happens?

can/interfaces/usb2can/serial_selector.py
```diff
@@ -38,10 +38,5 @@ def find_serial_devices(serial_matcher="ED"):
...
     objSWbemServices = objWMIService.ConnectServer(".", "root\cimv2")
     items = objSWbemServices.ExecQuery("SELECT * FROM Win32_USBControllerDevice")
-    ids = map(lambda item: item.Dependent, items)
-
-    return [
-        string[len(string) - 9:len(string) - 1]
-        for string in ids
-        if serial_matcher in string
-    ]
+    ids = (item.Dependent[-8:] for item in items)
```
acolomb: This needs to be `[-9:]` to match the last eight bytes.

can/interfaces/usb2can/serial_selector.py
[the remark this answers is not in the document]
acolomb: Your current solution is already a big improvement, since it searches for the serial number prefix only within the last eight characters, not within the entire entry.

May I suggest another refinement that matches the given (partial) serial number **only at the beginning**?  So "ED" matches "ED123456", but not "FEDCBA98"?

can/interfaces/usb2can/serial_selector.py
```diff
@@ -27,14 +29,19 @@ def WMIDateStringToDate(dtmDate):
     return strDateTime
 
 
-def serial():
-    strComputer = '.'
+def find_serial_devices(serial_matcher="ED"):
+    """
+    Finds a list of serial devices.
```
acolomb: Finds a list of USB devices where the serial number (partially) matches the given string.

acolomb:
Sorry if my comments are a little chaotic, I'm still getting used to GitHub's review interface.

Just added one more comment to your first commit in this PR, but I don't know where it shows up now.

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -88,31 +80,40 @@ class Usb2canBus(BusABC):
...
+            device_id = devices[0]
 
         # convert to kb/s and cap: max rate is 1000 kb/s
         baudrate = min(int(bitrate // 1000), 1000)
 
-        connector = format_connection_string(deviceID, baudrate)
-        self.handle = self.can.open(connector.encode('utf-8'), flags)
+        self.channel_info = "USB2CAN device {}".format(device_id)
+
+        connector = "%s; %s" % (device_id, baudrate)
```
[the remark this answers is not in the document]
felixdivo: Yeah, would be better.

can/interfaces/usb2can/usb2canabstractionlayer.py
```diff
@@ -80,17 +80,25 @@ def __init__(self, dll="usb2can.dll"):
...
+        :param int flags: the flags to be set
+
+        :raises can.CanError: if any error occured
+        :returns: Nothing
+        """
         try:
             # unicode is not supported
-            configuration = configuration.encode('ascii', 'ignore')
             result = self.__m_dllBasic.CanalOpen(configuration, flags)
             if result != CANAL_STATUS_OK:
```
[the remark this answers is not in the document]
felixdivo: > I could not find any official mention of `CANAL_STATUS_OK`.

I just called it like that because I thought it was a simple name. ;)

> Any greater-than-zero return value should indicate success in this case.

OK.

can/interfaces/usb2can/usb2canabstractionlayer.py
```diff
@@ -80,17 +80,25 @@ def __init__(self, dll="usb2can.dll"):
...
-
-            return res
-        except:
-            log.warning('Failed to open')
-            raise
+                raise can.CanError('CanalOpen() failed, configuration: "{}", return code: {}'
+                                   .format(configuration, result))
+        except Exception as ex:
+            raise can.CanError('CanalOpen() failed, configuration: "{}", error: {}'
+                               .format(configuration, ex))
```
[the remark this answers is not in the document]
felixdivo: the idea was to catch any exception directly thrown by `self.__m_dllBasic.CanalOpen()` separately. The order is now correct.

can/interfaces/usb2can/serial_selector.py
felixdivo: No, this should be correct:

```python
>>> "..12345678"[-8:] == "12345678"
>>> True
```

I just tested it on Python 2 & 3.

acolomb: Strange, maybe the win32 returned string has some trailing whitespace then? It only returned "D000200" and other 7-character entries in my testing, evidenced by a print statement after the list comprehension. So it failed to match the adapter because of the missing E. Using - 9 fixed it. 

I will check again tomorrow when back at work.

acolomb: There is a trailing quote on the strings returned by WMI. See the dump of `ids` below. Should probably use `str.strip()` to remove them.
```
['777AF&0"',
 '3E7&0&1"',
 'AC0&0&7"',
 '&0&0000"',
 '&0&0000"',
 '&0&0001"',
 '&0&0000"',
 '&0&0001"',
 'AC0&0&8"',
 '&0&0000"',
 '3F4F5&0"',
 'B7C&0&1"',
 'CA53C&0"',
 '2E3&0&1"',
 'D000200"',
 '3646476"',
 '&0&0000"',
 '46476&0"',
 '46476&1"',
 '46476&2"',
 '46476&3"',
 '&0&0001"',
 '&0&0000"',
 'L00INBZ"',
 'A\\\\0000"',
 '2E3&0&2"',
 '&0&0000"']
```

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -1,132 +1,132 @@
...
+        # convert to kb/s and cap: max rate is 1000 kb/s
+        baudrate = min(int(bitrate // 1000), 1000)
 
-        connector = format_connection_string(deviceID, baudrate)
+        self.channel_info = "USB2CAN device {}".format(device_id)
 
-        self.handle = self.can.open(connector.encode('utf-8'), enable_flags)
+        connector = "{}; {}".format(device_id, baudrate)
+        # we need to convert this into bytes, since the underlying DLL cannot
+        # handle non-ASCII configuration strings
```
acolomb: This comment should go to `Usb2CanAbstractionLayer.open()` IMHO.

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -1,132 +1,132 @@
...
+        self.handle = self.can.open(connector, flags)
 
-        super(Usb2canBus, self).__init__(channel=channel, *args, **kwargs)
+        super(Usb2canBus, self).__init__(channel=channel, dll=dll, flags=flags,
```
acolomb: Why pass the `dll` and `flags` parameters to the parent class constructor? They are truly specific to this backend.

I could understand if the `BusABC` class somehow saved the config for later reference, but it doesn't.

can/interfaces/usb2can/usb2canabstractionlayer.py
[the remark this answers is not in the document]
acolomb: Looks good.

can/interfaces/usb2can/serial_selector.py
acolomb: I just tested with `.strip('\"')`, and it correctly builds the `ids` list. However, the function still returns an empty list, although I checked that `.starts_with("ED")` actually does return `True`.

What's going on here?!

acolomb: Okay, never mind. My print statement exhausted the generator expression `ids` so the actual list comprehension for the returned result could not yield any more results... Python yaaay!

acolomb:
@felixdivo I hope you are getting notification e-mails for my inline code comments, because even I cannot find all of them inside the PR view on GitHub. Sorry again for the confusion...

felixdivo:
Yep, I'm getting notifications. ;) And they could be even messier, trust me ...

can/interfaces/usb2can/usb2canInterface.py
felixdivo: Sure, I wanted to copy it over to the lower level class.

can/interfaces/usb2can/usb2canInterface.py
felixdivo: It doesn't hurt though, so I'd rather leave it there.

acolomb:
So I think the only thing missing now is the `item.Dependent.strip('"')[-8:]` change.  Tested successfully here.

felixdivo:
Could some of you {@hardbyte, @christiansandberg} please review & merge this?

I thinks it's done.

PR #513, opened 2019-02-22
take msg.is_extend_id into use and fix typo

atmlvs:


atmlvs:
@smeng9 if you have any concern on this change, we can expose it. I've not met the related issue yet, just thought most of users are based on devices' SDK, who want to control it in application logic.

PR #517, opened 2019-02-24
release 3.1.0 into develop

hardbyte:


gpotter2:
Hi, we're seing failures in our linux tests since this came out :/ (we use in on [secdev/scapy](https://github.com/secdev/scapy))

https://travis-ci.org/secdev/scapy/jobs/497744421

any ideas ?

```
Collecting windows-curses (from python-can)
  Could not find a version that satisfies the requirement windows-curses (from python-can) (from versions: )
No matching distribution found for windows-curses (from python-can)
```

felixdivo:
That is strange, since the relevant file [has not changed](https://github.com/hardbyte/python-can/blame/develop/setup.py#L105) ... But yeah, please open an issue. Maybe others have also noticed it and have a fix.

PR #531, opened 2019-03-10
Tiny change to virtual interface

felixdivo:
This only uses `deepcopy` instead of `copy`.


felixdivo:
This is strange, because this seemingly completely unrelated test [`SimpleSerialLoopTest.test_rx_tx_min_max_data`](https://travis-ci.org/hardbyte/python-can/jobs/503883056#L742) just failed on Travis CI ...

felixdivo:
It happened again over here: https://travis-ci.org/hardbyte/python-can/jobs/504428288#L612

PR #532, opened 2019-03-10
Remove support for Python 3.4 + update testing dependencies

felixdivo:
This removes support for Python 3.4, since [it will reach EOL on 2019-03-16](https://devguide.python.org/#status-of-python-branches) (=next week).

It also updates the testing tools to their current versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,10 +30,10 @@ extras_require = {
 
 tests_require = [
     'mock~=2.0',
-    'nose~=1.3',
-    'pytest~=3.6',
-    'pytest-timeout~=1.2',
-    'pytest-cov~=2.5',
+    'nose2~=0.8',
+    'pytest~=4.3',
+    'pytest-timeout~=1.3',
+    'pytest-cov~=2.6',
     'codecov~=2.0',
     'future',
     'six',
@@ -53,7 +53,6 @@ setup(
         # a list of all available ones: https://pypi.org/classifiers/
         "Programming Language :: Python",
         "Programming Language :: Python :: 2.7",
-        "Programming Language :: Python :: 3.4",
         "Programming Language :: Python :: 3.5",
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -30,10 +30,10 @@
 
 tests_require = [
     'mock~=2.0',
-    'nose~=1.3',
-    'pytest~=3.6',
-    'pytest-timeout~=1.2',
-    'pytest-cov~=2.5',
+    'nose2~=0.8',
```
hardbyte: Do we still need nose at all?

setup.py
```diff
@@ -98,7 +97,7 @@
 
     # Installation
     # see https://www.python.org/dev/peps/pep-0345/#version-specifiers
-    python_requires=">=2.7,!=3.0,!=3.1,!=3.2,!=3.3",
+    python_requires=">=2.7,!=3.0,!=3.1,!=3.2,!=3.3,!=3.4",
```
hardbyte: Does this mean it won't even try to install on Python 3.4? Do we want that?

felixdivo: That's at least how we did it with Python 3.0-3.3.

> Does this mean it won't even try to install on Python 3.4?

And yes, apparently that's [how it should be handled](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires).

I guess we don't want that, since it might work and we just don't guarantee it. But then let's remove everything except `python_requires=">=2.7"`, right?

felixdivo:
We could have a look at this as well: https://setuptools.readthedocs.io/en/latest/setuptools.html#automatic-script-creation

setup.py
felixdivo: Apparently not. ^^

setup.py
hardbyte: I'm just surprised that no one complained about 3.0 through 3.3 not working! I'd have thought we'd get some issues if we remove support for `Python 3.4`... but then again people running Python3.4 probably aren't going to be running the latest release of python-can either.

I'm okay with you adding the `!=3.4`

setup.py
```diff
@@ -91,21 +90,19 @@
...
-        "": ["CONTRIBUTORS.txt", "LICENSE.txt", "CHANGELOG.txt"],
-        "doc": ["*.*"]
+        "": ["README.rst", "CONTRIBUTORS.txt", "LICENSE.txt", "CHANGELOG.txt"],
+        "doc": ["*.*"],
+        "examples": ["*.py"]
     },
 
     # Installation
     # see https://www.python.org/dev/peps/pep-0345/#version-specifiers
-    python_requires=">=2.7,!=3.0,!=3.1,!=3.2,!=3.3,!=3.4",
```
felixdivo: @hardbyte This commit now removed it ... and is merged into develop.

PR #533, opened 2019-03-13
Turn BusState into an enum

### LINKED ISSUE #519: Shouldn't BusState be an enum?
Currently, `BusState` is defined as

```python
BusState = namedtuple('BusState', 'ACTIVE, PASSIVE, ERROR')
```

But shouldn't it be an enum?

felixdivo:
Requires the `aenum` library on Python < 3.6. The `enum34` library was not used since it does not provide auto-assigned values.

Contains some small fixes/improvements to related code parts.

Closes #519.

@hardbyte you may add your other changes to this PR as well

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -11,12 +11,19 @@ import logging
 import threading
 from time import time
 from collections import namedtuple
+from aenum import Enum, auto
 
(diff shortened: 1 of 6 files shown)


felixdivo:
@hardbyte this now works

PR #534, opened 2019-03-14
Fix non-functional USB2CAN interface backend

acolomb:
During the refactorization and cleanup in #511, something slipped the review process.  I only got to testing on hardware with an actual CAN bus connected today and nothing worked.

Apparently the returned handle from `CanalOpen()` needs to be saved in a backend-local `.handle` member for all other API functions to use.  With all the confusion about CANAL API return values, the return statement for this got removed by accident.

### DIFF
--- a/can/interfaces/usb2can/usb2canabstractionlayer.py
+++ b/can/interfaces/usb2can/usb2canabstractionlayer.py
@@ -87,7 +87,7 @@ class Usb2CanAbstractionLayer:
         :param int flags: the flags to be set
 
         :raises can.CanError: if any error occurred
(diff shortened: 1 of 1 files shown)


felixdivo:
Ha. That was rather dumb. ^^ Thanks for testing it!

acolomb:
@felixdivo Would you mind merging this rather quickly?  I'd like to update my submodule reference today if possible ;-)

felixdivo:
Yep, I'm just waiting for the AppVeyor build to pass.

acolomb:
Thanks!

PR #536, opened 2019-03-20
Updated instances of **config to **kwargs, closes #492

### LINKED ISSUE #492: **config vs **kwargs
When naming the keyword arguments variable, the standard convention in python is to call it `kwargs`. Naming it `config` tends to throw off some introspection tools. Would you consider a PR to rename them to `kwargs` across the board?

Example: https://github.com/hardbyte/python-can/blob/develop/can/bus.py#L36

PeteAndersen:
`**config` -> `**kwargs`. Thanks for your patience!

Closes #492

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -33,7 +33,7 @@ class BusABC(object):
     RECV_LOGGING_LEVEL = 9
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


felixdivo:
Thanks for the cleanup!

PR #537, opened 2019-03-24
implemented FD for Pcan

bmeisels:
I have implemented FD support for PCAN (solving #507).
This code has been tested manually on PCAN-FD hardware.

Points to be discussed:

1. Parameters for the bit rate are passed as keyword args.
Should this be switched to be a single string? (this is what the Pcan API accepts).
The benefit to switching would be a simpler interface.
2. Messages with a non standard message size are padded right now with zeros.
Should we allow for an option to configure the padding?
What should the default padding be?

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -13,6 +13,7 @@ import time
 import can
 from can import CanError, Message, BusABC
 from can.bus import BusState
+from can.util import len2dlc, dlc2len
 from .basic import *
 
 boottimeEpoch = 0
@@ -65,6 +66,9 @@ pcan_bitrate_objs = {1000000 : PCAN_BAUD_1M,
                         5000 : PCAN_BAUD_5K}
 
 
+pcan_fd_parameter_list = ['nom_brp', 'nom_tseg1', 'nom_tseg2', 'nom_sjw', 'data_brp', 'data_tseg1', 'data_tseg2', 'data_sjw']
+
+
 class PcanBus(BusABC):
 
     def __init__(self, channel='PCAN_USBBUS1', state=BusState.ACTIVE, bitrate=500000, *args, **kwargs):
@@ -85,9 +89,80 @@ class PcanBus(BusABC):
         :param int bitrate:
             Bitrate of channel in bit/s.
             Default is 500 kbit/s.
+            Ignored if using CanFD.
+
+        :param bool fd:
+            Should the Bus be initialized in CAN-FD mode.
+
+        :param int f_clock:
+            Clock rate in Hz.
+            Any of the following:
+            20000000, 24000000, 30000000, 40000000, 60000000, 80000000.
+            Ignored if not using CAN-FD.
+            Pass either f_clock or f_clock_mhz.
+
+        :param int f_clock_mhz:
+            Clock rate in MHz.
+            Any of the following:
+            20, 24, 30, 40, 60, 80.
+            Ignored if not using CAN-FD.
+            Pass either f_clock or f_clock_mhz.
+
+        :param int nom_brp:
+            Clock prescaler for nominal time quantum.
+            In the range (1..1024)
+            Ignored if not using CAN-FD.
+
+        :param int nom_tseg1:
+            Time segment 1 for nominal bit rate,
+            that is, the number of quanta from (but not including)
+            the Sync Segment to the sampling point.
+            In the range (1..256).
+            Ignored if not using CAN-FD.
+
+        :param int nom_tseg2:
+            Time segment 2 for nominal bit rate,
+            that is, the number of quanta from the sampling
+            point to the end of the bit.
+            In the range (1..128).
+            Ignored if not using CAN-FD.
+
+        :param int nom_sjw:
+            Synchronization Jump Width for nominal bit rate.
+            Decides the maximum number of time quanta
+            that the controller can resynchronize every bit.
+            In the range (1..128).
+            Ignored if not using CAN-FD.
+
+        :param int data_brp:
+            Clock prescaler for fast data time quantum.
+            In the range (1..1024)
+            Ignored if not using CAN-FD.
+
+        :param int data_tseg1:
+            Time segment 1 for fast data bit rate,
+            that is, the number of quanta from (but not including)
+            the Sync Segment to the sampling point.
+            In the range (1..32).
+            Ignored if not using CAN-FD.
+
+        :param int data_tseg2:
+            Time segment 2 for fast data bit rate,
+            that is, the number of quanta from the sampling
+            point to the end of the bit.
+            In the range (1..16).
+            Ignored if not using CAN-FD.
+
+        :param int data_sjw:
+            Synchronization Jump Width for fast data bit rate.
+            Decides the maximum number of time quanta
+            that the controller can resynchronize every bit.
+            In the range (1..16).
+            Ignored if not using CAN-FD.
 
         """
         self.channel_info = channel
+        self.fd = kwargs.get('fd', False)
         pcan_bitrate = pcan_bitrate_objs.get(bitrate, PCAN_BAUD_500K)
 
         hwtype = PCAN_TYPE_ISA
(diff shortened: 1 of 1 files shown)


can/interfaces/pcan/pcan.py
```diff
@@ -210,50 +234,93 @@ def _recv_internal(self, timeout):
...
+        is_error_frame = (theMsg.MSGTYPE & PCAN_MESSAGE_ERRFRAME.value) == PCAN_MESSAGE_ERRFRAME.value
+
+
+        if self.fd:
+            dlc = pcan_dlc_list[theMsg.DLC]
+        else:
+            dlc = theMsg.LEN
 
-        dlc = theMsg.LEN
         timestamp = boottimeEpoch + ((itsTimeStamp.micros + 1000 * itsTimeStamp.millis + 0x100000000 * 1000 * itsTimeStamp.millis_overflow) / (1000.0 * 1000.0))
```
markuspi: According to PCAN Helpfile on `TPCANTimestampFD`:
> Represents the timestamp of a CAN message with flexible data rate. The time-stamp contains the number of microseconds since the start of Windows.
> *Python*
> `TPCANTimestampFD = c_ulonglong`

```suggestion
        if self.fd:
            timestamp = boottimeEpoch + (itsTimeStamp.value / (1000.0 * 1000.0))
        else:
            timestamp = boottimeEpoch + ((itsTimeStamp.micros + 1000 * itsTimeStamp.millis + 0x100000000 * 1000 * itsTimeStamp.millis_overflow) / (1000.0 * 1000.0))
```

bmeisels:
@markuspi good catch, thanks.
I fixed according to your suggestion and also fixed another mistake which would make the code not work on python3.

can/interfaces/pcan/pcan.py
```diff
@@ -102,7 +108,22 @@ def __init__(self, channel='PCAN_USBBUS1', state=BusState.ACTIVE, bitrate=500000
...
+        if self.fd:
+            f_clock_val = kwargs.get('f_clock', None)
+            if f_clock_val is None:
+                f_clock = "{}={}".format('f_clock_mhz', kwargs.get('f_clock_mhz', None))
+            else:
+                f_clock = "{}={}".format('f_clock', kwargs.get('f_clock', None))
+            
+            fd_parameters_values = [f_clock] + ["{}={}".format(key, kwargs.get(key, None)) for key in pcan_fd_parameter_list if kwargs.get(key, None) is not None]
+
+            self.fd_bitrate = bytes(' ,'.join(fd_parameters_values), 'ascii')
```
markuspi: Sending and receiving works fine now. Just a small adjustment is needed for python2 compatibility:
> `self.fd_bitrate = bytes(' ,'.join(fd_parameters_values), 'ascii')`
> `TypeError: str() takes at most 1 argument (2 given)`
```suggestion
            self.fd_bitrate = ' ,'.join(fd_parameters_values).encode("ascii")
```

bmeisels:
@markuspi fixed. Thanks again.

can/interfaces/pcan/pcan.py
```diff
@@ -210,50 +234,94 @@ def _recv_internal(self, timeout):
...
+            msgType |= PCAN_MESSAGE_FD.value
+        if msg.bitrate_switch:
+            msgType |= PCAN_MESSAGE_BRS.value
+        if msg.error_state_indicator:
+            msgType |= PCAN_MESSAGE_ESI.value
+
+        if self.fd:
+            # create a TPCANMsg message structure
+            if platform.system() == 'Darwin':
+                CANMsg = TPCANMsgFDMac()
```
felixdivo: We could also use this change to do some small cleanup here: The usual convention for variables names is lower_case_with_underscores.

felixdivo:
> Messages with a non standard message size are padded right now with zeros.
> Should we allow for an option to configure the padding?
> What should the default padding be?

This might be unexpected:

> Errors should never pass silently.
> [...]
> In the face of ambiguity, refuse the temptation to guess.
- [The Zen of Python](https://www.python.org/dev/peps/pep-0020/)

Maybe just raise a `ValueError`?

can/interfaces/pcan/pcan.py
bmeisels: I thought about doing some cleanup in a separate PR.
Would you rather i do some cleanup in this PR?

bmeisels:
> > Messages with a non standard message size are padded right now with zeros.
> > Should we allow for an option to configure the padding?
> > What should the default padding be?
> 
> This might be unexpected:
> 
> > Errors should never pass silently.
> > [...]
> > In the face of ambiguity, refuse the temptation to guess.
> > -> [The Zen of Python](https://www.python.org/dev/peps/pep-0020/)
> 
> Maybe just raise a `ValueError`?

A ValueError would be incorrect. I have seen multiple cases where this is intended behavior.
An explicit argument with a default value could work but it will complicate the interface.

can/interfaces/pcan/pcan.py
felixdivo: Separating them is of course cleaner! I won't stop you from doing so. ;)

felixdivo:
> A ValueError would be incorrect. I have seen multiple cases where this is intended behavior.
> An explicit argument with a default value could work but it will complicate the interface.

@christiansandberg What do you think? I'd be okay with that silently padding, as it might indeed make some applications quite a bit easier.

Padding with zeros to the next bigger size sounds like something people would expect, right?.

christiansandberg:
Padding is completely normal. I guess it could make sense to allow specification of other value than 0 for CRC purposes.

bmeisels:
After speaking to some colleagues i have changed my mind and believe throwing a `ValueError` would be correct. My colleagues made a point that the padding should not be handled by the data link layer (PcanBus) and should be handled at a higher level in the communication stack (implemented by the user).

I also saw that there is existing code in can.util for handling conversions from dlc to len and back.

I will switch the code to use this utility code instead of my implementation. In addition i will implement another function for checking the length is equal to the dlc.

bmeisels:
After looking through some code on other interfaces i noticed that i can use the len2dlc and dlc2len functions instead of implementing them myself.

Also i think can FD message length should be validated the  either by  _check in Message(as of now only checks that the length is less then 64 for FD) or the send in the Interface.

felixdivo:
> should be validated either by _check in Message (as of now only checks that the length is less then 64 for FD) or the send in the Interface

You are right, the _check method should validate that as well!

Maybe all interfaces should pass `check=True` when reading messages? It should never cause a false alarm, if we implemented it correctly.

bmeisels:
> Maybe all interfaces should pass `check=True` when reading messages? It should never cause a false alarm, if we implemented it correctly. We could also call `_check()` when sending a message.
If so then maybe check should default to True?
> But then, why don't we always check it? Performance?
I don't think performance is an issue. I can think of reasons one would like to send a malformed message and see it rejected by the other devices.

felixdivo:
> If so then maybe check should default to True?

Why? We can simply call it and ignore the result if everything is fine.

bmeisels:
> > If so then maybe check should default to True?
> 
> Why? We can simply call it and ignore the result if everything is fine.

i meant that the check parameter should default to true.
[message.py:88](https://github.com/hardbyte/python-can/blob/develop/can/message.py#L88)

felixdivo:
Ah, yes. I think that's a good idea. But I think we have already discussed it at some point ... @christiansandberg @hardbyte What do you think?

christiansandberg:
It could be a performance issue on e.g. a RapsberryPi on a high traffic bus.

felixdivo:
We could make it opt-out in the RC/config parameters.

bmeisels:
As it seems that the use of check is a larger issue,
I think this PR be merged and we should consider changes to the use of check separably.
What do you all think?

felixdivo:
@christiansandberg Could you also check if this PR shall be merged?

hardbyte:
> As it seems that the use of check is a larger issue,
I think this PR be merged and we should consider changes to the use of check separably.
What do you all think?

I agree, I'm not familiar with can-FD though so I'll defer to @christiansandberg

christiansandberg:
Sure, just document the new arguments. Otherwise people have to study the source code to figure out what config is accepted.

christiansandberg:
Maybe we should also standardize the various timing parameters between interfaces, but that can be done in a separate issue.

bmeisels:
> Sure, just document the new arguments. Otherwise people have to study the source code to figure out what config is accepted.

I documented the new arguments.


> Maybe we should also standardize the various timing parameters between interfaces, but that can be done in a separate issue.

I think this should extended to a discussion of standardizing interface parameters in general. For example the handling of the state parameter is different between interfaces.

felixdivo:
So there are two issues we should create out of this:
[(1)](https://github.com/hardbyte/python-can/pull/537#issuecomment-485246532) discussion of standardizing interface parameters in general, and
[(2)](https://github.com/hardbyte/python-can/pull/537#issuecomment-481327015) discuss whether the check parameter should default to true.

PR #540, opened 2019-04-01
Fixed simplecyclic test

bhass1:
Originally, the testcase would fail because timestamps would not match.
This was due to a Can.Message being initialized had a timestamp
value of 0, while a recv'd message had a virtual timestamp. When compared
they never matched.

Modified testcase now allows for some variation (0.016 s) in the timestamp
for a 0.01 s periodic message and compares the last message with the next
last message.

### DIFF
--- a/CONTRIBUTORS.txt
+++ b/CONTRIBUTORS.txt
@@ -26,3 +26,4 @@ Shaoyu Meng <shaoyu@life-foundry.com>
 Alexander Mueller<XelaRellum@web.de>
 Jan Goeteyn
 "ykzheng" <wishdo@gmail.com>
+Lear Corporation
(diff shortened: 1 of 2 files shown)


felixdivo:
But this does not check whether the message that should be sent was actually sent. Could we add that back again, while ignoring the timestamp on that comparison only?

bhass1:
Sure, how about by zeroizing the timestamp on the recv'd message before comparison? I didn't see an easier way to do it because of how the "allowed_timestamp_delta" works.

felixdivo:
Hm yeah, that should do. I didn't implement it more flexible than that. ;)

felixdivo:
But you can leave the current test in as well.

bhass1:
But you _did_ implement it which is saying more than most! Will push shortly. I'm having issues with running the tests off the develop branch in Windows which is why I was contributing from 3.1.1. I should probably open an issue ticket regarding it...

felixdivo:
If this is solving the only problem there was, you do not need to create a ticket for it.

bhass1:
This PR solves a different problem that was present on release-3.1.1 and hasn't been fixed yet. Once this PR is merged, release-3.1.1 passes all tests on my machine.

I created a new issue (#544) addressing the testing issues present in the develop branch.

PR #541, opened 2019-04-01
Fixes BusState.PASSIVE from config file for PCAN

bhass1:
Before, the config value "BusState" was not being actually sent to the PEAK hardware
due to the state setter not being invoked. Instead, the private _state variable
was being written to directly.

Now, the setter is being invoked and the state is being properly sent
to the PEAK hardware.


bhass1:
Happy to help.

Issue #506, opened 2019-02-15
add MF4 io writer


danielhrisca:
I would like to implement support for exporting to asam MF4 files (https://www.asam.net/standards/detail/mdf/). That would require my library as dependency ([asammdf](https://github.com/danielhrisca/asammdf)). Is this acceptable?

christiansandberg:
Wouldn’t that also require decoding raw messages to actual signal values?

danielhrisca:
That is not mandatory. 

The raw messages can be dumped and there are two recording modes that I would like to implement

1. unsorted mode : all messages are placed in the same channel group 
2. sorted mode: each message ID in dedicated channel groups

A database can be embedded in the MF4 file (dbc or arxml) and it can be used to later extract signal values. It is also possible to have anonymous logging if no database is attached.

christiansandberg:
I think it sounds like a good idea. But maybe add the library as an [extra](https://setuptools.readthedocs.io/en/latest/setuptools.html#id16) dependency.

felixdivo:
This sounds like a good idea! As not too many will be using this (i guess) we should add this as an extra. It would probably be a good time to add the allowed extras to a section called *Installation* or something like that in the Readme.

felixdivo:
@danielhrisca Do you need any help with this? I could also write a small tutorial on how to add new `Reader`s/`Writer`s. Would that help?

danielhrisca:
It would help, thank you. Unfortunately I did not have time to tackle this get.

felixdivo:
I wrote some text on that. See: https://github.com/hardbyte/python-can/pull/548/files#diff-200667e0b2eaf351c0585ba21fb7a144R70. Hope it helps. Please ask if questions remain,

danielhrisca:
I have a question regarding the message timestamp: does it always use the local time?

felixdivo:
I think usually [`time.time()`](https://docs.python.org/3/library/time.html#time.time) is used.

danielhrisca:
@felixdivo can you have a look at #554 ?

felixdivo:
Sure

PR #548, opened 2019-04-11
Add guide for new io formats

felixdivo:
This adds a guide on how to add new file format support.

Added in response to #506.

### DIFF
--- a/doc/internal-api.rst
+++ b/doc/internal-api.rst
@@ -73,7 +73,7 @@ About the IO module
 Handling of the different file formats is implemented in :mod:`can.io`.
 Each file/IO type is within a separate module and ideally implements both a *Reader* and a *Writer*.
 The reader usually extends :class:`can.io.generic.BaseIOHandler`, while
-the writer additionally extends :class:`can.Listener`,
+the writer often additionally extends :class:`can.Listener`,
 to be able to be passed directly to a :class:`can.Notifier`.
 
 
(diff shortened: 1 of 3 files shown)


danielhrisca:
@felixdivo 

each Writer is meant to handle a single bus channel? For example if we have a vector HW and we use bus channels 0 and 1 then the used should instantiate two Write objects for each of them?

felixdivo:
Writers are just dumb `can.Listeners`, which receive messages in the `on_message_received()` method. The writer does not care about the bus that this message came from or how it was brought there, except maybe that the `Message.channel` attribute is set that then gets serialized. So there may be zero to many  writers serializing the messages of zero to many buses. Distributing is often done by using a [Notifier](https://python-can.readthedocs.io/en/stable/api.html#notifier). Does that answer you question?

danielhrisca:
I got the point. I'll think about a strategy to handle this

doc/internal-api.rst
```diff
@@ -66,19 +67,52 @@ methods:
...
+
+
+Adding support for new file formats
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This assumes that you want to add a new file format, called *canstore*.
+
+1. Create a new module: *can/io/canstore.py*
+   (*or* simply copy some existing one like *can/io/csv.py*)
+2. Implement a reader ``CanstoreReader`` (which extends :class:`can.io.generic.BaseIOHandler`).
```
hardbyte: A `Reader` doesn't *have* to extend `BaseIOHandler`, that is a good starting point for any *file* based format though.

A small suggestion is to add a `~` to the path to improve readability in the rendered docs (but ensuring the link works).

```suggestion
2. Implement a reader ``CanstoreReader`` (which extends :class:`~can.io.generic.BaseIOHandler`).
```

doc/internal-api.rst
```diff
@@ -66,19 +67,52 @@ methods:
...
+The reader usually extends :class:`can.io.generic.BaseIOHandler`, while
+the writer additionally extends :class:`can.Listener`,
+to be able to be passed directly to a :class:`can.Notifier`.
+
+
+
+Adding support for new file formats
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This assumes that you want to add a new file format, called *canstore*.
```
hardbyte:
```suggestion
This assumes that you want to add a new file format, called *canstore*. Ideally add both reading and writing support for the new file format, although this is not strictly required.
```

doc/internal-api.rst
```diff
@@ -66,19 +67,52 @@ methods:
...
+
+1. Create a new module: *can/io/canstore.py*
+   (*or* simply copy some existing one like *can/io/csv.py*)
+2. Implement a reader ``CanstoreReader`` (which extends :class:`can.io.generic.BaseIOHandler`).
+   Besides from a constructor, only ``__iter__(self)`` needs to be implemented.
+3. Implement a writer ``CanstoreWriter`` (which extends :class:`can.io.generic.BaseIOHandler` and :class:`can.Listener`).
```
hardbyte: Again they are optional base classes. The only requirement is the writer must have a `on_message_received` method.

doc/internal-api.rst
felixdivo: I thought that it might be nice to know where this class sits, since this is documentation for the library developers.

PR #553, opened 2019-04-18
Add availability to set can custom speed in slcan backend

Lomonosow:
According the documentation, there are availability to set custom can speed instead use predefined speed values. I have add this future to slcan backend.

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -50,7 +50,7 @@ class slcanBus(BusABC):
     LINE_TERMINATOR = b'\r'
 
     def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
-                 sleep_after_open=_SLEEP_AFTER_SERIAL_OPEN,
+                 btr=None, sleep_after_open=_SLEEP_AFTER_SERIAL_OPEN,
                  rtscts=False, **kwargs):
         """
         :param str channel:
@@ -60,6 +60,8 @@ class slcanBus(BusABC):
             baudrate of underlying serial or usb device
         :param int bitrate:
             Bitrate in bit/s
+        :param str btr:
+            BTR register value to set custom can speed (overrides bitrate)
         :param float poll_interval:
             Poll interval in seconds when reading messages
         :param float sleep_after_open:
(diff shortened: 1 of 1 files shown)


Lomonosow:
Not exactly. Bitrate sets preset can bus speed, but using BTR you can set custom speed, like 33.3 kbit/s.

can/interfaces/slcan.py
```diff
@@ -88,6 +90,10 @@ def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
             else:
                 raise ValueError("Invalid bitrate, choose one of " + (', '.join(self._BITRATES)) + '.')
 
+        if btr is not None:
```
hardbyte: I assume *bad things happen* if the caller has provided a `bitrate` and a `btr`? If so I suggest we catch that and politely tell the caller of their mistake.

can/interfaces/slcan.py
```diff
@@ -60,6 +60,8 @@ def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
             baudrate of underlying serial or usb device
         :param int bitrate:
             Bitrate in bit/s
+        :param str btr:
+            BTR register value to set custom can speed
```
hardbyte: Add a note about the relationship to `bitrate` (does it override?)

can/interfaces/slcan.py
Lomonosow: As i know, if user set both options, btr will be applied, because it sets later than bitrate. If you want i can add catch if both options set.

felixdivo: Yeah, I think we should throw an exception in this case. (`ValueError`?)

Lomonosow: Added raise if both options "not None"

can/interfaces/slcan.py
Lomonosow: Done

can/interfaces/slcan.py
```diff
@@ -50,7 +50,7 @@ class slcanBus(BusABC):
     LINE_TERMINATOR = b'\r'
 
     def __init__(self, channel, ttyBaudrate=115200, bitrate=None,
-                 sleep_after_open=_SLEEP_AFTER_SERIAL_OPEN,
+                 btr=None, sleep_after_open=_SLEEP_AFTER_SERIAL_OPEN,
                  rtscts=False, **kwargs):
         """
```
felixdivo: > BTR register value to set custom can speed (overrides bitrate)

This is not true any more

```suggestion
         """
         :raise ValueError: if both *bitrate* and *btr* are set

```

Lomonosow: Thanks for your suggestion.

felixdivo: Ahh, the

> (overrides bitrate)

is still there ^^

Lomonosow: Removed

felixdivo:
@hardbyte This is now done.

PR #554, opened 2019-04-18
Add support for the mf4 io format

### LINKED ISSUE #506: add MF4 io writer

danielhrisca:
implements #506

(Closes #506)


danielhrisca:
This is a work in progress for which I would like to get some feed-back.

The mf4 io requires Python >= 3.6.0

can/io/__init__.py
```diff
@@ -18,3 +19,5 @@
 from .csv import CSVWriter, CSVReader
 from .sqlite import SqliteReader, SqliteWriter
 from .printer import Printer
+if sys.hexversion >= 0x03060000:
```
felixdivo: Same as above.

```suggestion
try:
    from .io import MF4Writer
except ImportError:
    pass
```

can/io/mf4.py
```diff
@@ -0,0 +1,160 @@
...
+
+from datetime import datetime
+from pathlib import Path
+import logging
+
+from ..listener import Listener
+from ..util import channel2int
+from .generic import BaseIOHandler
+
+from asammdf import MDF, Signal
```
felixdivo: We could surround this with a try/except block that sets `ASAMMDF_AVAILABLE = True` and wrap the `class MF4Writer ...` in an `if ASAMMDF_AVAILABLE:`-block. This unties all of this from version-checking and uses feature-checking instead.

can/io/mf4.py
```diff
@@ -0,0 +1,160 @@
...
+
+class MF4Writer(BaseIOHandler, Listener):
+    """Logs CAN data to an ASAM Measurement Data File v4 (.mf4).
+
+    If a message has a timestamp smaller than the previous one or None,
+    it gets assigned the timestamp that was written for the last message.
+    It the first message does not have a timestamp, it is set to zero.
+    """
+
+    def __init__(self, file, database=''):
```
felixdivo: I would set `database=None` as a default, though your approach works.

```suggestion
    def __init__(self, file, database=None):
```

Nice docs, by the way!

can/io/mf4.py
```diff
@@ -0,0 +1,160 @@
...
+            Signal(
+                name="CAN_ErrorFrame",
+                samples=np.array([], dtype=DTYPE),
+                timestamps=np.array([], dtype='<f8'),
+                attachment=attachment,
+            )
+        ]
+        self._mdf.append(sigs)
+
+        self._buffer = np.zeros(
```
felixdivo: This can go in a single line.

can/io/mf4.py
```diff
@@ -0,0 +1,160 @@
...
+            Signal(
+                name="CAN_DataFrame",
+                samples=np.array([], dtype=DTYPE),
+                timestamps=np.array([], dtype='<f8'),
+                attachment=attachment,
+            )
+        ]
+        self._mdf.append(sigs)
+
+        sigs = [
```
felixdivo: I think you accidentally duplicated something ;)

can/io/mf4.py
```diff
@@ -0,0 +1,160 @@
+# coding: utf-8
+
+"""
+Contains handling of MF4 logging files.
```
felixdivo: Do you plan on adding a reader as well? If not, we should mention that only writing is supported as of now.

setup.py
```diff
@@ -25,7 +25,8 @@
 # Dependencies
 extras_require = {
     'serial':   ['pyserial~=3.0'],
-    'neovi':    ['python-ics>=2.12']
+    'neovi':    ['python-ics>=2.12'],
+    'mf4':      ['asammdf>=5.5.0']
```
felixdivo: We also need to add `numpy` here.

```suggestion
    'mf4':      ['asammdf>=5.5.0', 'numpy>=WHATEVER_VERSION']
```

can/io/mf4.py
danielhrisca: Yes I do. The tricky part is that the mdf format makes a distinction between the normal and error frames so they are stored separately. I'll have to see how to interleave the two types for the reader part.

can/io/mf4.py
danielhrisca: This is necessary to correctly store normal and error frames

https://github.com/danielhrisca/python-can/blob/mf4_io/can/io/mf4.py#L86

https://github.com/danielhrisca/python-can/blob/mf4_io/can/io/mf4.py#L96

felixdivo: Ah, now I see it. I think it would be clearer if you wrote it like that:

```python
self._mdf.append([
    Signal(
        name="CAN_DataFrame",
        ...
    ),
    Signal(
        name="CAN_ErrorFrame",
        ...
    )
])
```

danielhrisca: I've made another simplification. The signals have to be appended separately to create individual channel groups for each of them

felixdivo: Looks good!

can/io/mf4.py
```diff
@@ -0,0 +1,427 @@
...
+    class MF4Reader(BaseIOHandler):
+        """
+        Iterator of CAN messages from a MF4 logging file.
+
+        """
+
+        def __init__(self, file):
+            """
+            :param file: a path-like object or as file-like object to read from
+                         If this is a file-like object, is has to opened in
```
felixdivo:
```suggestion
                         If this is a file-like object, is has to be opened in
```

test/logformats_test.py
```diff
@@ -393,6 +395,21 @@ def _setup_instance(self):
         )
 
 
+class TestMF4FileFormat(ReaderWriterTest):
```
felixdivo: You may wrap this in `try: from can import MF4Writer, MF4Writer [the actual "class TestMF4FileFormat ..."]; except ImportError: pass`

felixdivo:
You can also add this hot new feature to `README.rst#features`.

felixdivo:
But this looks like a great addition so far! :)

danielhrisca:
@felixdivo 

I've updated the listener code and now the tests pass:

```
test/logformats_test.py::TestCsvFileFormat::test_append_mode PASSED                                                                    [ 32%]
test/logformats_test.py::TestCsvFileFormat::test_file_like_context_manager PASSED                                                      [ 32%]
test/logformats_test.py::TestCsvFileFormat::test_file_like_explicit_stop PASSED                                                        [ 33%]
test/logformats_test.py::TestCsvFileFormat::test_path_like_context_manager PASSED                                                      [ 33%]
test/logformats_test.py::TestCsvFileFormat::test_path_like_explicit_stop PASSED                                                        [ 34%]
test/logformats_test.py::TestMF4FileFormat::test_append_mode SKIPPED                                                                   [ 35%]
test/logformats_test.py::TestMF4FileFormat::test_file_like_context_manager PASSED                                                      [ 35%]
test/logformats_test.py::TestMF4FileFormat::test_file_like_explicit_stop PASSED                                                        [ 36%]
test/logformats_test.py::TestMF4FileFormat::test_path_like_context_manager PASSED                                                      [ 36%]
test/logformats_test.py::TestMF4FileFormat::test_path_like_explicit_stop PASSED                                                        [ 37%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_append_mode PASSED                                                             [ 38%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_file_like_context_manager SKIPPED                                              [ 38%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_file_like_explicit_stop SKIPPED                                                [ 39%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_path_like_context_manager PASSED                                               [ 40%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_path_like_explicit_stop PASSED                                                 [ 40%]
test/logformats_test.py::TestSqliteDatabaseFormat::test_read_all PASSED                                                                [ 41%]

```

Once I release _asammdf 5.5.0_ everything will be in place for the MF4 io support

doc/listeners.rst
```diff
@@ -160,3 +160,29 @@ The following class can be used to read messages from BLF file:
...
+
+MF4 support requires Python >= 3.6
+
+The data is stored in a compressed format which makes it compact.
+
+.. note:: Channels will be converted to integers.
+
+.. note:: MF4Writer does not suppport append mode
+
+`MF4Writer` has the following init arguments
```
felixdivo: You could replace this section with

```suggestion
.. autoclass:: can.MF4Writer
    :members:
```

It adds the class documentation as well as the init parameters.

danielhrisca: Wouldn't that brake the builds for Python <3.6?

felixdivo: Ah, good point. But the docs are built on Python 3.6, so it would work. Though we would have do clearly document that it is only available on Python 3.6+. (Which you have already done!)

felixdivo:
The tests are not executed on [Travis CI](https://travis-ci.org/hardbyte/python-can/jobs/523221800#L624) or [AppVeyor](https://ci.appveyor.com/project/hardbyte/python-can/builds/24025901/job/9rmrtmrir5imc9c2#L130) yet. You need to:
- in `.travis.yml` change `travis_retry pip install .[test]` to `travis_retry pip install .[test,mf4]`, and
- in `.appveyor.yml` change `python -m pip install .[test,neovi]` to `python -m pip install .[test,neovi,mf4]`.

This should probably be added to the docs as a note, like:

> .. note:: MF4 support requires Python >= 3.6 and has to be installed as an extra with for example *pip install python-can[mf4]*.

danielhrisca:
@felixdivo 
I've released asammdf 5.5.0 

There seem to be some errors related to the canmatrix dependency on Linux. All the windows tests are passed no matter the python version used. Do you have any idea?

felixdivo:
Hm, your newly added tests [are not executed on windows](https://ci.appveyor.com/project/hardbyte/python-can/builds/24095513/job/979hhj2c3kn84kit#L152).

On Linux there seem to be two problems:
- the one with canmatrix (I have no idea what's going on)
- some problem with numexpr: `Failed building wheel for numexpr` on PyPy. Is it just not supported on that platform?

danielhrisca:
Is there a planned release date for python-can 4.0.0?

hardbyte:
I was thinking of getting python-can 3.2 out soon (in the next few weeks). 
Ideally that 3.2 release includes this Mf4 addition merged in, but the travis tests will have to pass

danielhrisca:
@altendky 
can you have a look at the canmatrix errors in failed travis jobs?

altendky:
https://travis-ci.org/hardbyte/python-can/jobs/525256576#L585
`TypeError: attrib() got an unexpected keyword argument 'factory'`

A newer version of attrs is required.  I believe canmatrix has up to date deps at least on the development branch.

The other stuff seems like a weird internal issue...  I don't understand though because the log seems to show canmatrix==0.7 being installed and it has the import right on the previous line.

https://github.com/ebroecker/canmatrix/blob/0.7/src/canmatrix/__init__.py#L6-L7
```python
import canmatrix._version
__version__ = canmatrix._version.get_versions()['version']
```

Running pytest locally on develop for python-can shows two other test failures, but not the canmatrix stuff from Travis.

So...

Is this a recurring issue?  It's also only on 3.7-dev.  Can't you run real 3.7 now?  Sorry I don't have a real answer for you yet.

felixdivo:
Another nice way to skip conditionally: https://docs.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency. But I did not check whether this fits here.

felixdivo:
We should also add a case in `can.io.player.LogReader`. I'll add that to the docs.

felixdivo:
> Can't you run real 3.7 now?

No, Travis does not support this on this version.

felixdivo:
The problem is that socketcan cannot be tested on TravisCI's Xenial environments. See [this branch](https://travis-ci.org/hardbyte/python-can/jobs/488566990#L175). We could run the other tests all on Xenial and only run the socketcan stuff once on the old platform.

altendky:
@felixdivo, well, I'll throw away my response then...  grr.  Thanks for the clarification.  Personally I'm shifting to Azure Pipelines.  As I do more there we'll see how it pans out but at least you can get all the Pythons on all the platforms in one place (and artifact storage, oh my).

felixdivo:
We could disable 3.7 testing for now, since I don't think it that it provides any huge benefits as it is now. We can sort out the current problems with CI testing in general at some later point. @hardbyte What do you think?

felixdivo:
I'm trying to fix the problem with Python 3.7-dev in #563.

Could you @danielhrisca have a look on why PyPy fails?

danielhrisca:
@felixdivo 
looks like a pandas compilation error. However since the MF4 requires Python >= 3.6 I think the current PYPY test should be skipped (PYPY2.7 and PYPY3.5 are used in the travis builds)

altendky:
Taking another look at the attrs/canmatrix stuff.

https://travis-ci.org/hardbyte/python-can/jobs/525256576#L484
```
Collecting canmatrix (from asammdf>=5.5.0->python-can==3.2.0a0)
  Downloading https://files.pythonhosted.org/packages/a5/a1/b2bd32bb0e0898e8e80f5e3ef5f42c54c3dea95d9df05929d8b2a2e71b17/canmatrix-0.7-py2.py3-none-any.whl (106kB)
```
Which requires attrs >= 18.1.0

https://github.com/ebroecker/canmatrix/blob/0.7/setup.py#L70
```python
    install_requires = ["future", "attrs>=18.1.0", "bitstruct", "pathlib2"],
```

but!

https://travis-ci.org/hardbyte/python-can/jobs/525256576#L497
```
Requirement already satisfied: attrs>=17.4.0 in /home/travis/virtualenv/python3.7-dev/lib/python3.7/site-packages (from pytest~=4.3->python-can==3.2.0a0)
```
... I dunno

I'll pull a branch off this and add some testing to explore.

felixdivo:
#563 is now finished (But still needs approval) and uses a real Python 3.7 environment. It also includes Python 3.8 dev & nightly.

altendky:
Order matters.  *sigh*  `pytest` is found first and it's requirement is `attrs>=17.4.0` and so `attrs>=18.1.0` from `canmatrix` is ignored.  Or something.  `pip` isn't really a dependency version resolver.  It makes some basic effort...  but yeah.  Here are a couple similar tests I did locally varying by installing `pytest canmatrix` vs `canmatrix pytest` (with old `attrs` already in place).

attrs still old: https://gist.github.com/altendky/d7c19e34a2c338851c5453067c69377a
attrs was updated: https://gist.github.com/altendky/21f093268277eace8f0523e53b65282d

(options are in no particular order)

So option 1 is `--upgrade`.

https://gist.github.com/altendky/07ab0470dac1c7b77c1c0c0159ac1f17

Let's say that option 2 would be creating a virtualenv (or venv in the py3 cases).  It'll be empty so you won't have this issue.  `pip` does 'ok' when you install everything at one time.  I'll lump in 'start using tox' with this since it would be creating an env for you.  Bonus, people could easily test multiple versions locally.

I'm thinking option 3 might be 'really take control' and use some other tool for picking the actual versions.  I use `pip-tools` but a quick look doesn't show how I could specify the `[test,mf4]` extras to it.  `:[`

There are probably more but I'm guessing option 1 will be chosen anyways, at least for now.

felixdivo:
I rebased this, so we can now see what is happening on Python 3.7.

felixdivo:
> So option 1 is `--upgrade`.
> 
> https://gist.github.com/altendky/07ab0470dac1c7b77c1c0c0159ac1f17
> 
> Let's say that option 2 would be creating a virtualenv (or venv in the py3 cases). It'll be empty so you won't have this issue. `pip` does 'ok' when you install everything at one time. I'll lump in 'start using tox' with this since it would be creating an env for you. Bonus, people could easily test multiple versions locally.
> 
> I'm thinking option 3 might be 'really take control' and use some other tool for picking the actual versions. I use `pip-tools` but a quick look doesn't show how I could specify the `[test,mf4]` extras to it. `:[`
> 
> There are probably more but I'm guessing option 1 will be chosen anyways, at least for now.

@hardbyte Would you be okay with Option 1 for now and opening a new issue to apply Option 2?

felixdivo:
PyPy now fails like [this](https://travis-ci.org/hardbyte/python-can/jobs/527672314#L5944):

> Exception: Cython-generated file 'pandas/_libs/algos.c' not found.
                    Cython is required to compile pandas from a development branch.
                    Please install Cython or download a release package of pandas.

But yeah, the tests should be skipped because of the Python version being to low.

felixdivo:
> But yeah, the tests should be skipped because of the Python version being to low.

This could be done like with [this dependency](https://github.com/hardbyte/python-can/blob/develop/setup.py#L104). The library would not be installed on Python (CPython/PyPy/...) version < 3.6 and as such would not be tested either.

setup.py
```diff
@@ -25,7 +25,8 @@
 # Dependencies
 extras_require = {
     'serial':   ['pyserial~=3.0'],
-    'neovi':    ['python-ics>=2.12']
+    'neovi':    ['python-ics>=2.12'],
+    'mf4':      ['asammdf>=5.5.0', 'numpy>=1.16.0']
```
felixdivo: This would be the coolest, but I don't know whether this actually works:

```suggestion
    'mf4;python_version>="3.6"':      ['asammdf>=5.5.0', 'numpy>=1.16.0']
```

This should work anyways:

```suggestion
    'mf4':      ['asammdf>=5.5.0;python_version>="3.6"', 'numpy>=1.16.0;python_version>="3.6"']
```

felixdivo:
image

What shall we do about this? Allowing all XCode dev builds to fail would be sad ...

hardbyte:
> > So option 1 is `--upgrade`.
> > https://gist.github.com/altendky/07ab0470dac1c7b77c1c0c0159ac1f17
> > Let's say that option 2 would be creating a virtualenv (or venv in the py3 cases). It'll be empty so you won't have this issue. `pip` does 'ok' when you install everything at one time. I'll lump in 'start using tox' with this since it would be creating an env for you. Bonus, people could easily test multiple versions locally.
> > I'm thinking option 3 might be 'really take control' and use some other tool for picking the actual versions. I use `pip-tools` but a quick look doesn't show how I could specify the `[test,mf4]` extras to it. `:[`
> > There are probably more but I'm guessing option 1 will be chosen anyways, at least for now.
> 
> @hardbyte Would you be okay with Option 1 for now and opening a new issue to apply Option 2?

Sorry I haven't chimed in yet. Yes I'd prefer option 2, but am happy with installing with `--upgrade` for now.

Regarding tox... happy to add the config file I've been using - https://github.com/hardbyte/python-can/blob/add-tox/tox.ini

doc/listeners.rst
```diff
@@ -160,3 +160,27 @@ The following class can be used to read messages from BLF file:
...
+
+.. note:: Channels will be converted to integers.
+
+.. note:: MF4Writer does not suppport append mode.
+
+
+.. autoclass:: can.MF4Writer
+    :members:
+
+MF4Reader can only replay files created with MF4Writer. The following class can be used to read messages from MF4 file:
```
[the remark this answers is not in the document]
danielhrisca: Hello @hardbyte ,

I've made changes according to your review. Let me know if the explanation is clear enough

felixdivo:
See #582.

setup.py
```diff
@@ -25,7 +25,8 @@
 # Dependencies
 extras_require = {
     'serial':   ['pyserial~=3.0'],
-    'neovi':    ['python-ics>=2.12']
+    'neovi':    ['python-ics>=2.12'],
+    'mf4':      ['asammdf>=5.5.0;python_version>="3.6"', 'numpy>=1.16.0;python_version>="3.6"']
```
felixdivo: By now, the `;python_version>="3.6"` can be removed, since we do not support Python < 3.6 anyways.

felixdivo:
I think this can be merged once the CI works (this could now be the case), and the three above comments are resolved.

can/io/__init__.py
```diff
@@ -6,6 +6,7 @@
 """
 
 from __future__ import absolute_import
+import sys
```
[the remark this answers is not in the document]
danielhrisca: It was no needed so I've removed it.

setup.py
danielhrisca: @felixdivo  
done

felixdivo:
Do we want to simply not support the combination of Pypy + MF4?

felixdivo:
Related: https://github.com/travis-ci/travis-ci/issues/9929#issuecomment-500073011

.travis.yml
```diff
@@ -16,12 +16,9 @@ python:
   # PyPy:
   - pypy3
 
-env:
-
-
 install:
   - if [[ "$TEST_SOCKETCAN" ]]; then sudo bash test/open_vcan.sh ; fi
-  - travis_retry pip install .[test]
+  - travis_retry pip install .[test,mf4]
```
felixdivo: Maybe this is not the right way to configure the tests. Maybe we should simply not support the combination of MF4 & PyPy (as mentioned above). This could be done by only installing `travis_retry pip install .[test]` and instead changing `setup.py` to:

```python
import platform
is_cpython = 'CPython' == platform.python_implementation()

tests_require = [
    'pytest~=4.3',
    'pytest-timeout~=1.3',
    'pytest-cov~=2.6',
    'codecov~=2.0',
    'six',
    'hypothesis'
] + extras_require['serial']

if is_cpython:
    tests_require += extras_require['mf4']
```

A very simple not in the docs should be added as well:
`.. note:: *MF4* is currently only supported on CPython and not on PyPy.`

felixdivo:
I raised a question about this [over here](https://travis-ci.community/t/problem-with-osx-python/3671).

.travis.yml
altendky: https://www.python.org/dev/peps/pep-0508/#environment-markers

```python
'mf4; platform_python_implementation == "CPython"'
```
Maybe?  Untested, but definitely want to get away from arbitrary code relating to setup.  Also maybe `!= "PyPy"` depending on intent.

felixdivo:
I would propose dropping OSX + Python 3.7-dev from our Travis CI list. @hardbyte Are you okay with that?

> This is probably due to the python binary file. You are using 3.7-dev, which is most likely not the 3.7 release (of any version). We have not compiled the mac archives in months (if not years). At the moment, we have no plans to resume compiling them. It is best to remove these rogue archives, so as not to confuse users.

felixdivo:
#622 is now merged. Let's see if everything except the PyPy stuff will work.

.travis.yml
felixdivo: This does not seem to work on my machine: I get `python-can 3.2.0 does not provide the extra 'mf4'` on CPython 3.6. Removing the spaces did not help either.

felixdivo: @danielhrisca Sorry for the long delay. With the above change, this should be ready to finally be merged.

altendky: @felixdivo, I think maybe I misread the code there.  Sorry.  The point is to document dependencies in a way that can be retrieved regardless of what Python is being used to run setup.py.  For example, what I wrote (or something like it) can go into a wheel intact and be processed when the wheel is installed.  `if is_cpython:` would be processed when the wheel is built and then when the wheel is used it would be fixed according to the Python used to create the wheel.  This is also quite relevant to locking dependencies.

But...  I was thinking at a different layer.  I don't know that there's a tidy all around solution for this yet.  Applying the environment markers to each dependency of the mf4 extra could make it a no-op at least and then it could be unconditionally (code-wise) included wherever.  Maybe that's what I would do, I'm not quite sure.

It would be nice if the extra could be made conditionally available but my naive attempt didn't work.

https://gist.github.com/altendky/357e6a19691912593080f18ddcc39235

It would also be nice if setup.py could reference its own extras as dependencies like `.[mf4]` or such but last I played with that I didn't find anything that worked.

felixdivo: I got that idea, but it's largely unsupported for now. :( I was just trying what is implemented and what not. That shortcoming of `setup.py` is really annoying as well. -.-

felixdivo:
@danielhrisca Do you have time to finish this?

danielhrisca:
@felixdivo 
the pypy errors seems related to this https://travis-ci.community/t/xenial-pypy3-image-includes-broken-numpy/3523

felixdivo:
Apparently the bug is not being cared about or fixed. :frowning_face: I would suggest disabling the PyPy support for now and opening an issue to change it once the above bug is fixed.

felixdivo:
I merged this into a new branch to apply some changes, and will merge that one into develop in #697.

PR #558, opened 2019-04-27
Fix broken execution of tests on some Windows environments

### LINKED ISSUE #544: Failing "setup.py test" on develop branch after 3.1.1 in Windows
OS: Windows 7
Python: 3.6.1
Branch: develop
Commit: dd89de4974d.....

Description:
Windows doesn't like the changes between release-3.1.1 and referenced branch+commit above. 

Running "py -3 setup.py test" from the tip of branch "release-3.1.1" completes with an error on simplecyclic test which is fixed in PR #540 . The same exact command run on the above referenced branch+commit fails with the following. See image below.

image

I haven't tried to root-cause the issue yet. Appears one of the newly supported interfaces causes a hiccup.

felixdivo:
This fixes the non-standard usage of `__test__ = False` to exclude the base class from execution and uses a simple `del` statement instead. The `__test__` attribute was a Nose idiom (though supposedly supported by pytest 🤷‍♂️). But as we don't use Nose any more, this seems to be the better way!

Fixes #544. (@bhass1 could you verify this, since you can reproduce the error nicely?)

### DIFF
--- a/test/logformats_test.py
+++ b/test/logformats_test.py
@@ -43,9 +43,12 @@ class ReaderWriterTest(unittest.TestCase, ComparingMessagesTestCase):
     """Tests a pair of writer and reader by writing all data first and
     then reading all data and checking if they could be reconstructed
     correctly. Optionally writes some comments as well.
-    """
 
-    __test__ = False
+    .. note::
+        This class is prevented from being executed as a test
+        case itself by a *del* statement in at the end of the file.
+        (Source: `*Wojciech B.* on StackOverlfow <https://stackoverflow.com/a/22836015/3753684>`_)
+    """
 
     __metaclass__ = ABCMeta
 
@@ -318,8 +321,6 @@ class ReaderWriterTest(unittest.TestCase, ComparingMessagesTestCase):
 class TestAscFileFormat(ReaderWriterTest):
     """Tests can.ASCWriter and can.ASCReader"""
 
-    __test__ = True
-
     def _setup_instance(self):
         super(TestAscFileFormat, self)._setup_instance_helper(
             can.ASCWriter, can.ASCReader,
@@ -331,8 +332,6 @@ class TestAscFileFormat(ReaderWriterTest):
 
 class TestBlfFileFormat(ReaderWriterTest):
     """Tests can.BLFWriter and can.BLFReader"""
-
-    __test__ = True
 
     def _setup_instance(self):
         super(TestBlfFileFormat, self)._setup_instance_helper(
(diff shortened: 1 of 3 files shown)


felixdivo:
I also checked that the tests are actually executed. 😉

hardbyte:
Would it work to rename the base class to not include "test"?

felixdivo:
No, that does not suffice:

image

I think pytest looks for attributes of a module with `isinstance(the_thing, unittest.TestCase) == True`.

bhass1:
Testing now.. will check in shortly.

bhass1:
On commit: d68e5434c8e8aee13, with just the following change:
image
the when I execute `py -3 setup.py test` the testcases execute now:
image


One problem, the testcases are failing with errors on 4 of the testcases as shown above.

With the alias for pytest back in (so pytest is used), when I execute `py -3 setup.py test` I get the following output: 
image

There are some discrepancies...

felixdivo:
Okay, but that means that the `logformats_test.py` now works as intended. I applied the same patch to `test_scripts.py`. Also I found some smaller problems in other test files that raised warnings.

felixdivo:
I now think this might work, because the `CanScriptTest` is not executed as a standalone TestCase any more.

felixdivo:
@bhass1 Can you confirm that?

hardbyte:
Really appreciate you two getting to the bottom of this! 

I'm a bit uncomfortable with the `del` method (although I agree it is better than what we had!). @felixdivo  you're of course correct about inheriting from `unittest.TestCase`. Another approach is to use mixins which I think would be a cleaner approach here. 

I've used them with pytest test cases recently in another project:
https://github.com/data61/anonlink/blob/develop/tests/test_e2e.py#L29-L81

felixdivo:
I don't like the approach *that* much either, though I've made bad experiences with mixins in code that is used by many people. Maybe this would be another possibility: https://docs.pytest.org/en/latest/skipping.html#skip-all-test-functions-of-a-class-or-module

bhass1:
@felixdivo Looks fixed to me! 
Running `py -3 setup.py test` with the unittest module (comment out the alias in `setup.cfg` as before):
image

Running `py -3 setup.py test` with the pytest alias:
image

Not sure why pytest runs many fewer tests, but at least there aren't errors anymore.

felixdivo:
> Not sure why pytest runs many fewer tests

Strange, can you post the two console outputs so we can look why the number of tests is different?

@hardbyte I would merge this as is, but feel free to change the above bodge ;)

hardbyte:
I went to check this out myself on a Windows 10 laptop with Python 3.7 and I see a **lot** of failures. Looking at [AppVeyor](https://ci.appveyor.com/project/hardbyte/python-can/builds/24252632/job/ys2r5jy7kh6wocnm) we don't seem to be testing on Python 3:

```
pytest
============================= test session starts =============================
platform win32 -- Python 2.7.16, pytest-4.4.1, py-1.8.0, pluggy-0.9.0 -- c:\python27\python.exe
```

I'll open an issue to track fixing that.

PR #561, opened 2019-04-30
Test MessageSync

### LINKED ISSUE #369: Add test for `can.MessageSync`

felixdivo:
Closes #369.

This PR:
- add tests for `can.MessageSync` (a small change to `ComparingMessagesTestCase` was required)
- add some types to the docs
- remove unnecessary logging
- fixes a possible bug that could have occured when for example the second message has a lower timestamp then the first one (now capps to a minimum 0.0)
- disable GC while running the (probably for the CI server still too) unreliable `test_cycle_time()`

Will have to check whether these new tests run reliably on the CI server.

### DIFF
--- a/test/simplecyclic_test.py
+++ b/test/simplecyclic_test.py
@@ -30,6 +30,7 @@ class SimpleCyclicSendTaskTest(unittest.TestCase, ComparingMessagesTestCase):
         with can.interface.Bus(bustype='virtual') as bus1:
             with can.interface.Bus(bustype='virtual') as bus2:
 
+                # disabling the garbage collector makes the time readings more reliable
                 gc.disable()
 
                 task = bus1.send_periodic(msg, 0.01, 1)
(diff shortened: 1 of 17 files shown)


felixdivo:
This actually works, it's just AppVeyor being strange. -.-

test/simplecyclic_test.py
```diff
@@ -24,34 +25,36 @@ def __init__(self, *args, **kwargs):
...
-        self.assertMessageEqual(msg, last_msg)
-	
-        bus1.shutdown()
-        bus2.shutdown()
+        msg = can.Message(is_extended_id=False, arbitration_id=0x123, data=[0, 1, 2, 3, 4, 5, 6, 7])
 
+        with can.interface.Bus(bustype='virtual') as bus1:
+            with can.interface.Bus(bustype='virtual') as bus2:
+
+                gc.disable()
```
hardbyte: Please add a 1 line comment as to why this is required

test/test_message_sync.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_skip(self):
+        messages = copy(TEST_FEWER_MESSAGES)
+        sync = MessageSync(messages, skip=0.005, gap=0.0)
+
+        before = time()
+        collected = list(sync)
+        after = time()
+        took = after - before
+
+        # the handling of the messages itself also take time: ~0.001 s/msg on my laptop
```
hardbyte: nit: instead of *my* laptop say something like "a high end 2018 laptop".

felixdivo: Sure 👍

PR #563, opened 2019-05-01
Update Travis CI to use Xenial for most jobs

felixdivo:
- Moves Travis from Trusty to Xenial for most jobs (except OSX jobs & testing socketcan, see below)
- testing socketcan remains as a test on Python 3.6 only on Trusty
- Makes building the docs run in a separate job
- Fixes various smaller problems with skipping socketcan tests outside of Travis CI
- Fixes: `Warning: 'classifiers' should be a list, got type 'tuple'` on Python 3.8a ("nightly")

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,7 +70,7 @@ script:
       # Run the tests
       python setup.py test
       # Upload the coverage to codecov.io
-      if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then codecov fi
+      if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then codecov; fi
     fi
 
 # Have travis deploy tagged commits to PyPi
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ script:
       python -m sphinx -an doc build
     else
       # Run the tests
-      python setup.py test
+      python setup.py test && # (preserve the error code semantics)
       # Upload the coverage to codecov.io
       if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then codecov; fi
     fi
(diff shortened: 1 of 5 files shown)


felixdivo:
This is done. Do you agree with the changes? Python 3.8 (dev & nightly) fails for some reason but that's not a problem for today.

felixdivo:
TODO: [Apparently](https://docs.codecov.io/docs/merging-reports) the coverage should be uploaded on all builds an codecov.io merges the reports, which results in a high coverage being reported. I will change it back to how it was.

.travis.yml
```diff
@@ -25,39 +26,54 @@ os:
...
 sudo: required
 
 matrix:
-  # see "os: ..." above
   include:
+    # building the docs
+    - python: "3.7"
+      env: BUILD_ONLY_DOCS=TRUE
+    # testing socketcan on Trusty & Python 3.6, since it is not available on Xenial
+    - os: linux
```
hardbyte: It might just be this one that needs sudo?

.travis.yml
```diff
@@ -25,39 +26,54 @@ os:
...
+    if [[ "$BUILD_ONLY_DOCS" ]]; then
+      # Build the docs with Sphinx
+      #   -a Write all files
+      #   -n nitpicky
+      python -m sphinx -an doc build
+    else
+      # Run the tests
+      python setup.py test && # (preserve the error code semantics)
+      # Upload the coverage to codecov.io
+      if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then codecov; fi
```
hardbyte: As you commented, this condition needs to be removed - code cov should also be updated when the tests fail. Won't the `&&` short circuit?

.travis.yml
felixdivo: Ah yes, that could work.

.travis.yml
felixdivo: Jup, this now works.

PR #564, opened 2019-05-01
[Do _NOT_ Merge] Mf4 io canmatrix debug [WIP]

altendky:
Diagnostics for #554


felixdivo:
This can now be closed, right? Thanks for your investigation, again!

PR #567, opened 2019-05-03
adding seeedstudio usb-can-analyzer interface

shedfly:
Hi Guys,  

I've added an interface for the seeedstudio canbus interface following a blog post:
Posted by Wilfried Voss on September 19, 2018 - copperhilltech.com

I'm not sure if this is something you'd want this in your repo? I thought a pull request was the easiest way to ask.  Note it probably still needs a bit of work I'll update it as I go.  Please let me know if you're keen.


setup.py
```diff
@@ -36,7 +36,8 @@
     'codecov~=2.0',
     'future',
     'six',
-    'hypothesis'
+    'hypothesis',
+    'crccheck'
```
felixdivo: This needs to be put in `install_requires` (found bottom of the file), since `test_require` is only installed for the tests.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
```diff
@@ -0,0 +1,276 @@
...
+        byte_msg.append(0xAA)     # Frame Start Byte 1
+        byte_msg.append(0x55)     # Frame Start Byte 2
+
+        byte_msg.append(0x12)     # Initialization Message ID
+
+        byte_msg.append(CanAnalyzer.BITRATE[self.bit_rate])  # CAN Baud Rate
+        byte_msg.append(CanAnalyzer.FRAMETYPE[self.frame_type]) 
+
+        for i in range(0, 4):
+            byte_msg.append(self.filter_id[i])
```
christiansandberg: You can also do like this:

```python
byte_msg.extend(self.filter_id)
```

Same applies at other places.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
```diff
@@ -0,0 +1,276 @@
...
+
+        for i in range(0, 4):
+            byte_msg.append(0x00)
+
+        print "byte_str = " + str(byte_msg[2:])
+
+        crc = Crc8Darc.calc(byte_msg[2:])
+        crc_byte = struct.pack('B', crc)
+
+        print "crc_byte: " + str(crc_byte) + " length: " + str(len(crc_byte))
```
christiansandberg: Replace this with a `logging` call instead. Same for other places.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
```diff
@@ -0,0 +1,276 @@
...
+        byte_msg.append(0xAA)
+        for i in range(0, 4):
+            byte_msg.append(timestamp[i])
+        byte_msg.append(msg.dlc)
+        for i in range(0, 4):
+            byte_msg.append(a_id[i])
+        for i in range(0, msg.dlc):
+            byte_msg.append(msg.data[i])
+        byte_msg.append(0xBB)
+        self.ser.write(byte_msg)
```
christiansandberg: I don't see the format of the sent frame in the notes. It is a little strange that the transmitted frames use timestamps but the received do not.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
```diff
@@ -0,0 +1,276 @@
...
+                    arb_id = (struct.unpack('<H', s_3_4))[0]
+
+                data = bytearray(self.ser.read(length))
+                end_packet = ord(self.ser.read())
+
+                if end_packet == 0x55:
+                    msg = Message(timestamp=time_stamp,
+                                  arbitration_id=arb_id,
+                                  extended_id=is_extended,
+                                  data=data)
```
christiansandberg: Could also add `is_remote_frame=not is_data`.

setup.py
christiansandberg: Or better put it as an extra like the neovi interface.

shedfly: setup.py has been updated with crccheck as an option in extras_require.  The option is added with the name 'crccheck' so its not too specific to one interface; also because usb-can-analyzer is a bit wordy.  

In addition a try - except on importing the module was added to point users in the right direction.  It will likely still crash out without crccheck installed if using usb-can-analyser.  Not sure if there's a better way?

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
shedfly: I believe I've completed this.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
shedfly: done.

shedfly: I've added debug level logs of written data as binary, let me know if would be better to format it?  I hesitated to add another import.  maybe binascii.hexlify()?

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
shedfly: It could take a week or two for me to resolve this.  

I suspect you've picked up that I've started from serial_can.py, and that I not got to implementing the required send packet.  My current application is only required to read.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
shedfly: should be done, 
its set when remote: ie 1='is_remote', changed var name and fixed mask to be 0x10.

christiansandberg: Oh and also add the dlc argument.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
```diff
@@ -0,0 +1,273 @@
...
+                else:
+                    s_3_4 = bytearray(self.ser.read(2))
+                    arb_id = (struct.unpack('<H', s_3_4))[0]
+
+                data = bytearray(self.ser.read(length))
+                end_packet = ord(self.ser.read())
+                if end_packet == 0x55:
+                    msg = Message(timestamp=time_stamp,
+                                  arbitration_id=arb_id,
+                                  extended_id=is_extended,
```
christiansandberg: Should be written as `is_extended_id` now.

setup.py
```diff
@@ -25,7 +25,8 @@
 # Dependencies
 extras_require = {
     'serial':   ['pyserial~=3.0'],
-    'neovi':    ['python-ics>=2.12']
+    'neovi':    ['python-ics>=2.12'],
+    'usb-can-analyzer': ['crccheck>=0.6']
```
christiansandberg: Also requires pyserial.

shedfly: in the last update I intended to put:
'crccheck': ['crccheck>=0.6']
I'll fix that shortly.  Regrading pyserial,  I believe that is already done?  if somethings missing let me know.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
shedfly: Maybe not so long:
https://github.com/SeeedDocument/USB-CAN_Analyzer/blob/master/res/USB-CAN%20software%20and%20drive(v7.10)/Document/USB%20(Serial%20port)%20to%20CAN%20protocol%20defines.pdf

setup.py
```diff
@@ -25,7 +25,8 @@
 # Dependencies
 extras_require = {
     'serial':   ['pyserial~=3.0'],
-    'neovi':    ['python-ics>=2.12']
+    'neovi':    ['python-ics>=2.12'],
+    'crccheck': ['crccheck>=0.6']
```
christiansandberg: The name should only specify the functionality that the user wants, e.g. the interface you want to use. The packages needed for that to work is not the user’s concern. In the future you may change what crc library to use.

In this case the name should be something like you had before and both pyserial and crccheck are needed.

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
felixdivo: The DLC is being set in line 252, right?

can/interfaces/usb_can_analyzer/usb_can_analyzer.py
felixdivo: I think it's fine.

joaofl:
I've been looking forward to this integration. Would also be nice if an example on how to use this specific converter was added to the "examples" folder for getting started easily.

marcel-kanter:
Can you please clearify the name a bit more to avoid confusion with others?
usb_can_analyzer is a bit generic and there are many things that match that name.
Ixxat has a cananalyzer
vectors canalyzer

something like 
seedstudio_usbcan
seedstudio_can

shedfly:
How's seeed_can for a name?  (3 e's like in www.seeedstudio.io)
I'll work on the doco over the weekend coming.

marcel-kanter:
That sounds good.
Can you please check if they provide other can interfaces, or is this one the only one?

christiansandberg:
Or just seeedstudio. Most interfaces here use the manufacturer name since the API is usually the same for all their CAN products and the word CAN is a bit redundant in a CAN library.

felixdivo:
Can this be merged?

shedfly:
working on docs now, I'll update when done, might be tomorrow.

shedfly:
debugging something, I'll update again tomorrow.

shedfly:
The last commits adds doco but made a few changes as well.  
1) I changed the name of the main class from 'CanAnalyzer' to 'SeeedBus', it seemed to fit better with the interface name and align with the other interfaces.
2) the debugging I was doing determined that will it was sometimes working with the CRC library but were configurations when it didn't.  I did a bit more digging and found that the CRC is actually very simple, so crccheck package is not needed.  I've removed it from the interface and the setup.py file.  Note: without this in the setup file it only needs serial, which is identical the existing serial option.  I've left them both for now but happy to remove seeedstudio from the setup.py file altogether if preferred. 

Apart from that its done. 
(Do I need to resolve the conversations above to close out the 'Changes requested'?)

felixdivo:
(No, that will only make the thread of comments more readable. The review has to be re-requested.)

felixdivo:
@christiansandberg Do you now approve this PR?

felixdivo:
@shedfly The linter (see [here](https://travis-ci.org/hardbyte/python-can/jobs/549141499#L263) for the problems) and formatter (please reformat `can/interfaces/seeedstudio/seeedstudio.py` for example with [this](https://black.now.sh)) are not yet happy, but if they are we can immediately merge this. It should be relatively easy to fix this.

PR #568, opened 2019-05-03
Adding CAN FD 64 frame support to blf reader

pierreluctg:


can/io/blf.py
```diff
@@ -236,6 +250,29 @@ def __iter__(self):
                                       data=can_data[:length],
                                       channel=channel - 1)
                         yield msg
+                    elif obj_type == CAN_FD_MESSAGE_64:
+                        (
```
felixdivo: no braces needed here

can/io/blf.py
```diff
@@ -247,6 +284,8 @@ def __iter__(self):
                                       data=can_data[:dlc],
                                       channel=channel - 1)
                         yield msg
+                    # else:
+                    #     LOG.warning("Unknown object type (%d)", obj_type)
```
felixdivo: could raise a exception/warning in this case

pierreluctg: That is exactly why I left the commented code in.

Open for discussion on if we should log a warning of not here.

can/io/blf.py
pierreluctg: It's not required, however this is the preferred way to wrap long lines (according to PEP 8).

The "more" PEP 8 version would be:
```python
(
    channel,
    dlc,
    _,
    _,
    can_id,
    _,
    fd_flags,
) = CAN_FD_MSG_64_STRUCT.unpack_from(data, pos)[:7]
```

And matching the rest of the code in this module would be:
```python
(channel, dlc, _, _, can_id, _,
 fd_flags) = CAN_FD_MSG_64_STRUCT.unpack_from(data, pos)[:7]
```

I do not mind changing it to what you preferred.

can/io/blf.py
felixdivo: Okay, so: I think we should raise an Error and be yelled at the user/developer, since it probably means that he/she put in a file expecting it to be supported, but it is not.

can/io/blf.py
```diff
@@ -236,6 +250,29 @@ def __iter__(self):
...
+                        msg = Message(
+                            timestamp=timestamp,
+                            arbitration_id=can_id & 0x1FFFFFFF,
+                            is_extended_id=bool(can_id & CAN_MSG_EXT),
+                            is_remote_frame=bool(fd_flags & REMOTE_FLAG_64),
+                            is_fd=bool(fd_flags & EDL_64),
+                            bitrate_switch=bool(fd_flags & BRS_64),
+                            error_state_indicator=bool(fd_flags & ESI_64),
+                            dlc=length,
+                            data=can_data[:length],
```
christiansandberg: It would probably be more efficient something like this instead of unpacking the raw data above:

```python
data_offset = pos + CAN_FD_MSG_64_STRUCT.size

data[data_offset:data_offset + length]
```

Also I've realized it should really be the valid payload length and not the DLC to support remote frames correctly, but it is more of a sanity thing.

can/io/blf.py
christiansandberg: Log files contain lots of other information which should just be ignored. But it is good to have this commented out for debugging.

can/io/blf.py
felixdivo: I actually did not know the first version was encouraged. Leave it as is or change it to the second one, you are right in that it is well readable.

felixdivo: Interesting, I didn't know that. Then leave/change it as you prefer. (I'd choose the second one, but that's just me 😃)

felixdivo:
Could we add such a CAN FD 64 frame to `test/data/logfile.blf`?

christiansandberg:
The test file was copied from Toby’s C++ library. Maybe we should create our own and check that it is correct in a Vector application. I however do not have access to one anymore.

felixdivo:
Okay, so maybe just open a new issue for that and merge this PR?

felixdivo:
See #581.

PR #573, opened 2019-05-10
Fix stuct packing in build_bcm_header on 32-bit

### LINKED ISSUE #470: send_periodic fails on 32bit system using socketcan
Trying to use send_periodic on a socketcan bus fails with the following error message:

```

  File "./test.py", line 8, in <module>
    bus.send_periodic(msg, 0.01, 9)
  File "/usr/local/lib/python3.5/dist-packages/can/bus.py", line 201, in send_periodic
    task = self._send_periodic_internal(msg, period, duration)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 601, in _send_periodic_internal
    task = CyclicSendTask(bcm_socket, msg, period, duration)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 254, in __init__
    self._tx_setup(message)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 274, in _tx_setup
    send_bcm(self.bcm_socket, header + frame)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 206, in send_bcm
    raise can.CanError(base + "You are probably referring to a non-existing frame.")
can.CanError: Couldn't send CAN BCM frame. OS Error 22: Invalid argument
You are probably referring to a non-existing frame.
```

I'm using python-can 3.0.0 with python 3.5.3 on 32-bit Debian Stretch.

The same set of flags and values sent to the BCM is working in C.
Debugging shows the problem is a wrong size of the bcm_header. On my machine sizeof(struct bcm_msg_header) is 40 bytes, but the BCM header created in build_bcm_header is 36 bytes long. This causes the invalid argument error from the kernel.
Adding 4 padding bytes at the end of the bcm header remedies the error.

karlding:
Previously `build_bcm_header` relied upon adding a "0q" at the end of the
struct format string in order to force a 8-byte wide alignment on the
struct, in order to interop with the structure defined in the Linux
headers:

    struct bcm_msg_head {
      __u32 opcode;
      __u32 flags;
      __u32 count;
      struct bcm_timeval ival1, ival2;
      canid_t can_id;
      __u32 nframes;
      struct can_frame frames[0];
    };

The Python code assumed that alignof(long long) == 8 bytes in order to
accomplish this. However, on certain 32-bit platforms, this assumption
is not true, and alignof(long long) == 4.

As the struct module does not provide logic to derive alignments of its
types, this changes the packing logic to use ctypes. This exposes the
alignment of the struct members, allowing us to determine whether padding
bytes are necessary.

Fixes #470

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -118,6 +118,41 @@ def build_can_frame(msg):
     return CAN_FRAME_HEADER_STRUCT.pack(can_id, msg.dlc, flags) + data
 
 
+def bcm_header_factory(fields, alignment=8):
+    curr_stride = 0
+    results = []
+    pad_index = 0
+    for field in fields:
+        field_alignment = ctypes.alignment(field[1])
+        field_size = ctypes.sizeof(field[1])
+
+        # If the current stride index isn't a multiple of the alignment
+        # requirements of this field, then we must add padding bytes until we
+        # are aligned
+        while curr_stride % field_alignment != 0:
+            results.append(("pad_{}".format(pad_index), ctypes.c_uint8))
+            pad_index += 1
+            curr_stride += 1
+
+        # Now can it fit?
+        # Example: If this is 8 bytes and the type requires 4 bytes alignment
+        # then we can only fit when we're starting at 0. Otherwise, we will
+        # split across 2 strides.
+        #
+        # | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 |
+        results.append(field)
+        curr_stride += field_size
+
+    # Add trailing padding to align to a multiple of the largest scalar member
+    # in the structure
+    while curr_stride % alignment != 0:
+        results.append(("pad_{}".format(pad_index), ctypes.c_uint8))
+        pad_index += 1
+        curr_stride += 1
+
+    return type("BcmMsgHead", (ctypes.Structure,), {"_fields_": results})
+
+
 def build_bcm_header(opcode, flags, count, ival1_seconds, ival1_usec, ival2_seconds, ival2_usec, can_id, nframes):
     # == Must use native not standard types for packing ==
     # struct bcm_msg_head {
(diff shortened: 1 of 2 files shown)


karlding:
I've tested this by installing `can-utils` and running the `examples/cyclic.py` demo and observing SocketCAN with `candump` on the following:

- [x] Ubuntu 18.04 LTS (64-bit)
- [x] Debian Stretch (32-bit)
- [x] Ubuntu 16.04 LTS (32-bit)
- [x] Raspbian Stretch

karlding:
Oh, I also had tried writing a unit test by mocking the return values of `ctypes.sizeof` and `ctypes.alignment` to emulate what we'd see on these platforms, and then checking the size and alignment of the final structure (as returned from the factory function), but I was having some trouble getting it to correctly work after `patching` the functions.

I ended up hacking around it by poking at the `_fields_` class attribute and comparing the `_fields_` to an expected result. I feel like that's not ideal though, but I couldn't seem to get `ctypes.sizeof` and `ctypes.alignof` to work otherwise.

It looks something like this:

```python
import unittest

try:
    from unittest.mock import Mock
    from unittest.mock import patch
    from unittest.mock import call
except ImportError:
    from mock import Mock
    from mock import patch
    from mock import call

import ctypes

from can.interfaces.socketcan.socketcan import bcm_header_factory


class BcmMsgHeadTest(unittest.TestCase):
    def setUp(self):
        self._ctypes_sizeof = ctypes.sizeof
        self._ctypes_alignment = ctypes.alignment

    @patch("ctypes.sizeof")
    @patch("ctypes.alignment")
    def test_bcm_header_factory_32_bit_sizeof_long_4_alignof_long_4(
        self, ctypes_sizeof, ctypes_alignment
    ):
        def side_effect_ctypes_sizeof(value):
            if value == ctypes.c_long:
                return 4
            return self._ctypes_sizeof(value)

        def side_effect_ctypes_alignment(value):
            if value == ctypes.c_long:
                return 4
            return self._ctypes_alignment(value)

        ctypes_sizeof.side_effect = side_effect_ctypes_sizeof
        ctypes_alignment.side_effect = side_effect_ctypes_alignment

        fields = [
            ("opcode", ctypes.c_uint32),
            ("flags", ctypes.c_uint32),
            ("count", ctypes.c_uint32),
            ("ival1_tv_sec", ctypes.c_long),
            ("ival1_tv_usec", ctypes.c_long),
            ("ival2_tv_sec", ctypes.c_long),
            ("ival2_tv_usec", ctypes.c_long),
            ("can_id", ctypes.c_uint32),
            ("nframes", ctypes.c_uint32),
        ]
        BcmMsgHead = bcm_header_factory(fields)

        # TODO: Figure out if we can just use ctypes.sizeof here?
        expected_fields = [
            ("opcode", ctypes.c_uint32),
            ("flags", ctypes.c_uint32),
            ("count", ctypes.c_uint32),
            ("ival1_tv_sec", ctypes.c_long),
            ("ival1_tv_usec", ctypes.c_long),
            ("ival2_tv_sec", ctypes.c_long),
            ("ival2_tv_usec", ctypes.c_long),
            ("can_id", ctypes.c_uint32),
            ("nframes", ctypes.c_uint32),

            # We expect 4 bytes of padding
            ("pad_0", ctypes.c_uint8),
            ("pad_1", ctypes.c_uint8),
            ("pad_2", ctypes.c_uint8),
            ("pad_3", ctypes.c_uint8),
        ]
        self.assertEqual(expected_fields, BcmMsgHead._fields_)

    # And then a similar test for 64-bit results

if __name__ == "__main__":
    unittest.main()
```

Would you want me to add this somewhere?

can/interfaces/socketcan/socketcan.py
```diff
@@ -127,18 +162,31 @@ def build_bcm_header(opcode, flags, count, ival1_seconds, ival1_usec, ival2_seco
...
+        ("flags", ctypes.c_uint32),
+        ("count", ctypes.c_uint32),
+        ("ival1_tv_sec", ctypes.c_long),
+        ("ival1_tv_usec", ctypes.c_long),
+        ("ival2_tv_sec", ctypes.c_long),
+        ("ival2_tv_usec", ctypes.c_long),
+        ("can_id", ctypes.c_uint32),
+        ("nframes", ctypes.c_uint32),
+    ]
+    BcmMsgHead = bcm_header_factory(fields)
```
hardbyte: Consider calling the factory once at import time or socketcan bus instantiation time.

hardbyte:
Your test would be a good addition too. I assume your testing was with Python 2 and 3?

We could also see if we can carry out testing on 32bit systems on travis.

can/interfaces/socketcan/socketcan.py
karlding: Is there a specific test file that I should put it under? Not sure if I just missed one for the socketcan interface

karlding:
Sure! 

AFAIK, Travis only supports 64-bit, so presumably the only way to get 32-bit runners is to run the tests in a 32-bit container via Docker. Not sure if that's something we want to do as part of this change?

karlding:
Re-tested using the cyclic example on the same platforms using a variant of Python 2 and 3 for each, and it doesn't seem like I've broken anything. :joy: 

Would you like me to squash my commits? Or would you rather do that when merging (or not at all)?

can/interfaces/socketcan/socketcan.py
karlding: Done

hardbyte:
Thanks for fixing this :+1:

Issue #579, opened 2019-05-12
Adding code formatter/linter to workflow?


karlding:
What are your thoughts on enforcing automated formatting across the project? With various people contributing code, ensuring that consistent code style is followed is always a challenge. This often adds noise to PRs, and just is another thing for PR reviewers to worry about.

An automated formatter (I like [black](https://github.com/python/black) due to the lack of configuration options it exposes, but that's just my preference) would automatically format code so that all code is formatted consistently. This can also run in CI to ensure that commits have been formatted correctly before merging.

It'll probably cause a lot of conflicts with currently open PRs (and cause 1 giant diff of a commit), but maybe it's worth considering. And if this is adopted, the format can be done incrementally before flipping the CI checks to reduce the amount of pain with open PRs.

Similarly, what are your opinions on adding a linter (pylint, flake8, pycodestyle, etc.) to the workflow? An incremental approach could be taken to slowly make things comply with whatever rules are chosen.

hardbyte:
I'm very open to both black and a linter. 

Now is a fairly good time since v3.2.0 is essentially minted and most likely the next release will be v4 which will drop support for Python2 and start to add typing annotations. In another [project](https://github.com/data61/clkhash/issues/99#issuecomment-381027362) I was also supporting Python2 and wanted type annotations and to use black but they were a bit incompatible a year ago.

Due to the slow nature of reviewing and adding new interfaces I think there will never be a conflict free time to do it.

karlding:
Sounds good!

I believe a [patch](https://github.com/python/black/pull/642/files) went into black at the end of last year that took a stab at supporting that. Perhaps that supports enough of the use case that you were hoping to cover?

felixdivo:
I was about to propose this a few weeks back as well! I would love it. :+1:

felixdivo:
@karlding added a formatter in #602

felixdivo:
Related (about Pylint): #613

PR #592, opened 2019-05-17
Enable pylint on Travis CI builds

karlding:
Run pylint on all Travis CI builds for rules that we currently are
conformant to.

This adds a .pylintrc-wip file, which is intended to be a temporary
stop-gap that contains the current rules that the project is conformant
to. This allows us to incrementally enable pylint rules and have them
them enforced in subsequent code, rather than fixing the violations all
at once.

The .pylintrc file is intended to be the file containing all the
preferred set of rules that the project will eventually be compliant
against.

Currently the `.pylintrc` files are generated via the default rules:

```bash
pylint --generate-rcfile >> .pylintrc
pylint --disable=all --generate-rcfile >| .pylintrc-wip
```

I don't really have a preference in terms of which linter is being
used, so let me know if you feel strongly about a particular linter.


.pylintrc
```diff
@@ -0,0 +1,570 @@
...
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
```
felixdivo: Maybe use `0` since we might run it on our local multicore systems?

felixdivo: Applies to both files.

felixdivo:
This fixes #579, right?

hardbyte:
This PR only addresses the linter aspect of #579.

.pylintrc
karlding: Sorry, I was out of town for the last few days. Do you still want me to change this?

felixdivo: It's not too important. We can just do it next time we remove some other exclusions.

PR #594, opened 2019-05-18
Remove support for old Python versions

felixdivo:
This removes support for Python < 3.6, both CPython and PyPy in the codebase and CI pipelines.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -4,8 +4,6 @@
 Enable basic CAN over a PCAN USB device.
 """
 
-from __future__ import absolute_import, print_function, division
-
 import logging
 import sys
 import time
@@ -16,9 +14,8 @@ from can.bus import BusState
 from can.util import len2dlc, dlc2len
 from .basic import *
 
-boottimeEpoch = 0
 try:
-    import uptime
+    import uptime # isn't this from a library?
     import datetime
     boottimeEpoch = (uptime.boottime() - datetime.datetime.utcfromtimestamp(0)).total_seconds()
 except:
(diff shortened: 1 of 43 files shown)


felixdivo:
A nice side effect is the CI being a lot faster. ^^

felixdivo:
Question (@christiansandberg):

```python
try:
    socket.CAN_BCM
except AttributeError:
    HAS_NATIVE_SUPPORT = False
else:
    HAS_NATIVE_SUPPORT = True
```

Can this be removed, since it will evaluate to `HAS_NATIVE_SUPPORT = True` anyways? Or can it be `False` even on Python 3.6+?

christiansandberg:
You can assume that native support is always available.

felixdivo:
I'll open two more separate PRs: One for removing deprecated members and one for changing the calls to super() to the Python 3 syntax.

felixdivo:
I tried to split the conversion in multiple parts but this one is the first and biggest one. The others are simpler. This PR is now done (apart from the reviews coming in 😅).

@christiansandberg The socketcan change was applied as well.

can/interfaces/pcan/pcan.py
```diff
@@ -16,9 +14,8 @@
 from can.util import len2dlc, dlc2len
 from .basic import *
 
-boottimeEpoch = 0
 try:
-    import uptime
+    import uptime # isn't this from a library?
```
christiansandberg: Correct. Keep this as before. It will produce absolute timestamps if uptime is installed.

can/listener.py
```diff
@@ -133,42 +124,41 @@ def stop(self):
...
+        Retrieve the latest message when awaited for::
         
-        def __anext__(self):
-            return self.buffer.get()
+            msg = await reader.get_message()
+
+        :rtype: can.Message
+        :return: The CAN message.
+        """
+        return self.buffer.get()
```
christiansandberg: Now we can mark this function `async` to make it clearer that it is an asynchronous function.

```python
async def get_message(self):
    return await self.buffer.get()
```

PR #596, opened 2019-05-18
Modernize all the test cases

felixdivo:
This modernizes all python files in `test/`. I basically ran `2to3` on it and the went over them by hand to correct or add some changes manually.


felixdivo:
This requires #594 to be merged first.

test/back2back_test.py
```diff
@@ -5,8 +5,6 @@
 This module tests two virtual buses attached to each other.
 """
 
-from __future__ import absolute_import, print_function
```
hardbyte: :rocket: we are in the future.

test/test_kvaser.py
```diff
@@ -8,10 +8,7 @@
 import time
 import logging
 import unittest
-try:
-    from unittest.mock import Mock, patch
-except ImportError:
-    from mock import patch, Mock
```
hardbyte: We can remove the `mock` dependency if not already done.

felixdivo: Done. As well as the future dependency. Six will be removed separately, I opened an issue for it.

PR #601, opened 2019-05-19
Remove some linter exclusions

felixdivo:
This basically removes all Pylint rules we were currently skipping that did not involve too much rewriting.


felixdivo:
Ah shoot. I messed with the `*args` position. Let me fix that first.

felixdivo:
The coverage went down since I deleted quite a lot of unused imports.

can/interfaces/pcan/basic.py
```diff
@@ -658,13 +658,15 @@ def GetValue(
           A touple with 2 values
         """
         try:
-            if Parameter == PCAN_API_VERSION or Parameter == PCAN_HARDWARE_NAME or Parameter == PCAN_CHANNEL_VERSION or Parameter == PCAN_LOG_LOCATION or Parameter == PCAN_TRACE_LOCATION or Parameter == PCAN_BITRATE_INFO_FD or Parameter == PCAN_IP_ADDRESS:
+            if Parameter in (PCAN_API_VERSION, PCAN_HARDWARE_NAME, PCAN_CHANNEL_VERSION,
```
hardbyte: nit: use a set with `{PCAN_API_VERSION, ...}` instead of a tuple

can/interfaces/pcan/pcan.py
```diff
@@ -164,7 +163,7 @@ def __init__(self, channel='PCAN_USBBUS1', state=BusState.ACTIVE, bitrate=500000
         self.m_PcanHandle = globals()[channel]
 
         if state is BusState.ACTIVE or state is BusState.PASSIVE:
-            self.state = state
+            self._state = state
```
hardbyte: Unless you're adding a property to expose this it should stay as `self.state` as it is part of the public Bus api.

https://python-can.readthedocs.io/en/master/bus.html#can.BusABC.state

felixdivo: Hm yeah, I thought about that. But it declares `self._state` syntactically outside of `__init__`. Will revert it and add an ignore comment.

can/interfaces/pcan/basic.py
felixdivo: Sounds good. I didn't know there was a special syntax for it. 👀

PR #602, opened 2019-05-21
Enable black formatter in CI

karlding:
Here's the second part that enables black formatter checks on the project.

The black formatter returns `0` if nothing would change, and `1` if files would be reformatted, which allows the CI job to fail.

```bash
black --check .
```

In order to provide some more verbose error reporting, we also run the following to provide the diff and show the modifications (although this isn't strictly necessary).

```bash
black --diff .
```

I've split the commits for easier review so that the final commit is purely the results of running

```bash
black .
```


karlding:
Also, I wrote this quick and dirty bash script to dump all the Python files that are currently touched by open PRs.

```bash
ACTIVE_PRS=(290 539 554 567)
DIFF_FILE="output.txt"

echo "" >| "${DIFF_FILE}"

for PR in "${ACTIVE_PRS[@]}"; do
  git fetch origin pull/"${PR}"/head:pr-"${PR}"
  git diff --name-only develop...pr-"${PR}" | grep '.py' | tee -a "${DIFF_FILE}"
done

sort "${DIFF_FILE}" -o "${DIFF_FILE}" -u
```

I looked at those PRs, and the changes don't seem too bad in context of this giant format patch. Otherwise, if we want to, we can simply exclude these files until those PRs are merged.

```
can/__init__.py
can/interface.py
can/interfaces/__init__.py
can/interfaces/pcan/pcan.py
can/interfaces/serial/__init__.py
can/interfaces/serial/serial_can.py
can/interfaces/serial/simpleserial.py
can/interfaces/serial/slcan.py
can/interfaces/slcan.py
can/interfaces/usb_can_analyzer/__init__.py
can/interfaces/usb_can_analyzer/usb_can_analyzer.py
can/io/__init__.py
can/io/mf4.py
examples/receive_all.py
examples/serial_com.py
setup.py
test/interface_simpleserial_test.py
test/interface_slcan_test.py
test/interface_test.py
test/logformats_test.py
test/serial_test.py
```

can/interfaces/__init__.py
```diff
@@ -10,25 +10,29 @@
...
-    'ixxat':            ('can.interfaces.ixxat',            'IXXATBus'),
-    'nican':            ('can.interfaces.nican',            'NicanBus'),
-    'iscan':            ('can.interfaces.iscan',            'IscanBus'),
-    'virtual':          ('can.interfaces.virtual',          'VirtualBus'),
-    'neovi':            ('can.interfaces.ics_neovi',        'NeoViBus'),
-    'vector':           ('can.interfaces.vector',           'VectorBus'),
-    'slcan':            ('can.interfaces.slcan',            'slcanBus'),
-    'canalystii':       ('can.interfaces.canalystii',       'CANalystIIBus'),
-    'systec':           ('can.interfaces.systec',           'UcanBus')
+    "kvaser": ("can.interfaces.kvaser", "KvaserBus"),
```
felixdivo: Do we really want this new formatting here? Or is there no option to exclude a block?

can/interfaces/canalystii.py
```diff
@@ -107,7 +115,12 @@ def __init__(self, channel, device=0, baud=None, Timing0=None, Timing1=None, can
             logger.error("VCI_OpenDevice Error")
 
         for channel in self.channels:
-            if CANalystII.VCI_InitCAN(VCI_USBCAN2, self.device, channel, byref(self.init_config)) == STATUS_ERR:
+            if (
```
felixdivo: Maybe:

```suggestion
            status = CANalystII.VCI_InitCAN(VCI_USBCAN2, self.device, channel, byref(self.init_config))
            if status == == STATUS_ERR:
```

can/interfaces/canalystii.py
```diff
@@ -147,7 +171,17 @@ def _recv_internal(self, timeout=None):
 
         timeout = -1 if timeout is None else int(timeout * 1000)
 
-        if CANalystII.VCI_Receive(VCI_USBCAN2, self.device, self.channels[0], byref(raw_message), 1, timeout) <= STATUS_ERR:
+        if (
```
felixdivo: split into two statements?

felixdivo:
I didn't go through all of the changes, since it was just tom much.

felixdivo:
We could also add the badge: https://github.com/python/black#show-your-style

felixdivo:
I did really like the previous semantics of quotes: Double quotes for user facing text, single quotes for everything else. But I guess that's a matter of taste as well.

.travis.yml
```diff
@@ -92,6 +92,14 @@ jobs:
...
         # re-introduced
         - pylint --rcfile=.pylintrc-wip can/
+    - stage: linter
+      name: "Formatting Checks"
+      python: "3.7"
+      before_install:
+        - travis_retry pip install -r requirements-lint.txt
+      script:
+        - black --diff .
+        - black --check .
```
felixdivo: I could image `black --check --verbose .` being enough, [since black itself uses it](https://github.com/python/black/blob/master/.travis.yml#L25)

can/interfaces/__init__.py
karlding: black provides a way to exclude sections from formatting, but IMO that kind of goes against the point of having a formatter.

If someone feels strongly about keeping it untouched by the formatter, I can add

```python
# fmt: off

# ...

# fmt: on
```

Issue #269, opened 2018-02-23
CAN FD support


felixdivo:
- [ ] That is not yet tested by `logformats_test.py`
- [ ] Messages are not correctly written & read by `can.io.*`

PR #604, opened 2019-05-23
Adding CAN FD frame support to asc writer

pierreluctg:


pierreluctg:
I will add proper tests as soon as possible...

felixdivo:
I did a rebase but maybe some things need to be adjusted because the black formatter requires it.

felixdivo:
This could also resolve the old #269.

christiansandberg:
I’m doing some refactoring of the blf module and in the meantime adding general CAN-FD testing capability.

PR #610, opened 2019-06-02
Add support for multiple Cyclic Messages in Task

### LINKED ISSUE #606: Incorrect frame re-transmission when arbitration ID is duplicate
When arbitration ID is duplicate, only the last message with that arbitration ID is re-transmitted according to period.

Below is a code snippet of how messages are defined.
Note that pm_snapshot_msg, vin1_msg, vin2_msg & vin3_msg have the same arbitration_id 0x400.

```
self.pm_msg = can.Message(
            arbitration_id=0xFF, 
            data=self.pm_data, 
            extended_id=False, 
            dlc=8)

        self.pm_snapshot_msg = can.Message(
            arbitration_id=0x400, 
            data=self.pm_snapshot_data,
            extended_id=False, 
            dlc=8)
        self.nm_msg = can.Message(
            arbitration_id=0x51E, 
            data=self.nm_data, 
            extended_id=False, 
            dlc=8)

        self.vs_msg = can.Message(
            arbitration_id=0x11, 
            data=self.vs_data, 
            extended_id=False, 
            dlc=8)

        self.vin1_msg = can.Message(
            arbitration_id=0x400, 
            data=self.vin1_data,
            extended_id=False, 
            dlc=8)

        self.vin2_msg = can.Message(
            arbitration_id=0x400, 
            data=self.vin2_data,
            extended_id=False, 
            dlc=8)

        self.vin3_msg = can.Message(
            arbitration_id=0x400, 
            data=self.vin3_data,
            extended_id=False, 
            dlc=8)
```

Below is code where messages are sent.

```
def wake(self):
        if self.can_state == "sleep":
            self.nm_data[1] = 0x2
            self.nm_msg.data = self.nm_data
            self.nm_task = self.bus.send_periodic(self.nm_msg, self.nm_period)
            if not isinstance(self.nm_task, can.ModifiableCyclicTaskABC):
                print("This interface doesn't seem to support modification")
                task.stop()
                exit(-1)
            self.pm_task = self.bus.send_periodic(self.pm_msg, self.pm_period)
            self.pm_snapshot_task = self.bus.send_periodic(self.pm_snapshot_msg, self.pm_snapshot_period)
            self.vs_task = self.bus.send_periodic(self.vs_msg, self.vs_period)
            self.vin1_task = self.bus.send_periodic(self.vin1_msg, self.vin_period)
            self.vin2_task = self.bus.send_periodic(self.vin2_msg, self.vin_period)
            self.vin3_task = self.bus.send_periodic(self.vin3_msg, self.vin_period)
            self.can_state = "active"
        else:
            print("Sim already running")
```

Below is the candump after wake() is called. Note that all messages are transmitted at first, but after that only the last message with arbitration id of 0x400 is sent, namely vin3_msg, which was the last message to be defined with arbitration id of 0x400.

```
can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  400   [8]  B0 A1 B1 C1 D1 E1 04 00 <-- pm_snapshot_msg 0x400
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  400   [8]  57 00 00 00 4B 4C 31 54  <-- vin1_msg 0x400
  can0  400   [8]  58 44 35 44 45 33 42 42  <-- vin2_msg 0x400
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  51E   [8]  10 02 00 00 00 00 00 00
  can0  400   [8]  59 31 39 35 37 35 36 00  <-- vin3_msg 0x400
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00
  can0  0FF   [8]  00 00 04 60 00 00 00 00
  can0  011   [8]  00 C0 0C 00 00 00 00 00

```

karlding:
This adds support for multiple Cyclic Messages in a Cyclic Task. The
default implementation is also changed to provide support for this, by
iterating over the list of Cyclic Messages.

The SocketCAN interface takes advantage of the Linux BCM APIs to do so,
while the IXXAT interface maintains its original behaviour.

This also introduces a new example that illustrates how to use Cyclic
Messages, backed by the SocketCAN interface.

Fixes #606

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -12,6 +12,8 @@ import logging
 import threading
 import time
 
+import can
+
 log = logging.getLogger("can.bcm")
 
 
@@ -34,28 +36,65 @@ class CyclicSendTaskABC(CyclicTask):
     Message send task with defined period
     """
 
-    def __init__(self, message, period):
+    def __init__(self, messages, period):
         """
-        :param can.Message message: The message to be sent periodically.
-        :param float period: The rate in seconds at which to send the message.
+        :param Union[Sequence[can.Message], can.Message] messages:
+            The messages to be sent periodically.
+        :param float period: The rate in seconds at which to send the messages.
         """
-        self.message = message
-        self.can_id = message.arbitration_id
-        self.arbitration_id = message.arbitration_id
+        messages = self._check_and_convert_messages(messages)
+
+        # Take the Arbitration ID of the first element
+        self.arbitration_id = messages[0].arbitration_id
         self.period = period
-        super().__init__()
+        self.messages = messages
+
+    @staticmethod
+    def _check_and_convert_messages(messages):
+        """Helper function to convert a Message or Sequence of messages into a
+        tuple, and raises an error when the given value is invalid.
+
+        Performs error checking to ensure that all Messages have the same
+        arbitration ID and channel.
+
+        Should be called when the cyclic task is initialized
+        """
+        if not isinstance(messages, (list, tuple)):
+            if isinstance(messages, can.Message):
+                messages = [messages]
+            else:
+                raise ValueError("Must be either a list, tuple, or a Message")
+        if not messages:
+            raise ValueError("Must be at least a list or tuple of length 1")
+        messages = tuple(messages)
+
+        all_same_id = all(
+            message.arbitration_id == messages[0].arbitration_id for message in messages
+        )
+        if not all_same_id:
+            raise ValueError("All Arbitration IDs should be the same")
+
+        all_same_channel = all(
+            message.channel == messages[0].channel for message in messages
+        )
+        if not all_same_channel:
+            raise ValueError("All Channel IDs should be the same")
+
+        return messages
 
 
 class LimitedDurationCyclicSendTaskABC(CyclicSendTaskABC):
-    def __init__(self, message, period, duration):
+    def __init__(self, messages, period, duration):
         """Message send task with a defined duration and period.
 
-        :param can.Message message: The message to be sent periodically.
-        :param float period: The rate in seconds at which to send the message.
+        :param Union[Sequence[can.Message], can.Message] messages:
+            The messages to be sent periodically.
+        :param float period: The rate in seconds at which to send the messages.
         :param float duration:
-            The duration to keep sending this message at given rate.
+            Approximate duration in seconds to continue sending messages. If
+            no duration is provided, the task will continue indefinitely.
         """
-        super().__init__(message, period)
+        super().__init__(messages, period)
         self.duration = duration
 
 
@@ -71,33 +110,61 @@ class RestartableCyclicTaskABC(CyclicSendTaskABC):
 class ModifiableCyclicTaskABC(CyclicSendTaskABC):
     """Adds support for modifying a periodic message"""
 
-    def modify_data(self, message):
-        """Update the contents of this periodically sent message without altering
-        the timing.
+    def _check_modified_messages(self, messages):
+        """Helper function to perform error checking when modifying the data in
+        the cyclic task.
 
-        :param can.Message message:
-          The message with the new :attr:`can.Message.data`.
-          Note: The arbitration ID cannot be changed.
+        Performs error checking to ensure the arbitration ID and the number of
+        cyclic messages hasn't changed.
+
+        Should be called when modify_data is called in the cyclic task.
         """
-        self.message = message
+        if len(self.messages) != len(messages):
+            raise ValueError(
+                "The number of new cyclic messages to be sent must be equal to "
+                "the number of messages originally specified for this task"
+            )
+        if self.arbitration_id != messages[0].arbitration_id:
+            raise ValueError(
+                "The arbitration ID of new cyclic messages cannot be changed "
+                "from when the task was created"
+            )
+
+    def modify_data(self, messages):
+        """Update the contents of the periodically sent messages, without
+        altering the timing.
+
+        :param Union[Sequence[can.Message], can.Message] messages:
+            The messages with the new :attr:`can.Message.data`.
+
+            Note: The arbitration ID cannot be changed.
+
+            Note: The number of new cyclic messages to be sent must be equal
+            to the original number of messages originally specified for this
+            task.
+        """
+        messages = self._check_and_convert_messages(messages)
+        self._check_modified_messages(messages)
+
+        self.messages = messages
 
 
 class MultiRateCyclicSendTaskABC(CyclicSendTaskABC):
     """A Cyclic send task that supports switches send frequency after a set time.
     """
 
-    def __init__(self, channel, message, count, initial_period, subsequent_period):
+    def __init__(self, channel, messages, count, initial_period, subsequent_period):
         """
         Transmits a message `count` times at `initial_period` then continues to
-        transmit message at `subsequent_period`.
+        transmit messages at `subsequent_period`.
 
         :param channel: See interface specific documentation.
-        :param can.Message message:
+        :param Union[Sequence[can.Message], can.Message] messages:
         :param int count:
         :param float initial_period:
         :param float subsequent_period:
         """
-        super().__init__(channel, message, subsequent_period)
+        super().__init__(channel, messages, subsequent_period)
 
 
 class ThreadBasedCyclicSendTask(
@@ -105,10 +172,10 @@ class ThreadBasedCyclicSendTask(
 ):
     """Fallback cyclic send task using thread."""
 
-    def __init__(self, bus, lock, message, period, duration=None):
-        super().__init__(message, period, duration)
+    def __init__(self, bus, lock, messages, period, duration=None):
+        super().__init__(messages, period, duration)
         self.bus = bus
-        self.lock = lock
+        self.send_lock = lock
         self.stopped = True
         self.thread = None
         self.end_time = time.time() + duration if duration else None
@@ -120,23 +187,25 @@ class ThreadBasedCyclicSendTask(
     def start(self):
         self.stopped = False
         if self.thread is None or not self.thread.is_alive():
-            name = "Cyclic send task for 0x%X" % (self.message.arbitration_id)
+            name = "Cyclic send task for 0x%X" % (self.messages[0].arbitration_id)
             self.thread = threading.Thread(target=self._run, name=name)
             self.thread.daemon = True
             self.thread.start()
 
     def _run(self):
+        msg_index = 0
         while not self.stopped:
             # Prevent calling bus.send from multiple threads
-            with self.lock:
+            with self.send_lock:
                 started = time.time()
                 try:
-                    self.bus.send(self.message)
+                    self.bus.send(self.messages[msg_index])
                 except Exception as exc:
                     log.exception(exc)
                     break
             if self.end_time is not None and time.time() >= self.end_time:
                 break
+            msg_index = (msg_index + 1) % len(self.messages)
             # Compensate for the time it takes to send the message
             delay = self.period - (time.time() - started)
             time.sleep(max(0.0, delay))
--- a/can/bus.py
+++ b/can/bus.py
@@ -5,6 +5,7 @@ Contains the ABC bus implementation and its documentation.
 """
 
 from abc import ABCMeta, abstractmethod
+import can
 import logging
 import threading
 from time import time
@@ -163,8 +164,8 @@ class BusABC(metaclass=ABCMeta):
         """
         raise NotImplementedError("Trying to write to a readonly bus?")
 
-    def send_periodic(self, msg, period, duration=None, store_task=True):
-        """Start sending a message at a given period on this bus.
+    def send_periodic(self, msgs, period, duration=None, store_task=True):
+        """Start sending messages at a given period on this bus.
 
         The task will be active until one of the following conditions are met:
 
@@ -174,12 +175,12 @@ class BusABC(metaclass=ABCMeta):
         - :meth:`BusABC.stop_all_periodic_tasks()` is called
         - the task's :meth:`CyclicTask.stop()` method is called.
 
-        :param can.Message msg:
-            Message to transmit
+        :param Union[Sequence[can.Message], can.Message] msgs:
+            Messages to transmit
         :param float period:
             Period in seconds between each message
         :param float duration:
-            The duration to keep sending this message at given rate. If
+            Approximate duration in seconds to continue sending messages. If
             no duration is provided, the task will continue indefinitely.
         :param bool store_task:
             If True (the default) the task will be attached to this Bus instance.
@@ -191,18 +192,26 @@ class BusABC(metaclass=ABCMeta):
 
         .. note::
 
-            Note the duration before the message stops being sent may not
+            Note the duration before the messages stop being sent may not
             be exactly the same as the duration specified by the user. In
             general the message will be sent at the given rate until at
             least **duration** seconds.
 
         .. note::
 
-            For extremely long running Bus instances with many short lived tasks the default
-            api with ``store_task==True`` may not be appropriate as the stopped tasks are
-            still taking up memory as they are associated with the Bus instance.
+            For extremely long running Bus instances with many short lived
+            tasks the default api with ``store_task==True`` may not be
+            appropriate as the stopped tasks are still taking up memory as they
+            are associated with the Bus instance.
         """
-        task = self._send_periodic_internal(msg, period, duration)
+        if not isinstance(msgs, (list, tuple)):
+            if isinstance(msgs, can.Message):
+                msgs = [msgs]
+            else:
+                raise ValueError("Must be either a list, tuple, or a Message")
+        if not msgs:
+            raise ValueError("Must be at least a list or tuple of length 1")
+        task = self._send_periodic_internal(msgs, period, duration)
         # we wrap the task's stop method to also remove it from the Bus's list of tasks
         original_stop_method = task.stop
 
@@ -221,21 +230,22 @@ class BusABC(metaclass=ABCMeta):
 
         return task
 
-    def _send_periodic_internal(self, msg, period, duration=None):
+    def _send_periodic_internal(self, msgs, period, duration=None):
         """Default implementation of periodic message sending using threading.
 
         Override this method to enable a more efficient backend specific approach.
 
-        :param can.Message msg:
-            Message to transmit
+        :param Union[Sequence[can.Message], can.Message] msgs:
+            Messages to transmit
         :param float period:
             Period in seconds between each message
         :param float duration:
-            The duration to keep sending this message at given rate. If
+            The duration between sending each message at the given rate. If
             no duration is provided, the task will continue indefinitely.
         :return:
-            A started task instance. Note the task can be stopped (and depending on
-            the backend modified) by calling the :meth:`stop` method.
+            A started task instance. Note the task can be stopped (and
+            depending on the backend modified) by calling the :meth:`stop`
+            method.
         :rtype: can.broadcastmanager.CyclicSendTaskABC
         """
         if not hasattr(self, "_lock_send_periodic"):
@@ -244,7 +254,7 @@ class BusABC(metaclass=ABCMeta):
                 threading.Lock()
             )  # pylint: disable=attribute-defined-outside-init
         task = ThreadBasedCyclicSendTask(
-            self, self._lock_send_periodic, msg, period, duration
+            self, self._lock_send_periodic, msgs, period, duration
         )
         return task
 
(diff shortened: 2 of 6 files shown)


can/broadcastmanager.py
```diff
@@ -34,28 +35,43 @@ class CyclicSendTaskABC(CyclicTask):
...
         """
-        :param can.Message message: The message to be sent periodically.
-        :param float period: The rate in seconds at which to send the message.
+        :param List[can.Message] messages: The message to be sent periodically.
+        :param float period: The rate in seconds at which to send the messages.
         """
-        self.message = message
-        self.can_id = message.arbitration_id
-        self.arbitration_id = message.arbitration_id
+        if not isinstance(messages, list):
```
karlding: TBH I'm not really a fan of keeping backwards compatibility and supporting both a `List` and a `Message` type. It seems kludgy to muddle the API like that. Plus, a single Message can still be supported by a list of length 1...

christiansandberg: No one has really requested this feature for years and neither one of IXXAT, Kvaser, NI-CAN, or Systec interfaces which provide cyclic message transmissions have seen the need for this and won't support it. Therefore I believe it is not justified to force everyone to change their implementations for the very few people that need this. At the very least we must have a deprecation period to allow implementations to adjust. A library that changes its API in backwards incompatible ways too often are not very popular among developers relying on it for critical applications.

Maybe it only has to be converted to a list at one place. Maybe in the `send_periodic()` method?

can/broadcastmanager.py
```diff
@@ -34,28 +35,43 @@ class CyclicSendTaskABC(CyclicTask):
...
+
+        all_same_id = all(
+            message.arbitration_id == messages[0].arbitration_id for message in messages
+        )
+        if not all_same_id:
+            raise ValueError("All Arbitration IDs should be the same")
+
+        # Take the Arbitration ID of the first element
+        self.can_id = messages[0].arbitration_id
+        self.arbitration_id = messages[0].arbitration_id
```
christiansandberg: Could we perhaps remove one of these now?

can/broadcastmanager.py
```diff
@@ -120,18 +136,20 @@ def stop(self):
...
     def _run(self):
+        msg_index = 0
         while not self.stopped:
             # Prevent calling bus.send from multiple threads
             with self.lock:
                 started = time.time()
                 try:
-                    self.bus.send(self.message)
+                    self.bus.send(self.messages[msg_index])
+                    msg_index = (msg_index + 1) % len(self.messages)
```
christiansandberg: You could move the msg_index increment outside the lock.

can/interfaces/ixxat/canlib.py
```diff
@@ -716,29 +716,38 @@ def shutdown(self):
...
-        self._msg.uMsgInfo.Bits.ext = 1 if msg.is_extended_id else 0
-        self._msg.uMsgInfo.Bits.rtr = 1 if msg.is_remote_frame else 0
-        self._msg.uMsgInfo.Bits.dlc = msg.dlc
-        for i, b in enumerate(msg.data):
-            self._msg.abData[i] = b
-        self.start()
+    def __init__(self, scheduler, msgs, period, duration, resolution):
+        super().__init__(msgs, period, duration)
+        # Only supports 1 Message in the group?
+        if len(self.messages) == 1:
```
christiansandberg: I would probably just do a check in the beginning instead.

```python
if len(self.messages) != 1:
    raise ValueError("IXXAT Interface only supports 1 element")
```

can/interfaces/ixxat/canlib.py
```diff
@@ -716,29 +716,38 @@ def shutdown(self):
...
+            raise ValueError("IXXAT Interface only supports 1 element")
+            # TODO: Fall back to the default implementation?
 
     def start(self):
         """Start transmitting message (add to list if needed)."""
-        if self._index is None:
-            self._index = ctypes.c_uint32()
-            _canlib.canSchedulerAddMessage(self._scheduler, self._msg, self._index)
-        _canlib.canSchedulerStartMessage(self._scheduler, self._index, self._count)
+        if len(self.messages) == 1:
```
christiansandberg: Don't think we need to check self.messages again since it should not have changed.

can/interfaces/socketcan/socketcan.py
```diff
@@ -296,38 +295,47 @@ class CyclicSendTask(
...
-            ival2 = 0
-        else:
-            count = 0
-            ival1 = 0
-            ival2 = self.period
+        header = bytearray()
+        body = bytearray()
+        for message in messages:
+            self.can_id_with_flags = _add_flags_to_can_id(message)
+            self.flags = CAN_FD_FRAME if message.is_fd else 0
```
christiansandberg: These two can be set outside the loop to the value of the first message.

can/interfaces/socketcan/socketcan.py
```diff
@@ -296,38 +295,47 @@ class CyclicSendTask(
...
+            if self.duration:
+                count = int(self.duration / self.period)
+                ival1 = self.period
+                ival2 = 0
+            else:
+                count = 0
+                ival1 = 0
+                ival2 = self.period
+            body += build_can_frame(message)
+            log.debug("Sending BCM command")
```
christiansandberg: Move outside loop.

can/interfaces/socketcan/socketcan.py
```diff
@@ -341,22 +349,33 @@ def stop(self):
...
+        self.messages = messages
+
+        header = bytearray()
+        body = bytearray()
+        header = build_bcm_update_header(
+            can_id=self.can_id_with_flags, msg_flags=self.flags, nframes=len(messages)
+        )
+        for message in messages:
+            self.can_id_with_flags = _add_flags_to_can_id(message)
+            self.flags = CAN_FD_FRAME if message.is_fd else 0
```
christiansandberg: Move outside loop using first entry.

can/interfaces/socketcan/socketcan.py
```diff
@@ -341,22 +349,33 @@ def stop(self):
...
+            body += build_can_frame(message)
+            log.debug("Sending BCM command")
```
christiansandberg: Move outside loop.

can/broadcastmanager.py
karlding: Maybe I'm missing something, but why is this lock even necessary? What is this lock intended to protect?

From what I understand, there is a `1:1` mapping between cyclic tasks and threads, so `_run` is essentially running in its own context for each thread. As such, the thread can get preempted, but nothing should be modifying its context state while it is context switched out.

According to `git blame`, you were the one that added this locking, so I thought I'd take the opportunity to ask you :joy:

If this lock is actually protecting against other threads poking into the context, then I believe the increment still needs to be protected by the lock, since Python doesn't provide atomics.

felixdivo: I think the lock prevents `Bus.send(...)` from being called concurrently, and not the update of `msg_index`. This is required since `Bus.send(...)` is not guaranteed to be thread safe.

karlding: Ah, thanks!

can/broadcastmanager.py
```diff
@@ -34,28 +35,42 @@ class CyclicSendTaskABC(CyclicTask):
     Message send task with defined period
     """
 
-    def __init__(self, message, period):
+    def __init__(self, messages, period):
         """
-        :param can.Message message: The message to be sent periodically.
-        :param float period: The rate in seconds at which to send the message.
+        :param List[can.Message] messages: The message to be sent periodically.
```
pierreluctg: This should probably be something like `Union[List[can.Message], can.Message]`, and the docstring should be updated.

can/broadcastmanager.py
```diff
@@ -34,28 +35,42 @@ class CyclicSendTaskABC(CyclicTask):
...
+        :param float period: The rate in seconds at which to send the messages.
         """
-        self.message = message
-        self.can_id = message.arbitration_id
-        self.arbitration_id = message.arbitration_id
+        if not isinstance(messages, list):
```
pierreluctg: `tuple` should also be supported. And unless we want to allow modification of `self.messages` sequence without using `ModifiableCyclicTaskABC.modify_data`, we should be using a `tuple` for `self.messages`

can/broadcastmanager.py
```diff
@@ -120,23 +135,25 @@ def stop(self):
...
     def _run(self):
+        msg_index = 0
         while not self.stopped:
             # Prevent calling bus.send from multiple threads
-            with self.lock:
+            with self.send_lock:
                 started = time.time()
                 try:
-                    self.bus.send(self.message)
+                    self.bus.send(self.messages[msg_index])
```
pierreluctg: There is a chance here that the `msg_index` is no longer valid if `self.messages` is modified.

can/interfaces/socketcan/socketcan.py
```diff
@@ -296,24 +295,27 @@ class CyclicSendTask(
 
     """
 
-    def __init__(self, bcm_socket, message, period, duration=None):
+    def __init__(self, bcm_socket, messages, period, duration=None):
         """
         :param bcm_socket: An open bcm socket on the desired CAN channel.
-        :param can.Message message: The message to be sent periodically.
+        :param List[can.Message] messages: The message to be sent periodically.
```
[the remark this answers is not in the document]
karlding: I think this should be handled implicitly via the multiple inheritance initializers? I believe the `super().__init__` should resolve to `LimitedDurationCyclicSendTaskABC`.

I've changed the code to use the assigned `self.messages` instead.

can/broadcastmanager.py
karlding: The Linux SocketCAN BCM APIs require that the new array size must be equal to the given array size. Is this a restriction that we also want to impose in the generic implementation?

can/broadcastmanager.py
karlding: Good point @pierreluctg!

I've changed the internal representation to a `tuple` type.

examples/cyclic_multiple.py
```diff
@@ -0,0 +1,147 @@
...
+        )
+    )
+    messages.append(
+        can.Message(
+            arbitration_id=0x401,
+            data=[0x55, 0x55, 0x55, 0x55, 0x55, 0x55],
+            is_extended_id=False,
+        )
+    )
+    task = bus.send_periodic(messages, 1)
```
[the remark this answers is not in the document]
karlding: Do you mean to test that we're receiving these CAN messages over the SocketCAN interface? I would consider exercising periodic functionality an integration test, do we have a specific place for these?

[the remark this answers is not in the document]
karlding: I've added `test_socketcan_cyclic_multiple.py`, which is hopefully in line with what you were hoping?

can/broadcastmanager.py
```diff
@@ -71,44 +89,44 @@ def start(self):
...
 
-        :param can.Message message:
-          The message with the new :attr:`can.Message.data`.
-          Note: The arbitration ID cannot be changed.
+        :param Union[List[can.Message], tuple(can.Message), can.Message] messages:
+            The messages with the new :attr:`can.Message.data`.
+            Note: The arbitration ID cannot be changed.
         """
-        self.message = message
+        self.messages = messages
```
pierreluctg: `messages` need to get the same processing as in `CyclicSendTaskABC.__init__`

can/bus.py
```diff
@@ -191,18 +192,26 @@ def send_periodic(self, msg, period, duration=None, store_task=True):
...
-            still taking up memory as they are associated with the Bus instance.
+            For extremely long running Bus instances with many short lived
+            tasks the default api with ``store_task==True`` may not be
+            appropriate as the stopped tasks are still taking up memory as they
+            are associated with the Bus instance.
         """
-        task = self._send_periodic_internal(msg, period, duration)
+        if not isinstance(msgs, (list, tuple)):
+            if isinstance(msgs, can.Message):
+                msgs = [msgs]
```
pierreluctg: Minor but you may want to start using tuple from here.

`msgs = (msgs,)`

can/broadcastmanager.py
hardbyte: That seems a reasonable restriction

can/broadcastmanager.py
```diff
@@ -71,44 +102,55 @@ def start(self):
...
 
-        :param can.Message message:
-          The message with the new :attr:`can.Message.data`.
-          Note: The arbitration ID cannot be changed.
+            Note: The number of new cyclic messages to be sent must be equal
+            to the original number of messages originally specified for this
+            task.
         """
-        self.message = message
+        messages = self._check_and_convert_messages(messages)
```
pierreluctg: Since we are changing the code here, maybe we should enforce the fact that the arbitration ID cannot be changed.

examples/cyclic_multiple.py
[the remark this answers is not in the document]
hardbyte: Perfect!

can/interfaces/socketcan/socketcan.py
```diff
@@ -341,22 +355,38 @@ def stop(self):
...
+
+        Note: The messages must all have the same
+        :attr:`~can.Message.arbitration_id` like the first message.
 
-        Note the Message must have the same :attr:`~can.Message.arbitration_id`
-        like the first message.
+        Note: The number of new cyclic messages to be sent must be equal to the
+        original number of messages originally specified for this task.
+
+        :param Union[List[can.Message], tuple(can.Message), can.Message] messages:
```
hardbyte: nit: I'd do this with typing.Sequence: `Union[Sequence[can.Message], can.Message]`

felixdivo: Or [typing.Iterable](https://docs.python.org/3/library/typing.html#typing.Iterable)?

karlding: @felixdivo correct me if I'm wrong, but I don't think you can get the length of an `Iterable` without actually performing the iteration and storing the result.

If want to support an `Iterable`, do we want to deal with buffering potentially unbounded input? Or implement specific bounded iterators for each interface that only iterates over the maximum number of elements the interface task can support, and errors otherwise?

felixdivo: Ah I see. We could wrap it in a list if it's not already a sequence. Implementing something like that ourselves seems overkill. If you want, stay with what you have or a sequence.

can/broadcastmanager.py
```diff
@@ -34,28 +36,86 @@ class CyclicSendTaskABC(CyclicTask):
...
+
+        all_same_channel = all(
+            message.channel == messages[0].channel for message in messages
+        )
+        if not all_same_channel:
+            raise ValueError("All Channel IDs should be the same")
+
+        return messages
+
+    def _check_modified_messages(self, messages):
```
pierreluctg: Not sure this is the write location for this method since it's only ever use by `ModifiableCyclicTaskABC.modify_data`. Maybe it should be in the `ModifiableCyclicTaskABC` class or nested inside `ModifiableCyclicTaskABC.modify_data`.

can/broadcastmanager.py
```diff
@@ -71,44 +110,72 @@ def start(self):
...
+
+            Note: The arbitration ID cannot be changed.
+
+            Note: The number of new cyclic messages to be sent must be equal
+            to the original number of messages originally specified for this
+            task.
         """
-        self.message = message
+        messages = self._check_and_convert_messages(messages)
+        messages = self._check_modified_messages(messages)
```
pierreluctg: Minor, but `self._check_modified_messages(messages)` do not make any modification to messages, so you do not really need to return `messages`

can/interfaces/socketcan/socketcan.py
```diff
@@ -599,17 +634,17 @@ def _send_once(self, data, channel=None):
...
-        The kernel's broadcast manager will be used.
+        The kernel's Broadcast Manager SocketCAN API will be used.
 
-        :param can.Message msg:
-            Message to transmit
+        :param Union[Sequence[can.Message], can.Message] messages:
+            The messages to be sent periodically
         :param float period:
-            Period in seconds between each message
+            The rate in seconds at which th send the messages.
```
christiansandberg: Typo “to”.

Issue #614, opened 2019-06-04
Unify bit timing parameters


christiansandberg:
I've had a quick look at the various interfaces and how you can specify custom bit timing parameters. They vary in abstraction. Some supports setting BTR registers directly, usually using some external tool for help. Some lets you specify SJW, TSEG1, TSEG2, SAM and then BRP is calculated from desired bitrate.

For CAN-FD you must specify different settings for arbitration phase and data phase.

Interfaces supporting BTR register:
* canalyst (named Timing0 and Timing1)
* slcan (named btr as hexadecimal string)
* pcan (possible, see #538)
* ixxat (possible)
* systec (possible)
* ...?

Interfaces supporting SJW, TSEG1, TSEG2, SAM:
* kvaser (named sjw, tseg1, tseg2, no_samp)
* vector

Interfaces supporting CAN-FD:
* kvaser (uses same sjw, tseg1, tseg2 for both arbitration and data)
* vector (sjwAbr, tseg1Abr, tseg2Abr and sjwDbr, tseg1Dbr, tseg2Dbr)
* pcan (f_clock, nom_brp, nom_tseg1, nom_tseg2, nom_sjw and data_brp, data_tseg1, data_tseg2, data_sjw)

I suggest to use the same names wherever possible. Here's a suggestion:

* BTR: btr0, btr1
* SJW, TSEG1, TSEG2, SAM: sjw, tseg1, tseg2, no_samp
* CAN-FD arbitration phase: sjw, tseg1, tseg2, no_samp
* CAN-FD data phase: data_sjw, data_tseg1, data_tseg2

bmeisels:
Considering what I learned while implementing CAN FD for Pcan I think we should have a timing parameter class.

This class will have multiple constructors from the different parameter options and supply an interface which will calculate the missing parameters based on the existing parameters. This class should be accepted by all CAN bus classes.

I think this change will allow us to simplify the interfaces and allow more easily changes to the parameters.

christiansandberg:
That would at least make the Bus argument list less cluttered. How would you handle the differences in API’s and controllers?

bmeisels:
Please correct me if I am wrong, generally it is possible to convert between different specification formats. Each interface should internally use the properties it needs from the timing parameters instance. If these parameters don't exist they should be generated based on the existing parameters.

If the previous statement is wrong and the parameters cannot be converted, an exception with an indicative error message should be thrown.

christiansandberg:
[...]

PR #615, opened 2019-06-04
Add bit timing class

christiansandberg:
A class for holding bit timing parameters. This is intended to replace the various bit timing arguments in some interfaces.

Part of #614.


can/bit_timing.py
```diff
@@ -0,0 +1,213 @@
...
+        """Time segment 2.
+
+        The number of quanta from the sampling point to the end of the bit.
+        """
+        if not self._tseg2:
+            raise ValueError("tseg2 must be specified")
+        return self._tseg2
+
+    @property
+    def nof_samples(self) -> int:
```
hardbyte: Consider just `samples`?

christiansandberg: Maybe it’s just me but that makes me think it is a list of samples rather than an integer.

I was also thinking about using a bool for using three samples or not.

PR #621, opened 2019-06-11
Use inter-process mutex to prevent concurrent neoVI device open

pierreluctg:
When neoVI server is enabled, there is an issue with concurrent device open.


felixdivo:
Maybe we could make `os.path.join(tempfile.gettempdir(), "neovi.lock")` a direct member of the module so it can be modified?

setup.py
```diff
@@ -90,6 +90,7 @@
         "wrapt~=1.10",
         "aenum",
         'windows-curses;platform_system=="Windows"',
+        "filelock",
```
christiansandberg: Should be added to neovi extra requirement as well.

pierreluctg: Done

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -11,8 +11,12 @@
 """
 
 import logging
+import os
+import tempfile
 from collections import deque
 
+from filelock import FileLock
```
christiansandberg: Btw, will the documentation be able to build now? Usually we try to make the modules importable without external packages installed.

pierreluctg: @christiansandberg I have added a dummy file lock in case `filelock` module is not there

setup.py
felixdivo: Isn't that kind of duplicated if it's here and there?

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -28,6 +30,35 @@
...
+except ImportError as ie:
+
+    logger.warning(
+        "Using ICS NeoVi can backend without the "
+        "filelock module installed may cause some issues!: %s",
+        ie,
+    )
+
+    class FileLock:
+        """Dummy file lock that do not actually do anything"""
```
felixdivo: "do not" -> "does not"

PR #622, opened 2019-06-11
Remove Travis builds of Python 3.7 & nightly on OSX

felixdivo:
It is unsupported and causes problems. See [here](https://github.com/hardbyte/python-can/pull/554#issuecomment-500453610).


felixdivo:
This should allow #554 to be merged eventually.

PR #626, opened 2019-06-19
More flexible configuration context handling and allow CAN_CONFIG env var

pierreluctg:
* Simplifying the configuration context handling
* Also allowing to use the context with environment variables
* Allowing to set other bus `config` using CAN_CONFIG/CAN_CONFIG_CONTEXT environment variable JSON
    * For example `CAN_INTERFACE=socketcan CAN_CONFIG={"receive_own_messages": true, "fd": true} python -c "import can; can.interface.Bus()"`


pierreluctg:
@felixdivo documentation updated

PR #633, opened 2019-06-29
setup.py: require pytest-runner only when necessary

texierp:
This optimizes setup.py for cases when pytest-runner is not needed,
using the approach that is suggested upstream:

https://pypi.python.org/pypi/pytest-runner#conditional-requirement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,7 @@ from os import listdir
 from os.path import isfile, join
 import re
 import logging
+import sys
 from setuptools import setup, find_packages
 
 logging.basicConfig(level=logging.WARNING)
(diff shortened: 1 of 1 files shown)


texierp:
Sure, no problem, I can update the PR accordingly ;)

Thanks !

texierp:
Updated :)

felixdivo:
Then thank you very much for your contribution! It seems like the black formatter is not yet completely convinced of the beauty of that code. If he is, we can merge this. ;)

texierp:
Doh, I will update ;)

texierp:
Seems Ok now :)

felixdivo:
@texierp Thanks for the contribution! :smile:

PR #635, opened 2019-07-06
VectorBus - add xlCanRequestChipState to vxlapi.py

zariiii9003:
- added xlCanRequestChipState
- added some constants for chipState and canErrors
- added a few descriptions from vxlapi.h
- edited some numbers and order of constants to be more consistent with vxlapi.h


zariiii9003:
> Using the vxlapi module directly is not part of the official API but you are of course free to do so if you’re ok with backwards incompatible changes in the future.

This is true. But as far as i know there is no BusState implementation for VectorBus yet (is there one for any other interface?). And i see this as a small step towards that.

felixdivo:
[This search](https://github.com/hardbyte/python-can/search?q=busstate&unscoped_q=busstate) reveals that ucan, pcan and systec (already) use `BusState`, though I haven't looked into it more deeply.

zariiii9003:
Maybe there is a misunderstanding on my side. I assumed the BusState was intended for the error state like XL_CHIPSTAT_ERROR_PASSIVE or XL_CHIPSTAT_ERROR_ACTIVE.

But for pcan it defines whether the interfaces sends an ACK and is able to transmit messages:

```
        elif new_state is BusState.PASSIVE:
            # When this mode is set, the CAN controller does not take part on active events (eg. transmit CAN messages)
            # but stays in a passive mode (CAN monitor), in which it can analyse the traffic on the CAN bus used by a
            # PCAN channel. See also the Philips Data Sheet "SJA1000 Stand-alone CAN controller".
            self.m_objPCANBasic.SetValue(
                self.m_PcanHandle, PCAN_LISTEN_ONLY, PCAN_PARAMETER_ON
            )
```

This would correnspond to XL_OUTPUT_MODE_SILENT or XL_OUTPUT_MODE_NORMAL for Vector hardware. Is there any existing or planned BusABC-method to get the error state or maybe even the error counts?

christiansandberg:
I remember it was discussed if it really was a good idea to combine two related but different things in a single property. But that’s how it is anyway.

Bus statistics is probably something to attack next but nothing’s planned.

PR #638, opened 2019-07-11
Read the BCM status before creating a Task

### LINKED ISSUE #605: Error when attempting to stop messages with same arbitration ID
This error is returned when attempting to stop messages that have the same arbitration ID
  ```
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 280, in send_bcm
    return bcm_socket.send(data)
OSError: [Errno 22] Invalid argument

During handling of the above exception, another exception occurred:

  File "simple-nm-sim.py", line 238, in <module>
    sim.force_stop()
  File "simple-nm-sim.py", line 179, in force_stop
    self.vin1_task.stop()
  File "/usr/local/lib/python3.5/dist-packages/can/bus.py", line 218, in wrapped_stop_method
    original_stop_method()
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 365, in stop
    send_bcm(self.bcm_socket, stopframe)
  File "/usr/local/lib/python3.5/dist-packages/can/interfaces/socketcan/socketcan.py", line 285, in send_bcm
    raise can.CanError(base + "You are probably referring to a non-existing frame.")
can.CanError: Couldn't send CAN BCM frame. OS Error 22: Invalid argument
You are probably referring to a non-existing frame.
```

Below is the code snippet where messages are defined and also where messages are being stopped. pm_snapshot_task is stopped first then an error is thrown when attempting to stop vin1_task as it has the same arbitration ID.
       
      
```
self.pm_snapshot_msg = can.Message(
            arbitration_id=0x400,
            data=self.pm_snapshot_data,
            extended_id=False,
            dlc=8)

        self.nm_msg = can.Message(
            arbitration_id=0x51E,
            data=self.nm_data,
            extended_id=False,
            dlc=8)

        self.vs_msg = can.Message(
            arbitration_id=0x11,
            data=self.vs_data,
            extended_id=False,
            dlc=8)

        self.vin1_msg = can.Message(
            arbitration_id=0x400,
            data=self.vin1_data,
            extended_id=False,
            dlc=8)
```
Stopping the tasks

```
if self.pm_snapshot_task is not None:
            self.pm_snapshot_task.stop()
            self.pm_snapshot_task = None

        if self.vs_task is not None:
            self.vs_task.stop()
            self.vs_task = None

        if self.vin1_task is not None:
            self.vin1_task.stop()
            self.vin1_task = None
```

### PR DESCRIPTION
**Pull request by karlding**
Currently `bus.send_periodic` blindly wraps the BCM socket API, and sends
a BCM operation with the `TX_SETUP` opcode. However, the semantics that
the kernel driver provides are an "upsert". As such, when one attempts
to create two Tasks with the same CAN Arbitration ID, it ends up
silently modifying the periodic messages, which is contrary to what our
API implies.

This fixes the problem by first sending a `TX_READ` opcode with the CAN
Arbitration ID that we wish to create. The kernel driver will return
`-EINVAL` if a periodic transmission with the given Arbitration ID does
not exist. If this is the case, then we can continue creating the Task,
otherwise we error and notify the user.

Fixes #605

### CODE REVIEW

File: can/interfaces/socketcan/socketcan.py [hunk 1]
```diff
@@ -330,6 +330,31 @@ def _tx_setup(self, messages):
...
+            ival1_usec=0,
+            ival2_seconds=0,
+            ival2_usec=0,
+            can_id=self.can_id_with_flags,
+            nframes=0,
+        )
+        try:
+            self.bcm_socket.send(check_header)
+        except OSError as e:
+            assert e.errno == errno.EINVAL
```
felixdivo: I think this should carry a comment for the user, since the default assertion error message is not very helpful for most users at first glance. For example, we might do:

```suggestion
            if e.errno != errno.EINVAL:
                raise e
```

PR #640, opened 2019-07-14
PCAN: add support for detecting available channels

jsee23:


felixdivo:
Thank you for improving this! It seems like some tests do net yet work, any the formatter will probably spit out some improvements as well. Generally, the `channels` variable can be defined right above the for loop.

can/interfaces/pcan/pcan.py
```diff
@@ -477,6 +477,31 @@ def state(self, new_state):
...
+        status = TPCANStatus(0)
+        for i in interfaces:
+            error, value = libraryHandle.GetValue(i['id'], PCAN_CHANNEL_CONDITION)
+            if error != PCAN_ERROR_OK or value != PCAN_CHANNEL_AVAILABLE:
+                continue
+            hasFd = False
+            error, value = libraryHandle.GetValue(i['id'], PCAN_CHANNEL_FEATURES)
+            if error == PCAN_ERROR_OK:
+                hasFd = bool(value & FEATURE_FD_CAPABLE)
+            channels.append({'interface': 'pcan', 'channel': i['name'], 'supportsFd': hasFd})
```
felixdivo: This should probably be like that, to be consistent with the naming scheme:

```suggestion
            channels.append({'interface': 'pcan', 'channel': i['name'], 'supports_fd': hasFd})
```

jsee23:
Thanks for your feedback, I will check the tests by this evening.

PR #641, opened 2019-07-14
vector: specify CAN FD support in config detection

jsee23:
This adds the CAN FD support in the config detection for Vector. Additionally, channels without CAN support are skipped.

dict entry is aligned with PCAN backend, see https://github.com/hardbyte/python-can/pull/640.


can/interfaces/vector/canlib.py
```diff
@@ -484,6 +486,7 @@ def _detect_available_configs():
                     "interface": "vector",
                     "app_name": None,
                     "channel": channel_config.channelIndex,
+                    'supportsFd': bool(channel_config.channelBusCapabilities & vxlapi.XL_CHANNEL_FLAG_CANFD_ISO_SUPPORT)}
```
felixdivo: This should probably be like that, to be consistent with the naming scheme:

```suggestion
                    'supports_fd': bool(channel_config.channelBusCapabilities & vxlapi.XL_CHANNEL_FLAG_CANFD_ISO_SUPPORT)}
```

zariiii9003:
does the bitshift `XL_BUS_ACTIVE_CAP_CAN = (XL_BUS_TYPE_CAN << 16)` return the same result as in C?

felixdivo:
> does the bitshift `XL_BUS_ACTIVE_CAP_CAN = (XL_BUS_TYPE_CAN << 16)` return the same result as in C? (@zariiii9003)

Yeah, I guess. (See [here](https://wiki.python.org/moin/BitManipulation))

felixdivo:
I can only merge this is the formatter approves. @jsee23 Could you fix this?

PR #643, opened 2019-07-16
Implemented error statistics and silent mode for VectorBus

zariiii9003:
Builds upon #635 .

- added ErrorState as enum in bus.py
- added variables error_state, TEC and REC to BusABC.__init__()
- added property error_statistics to BusABC which returns a tuple of the mentioned variables
'
- implemented method request_chip_state in VectorBus
- call request_chip_state once in VectorBus.__init__()
- chip_state event is received in VectorBus._recv_internal()
- request_chip_state is called after every received chip_state event
- implemented property error_statistics in VectorBus
'
- added silent mode to VectorBus based on implementation for pcan (BusState)
- function xlCanSetChannelOutput was added to vxlapi.py to enable silent mode

Please also take a look at my docstrings as i have no clue how that sphinx thing works :)


can/bus.py
```diff
@@ -56,6 +66,11 @@ def __init__(self, channel, can_filters=None, **kwargs):
         self._periodic_tasks = []
         self.set_filters(can_filters)
 
+        #: initialize Error status, transmit error count TEC, receive error count REC
+        self.error_state: ErrorState = ErrorState.ERROR_ACTIVE
+        self.TEC: int = -1
```
felixdivo: I would spell them out and document these parameters since they are "public".

can/bus.py
```diff
@@ -24,6 +25,15 @@ class BusState(Enum):
     ERROR = auto()
 
 
+class ErrorState(Enum):
+    """The fault confinement state in which CAN interface can be."""
+
+    BUS_OFF = auto()
+    ERROR_PASSIVE = auto()
```
felixdivo: Newcomers should be able to find some documentation on when these error states occur.

zariiii9003: I added a more detailed description to the property BusABC.error_statistics as i think the ErrorState class docstring will not be visible in the documentation.
 Please correct me if i'm wrong.

can/bus.py
zariiii9003: You're right. I changed it to txErrorCount and rxErrorCount.

zariiii9003:
You can test this with:
```
import can
import time
from can.interfaces.vector import VectorBus


class SomeListener(can.Listener):
        pass


bus = can.Bus(channel=0, interface="vector", fd=False)
# bus = can.Bus(channel=0, interface='vector', state=can.BusState.ACTIVE, fd=False)
# bus = VectorBus(channel=0, state=can.BusState.ACTIVE, fd=False)

listener = SomeListener()
notifier = can.Notifier(bus, [listener])

for i in range(10):
    print(bus.error_statistics)
    time.sleep(1)
```

I noticed some weird behavior btw, when i intialize the bus with `can.Bus(channel=0, interface="vector", fd=False)` it actually uses CAN FD. `self.fd` is True immediately after the `__init__()` arguments.

When i use `VectorBus(channel=0, state=can.BusState.ACTIVE, fd=False)` it works as expected.
Any idea what causes this?

PR #644, opened 2019-07-16
Add interprocess virtual bus

### LINKED ISSUE #530: Interprocess virtual interface
The current `virtual` interface is of limited use since it only distributes messages within the same process. If we could make it work with multiple processes it could be used in cases where a system of multiple processes needs to be run or tested.

It would be simple to implement if we had a central daemon that did this, but avoiding this would be elegant since a user of the library would need to set up less and there would be no central entity that could have problems.

I imagined it to be like this: Process A sends `("my_network_1", can.Message(...))` to some multicast/broadcast system and Process B does `net, msg = broadcast.read(); if net == "my_network_1": process(msg)`.

Does someone have an idea on how to implement such a multicast/broadcast, possibly in a cross-platform way? [Over here](https://wiki.python.org/moin/UdpCommunication#Multicasting.3F) in the Official Python Wiki, people said it is tricky to get right. Is that still the case? Should be use [zeroconf](https://github.com/jstasiak/python-zeroconf) instead (probably not)?

felixdivo:
This PR adds a virtual interface, that allows multiple processes to exchange can data. This is done by wrapping the CAN frames in IPv6/4 packets and sending them to a multicast group. The wrapping is done using [MessagePack](https://msgpack.org/) as proposed by @windelbouwman. That theoretically allows processes running other software than this one to interoperate with python-can, though I doubt it will used as that much. The TTL of the IP packets is set to one by default, disallowing the packet to escape the localhost. It is configurable however, and would allow the communicating processes to run on different machines as well. However this has not been tested.

Things to discuss before this can be merged:
- [x] How shall it be called? `virtual_interprocess`, `can_over_ip`? Other suggestions? -> I would opt for `can_over_ip`.
- [x] How does this relate to the current `virtual` interface? Is it a replacement, or an addition? In principle, it is at feature-rich as the current virtual interface, but might be a bit slower and requires an external library. I would suggest that it is a addition. Shall `msgpack` be added as a general required package? ->  added to the required ones as it is tiny and does not have any transitive dependencies
- [x] Do we want to provide the MessagePack wrapping of the CAN frames to other interfaces or the user as well? It might be useful beyond this interface at some point and is fairly general. -> Can still be done

Things for me to do
- [x] implement `receive_own_messages` -> raise warning if used for now
- [x]  add tests
- [x]  add docs/examples
- [x]  register as an interface

Closes #530.


windelbouwman:
This is nice, and will be handy! The multi-cast solution is handy, but multicast also is often beyond the grasping of many engineers. Maybe we could add a tcp connection based solution as well, with a central server?

felixdivo:
I do not like the central server, since it would not work out of the box by simply firing up an interface. I would propose giving some example IPv4 & IPv6 multicast addresses in the documentation so if someone needs more interfaces her/she can simply plug them in. That should be the simplest, right?

Currently I have the problem that the bus receives it's own messages, which might be desirable but should be allowed to be turned off I guess. I will first have to do some further reading I guess.

windelbouwman:
Potential issues with multicast:
- Firewall configuration, is this a special case?
- Is multicast supported on all platforms?
- Can communication over the internet, does this work with multicast? Is this a complex setup?

christiansandberg:
I have made a server based interface here if you’re interested: https://github.com/christiansandberg/python-can-remote

felixdivo:
@windelbouwman Hm, but multicast is made exactly for this. :dart: I know that there is always some hesitation when using it, but man ... It's a perfect fit from a networking perspective. Basically the use of multicast instead of unicast is not a big deal, it's like a dozen lines of additional code plus a different addressing scheme. But getting rid of a central node is really nice. I'll have a look at maybe combining multicast and unicast (and have a look at @christiansandberg's library).

felixdivo:
I think this is done, except for the docs of course. I'd lease it as a multicast-only implementation and would link to @christiansandberg 's [python-can-remote](https://github.com/christiansandberg/python-can-remote) in the docs. This implementation could later be changed, if there's need for it.

felixdivo:
Hm, so this should actually work on Travis: travis-ci/travis-ci#1253. Will investigate further.

felixdivo:
@windelbouwman I know this is old, but what do you think? Do the docs suffice to address your use case (from a year ago)?

can/interfaces/multicast_ip/bus.py
```diff
@@ -0,0 +1,338 @@
...
+    def _create_socket(self, address_family: socket.AddressFamily) -> socket.socket:
+        """Creates a new socket. This might fail and raise an exception!
+
+        :param address_family: whether this is of type `socket.AF_INET` or `socket.AF_INET6`
+        :raises OSError: if the socket could not be opened or configured correctly; in this case, it is
+                         guaranteed to be closed/cleaned up
+        """
+        # create the UDP socket
+        # this might already fail but then there is nothing to clean up
+        sock = socket.socket(address_family, socket.SOCK_DGRAM)
```
windelbouwman, line 199: Using UDP causes fire and forget. Possible issues here:
- How to know when a CAN message is transmitted on the "bus" successfully?
- How to know when a CAN message is lost? In hardware CAN controllers this is not possible, a message is either transmitted on the bus, or in the queue.

can/interfaces/multicast_ip/bus.py
```diff
@@ -0,0 +1,338 @@
...
+
+        :param timeout: the timeout in seconds after which an Exception is raised is sending has failed
+        :param data: the data to be sent
+        :param start: where to start the slice to be sent
+        :param count: how many bytes to send or `None` to send everything
+        :raises OSError: if an error occurred while writing to the underlying socket
+        :raises socket.timeout: if the timeout ran out before sending was completed (this is a subclass of
+                                *OSError*)
+        """
+        if not count:
```
windelbouwman, line 265: A strategy I've been applying a lot lately is "inverse the if not". In this case, I would change `if not count` into `if count` and swap the arms of the if. Non-negative logic is often easier to read.

doc/interfaces/multicast_ip.rst
```diff
@@ -0,0 +1,57 @@
...
+This module implements transport of CAN and CAN FD messages over UDP on Multicast IPv4 and IPv6.
+This virtual interface allows to communicate between multiple processes and even hosts.
+This differentiates it from the :ref:`virtual_interface_doc` interface,
+which can only passes messages within a single process but does not require a network stack.
+It runs on UDP to have the lowest possible latency (as opposed to using TCP).
+
+.. note::
+    There are cases in which multicast IP is not an option, e.g. because it is unsupported or
+    blocked intermediate by routers. In this case you can also consider using the related tool
+    `python-can-remote <https://github.com/christiansandberg/python-can-remote>`_.
```
windelbouwman, line 15: I also made an attempt at virtualcan, it can be found here: https://github.com/windelbouwman/virtualcan

It also contains a python-can entry point. Maybe it can be mentioned here as well?

can/interfaces/multicast_ip/bus.py
felixdivo: You are right, this is potentially problematic. On the other hand, Multicast + normal TCP seems to be a weird couple too, as basic multicast is inherently unreliable. We don't even know the recipients of our messages. I'd leave it as is, and maybe put a bigger disclaimer in the docs? In practice and in many local/small networks, UDP it's definitely reliable enough for this. And on a loopback/link-local network too. And that's one major area of application.

doc/interfaces/multicast_ip.rst
```diff
@@ -0,0 +1,57 @@
+.. _multicast_ip_doc:
+
+Multicast IP Interface
+======================
+
+This module implements transport of CAN and CAN FD messages over UDP on Multicast IPv4 and IPv6.
```
windelbouwman, line 6: I agree that multicast is very handy, but I have one question about this:

- How is CAN arbitration handled? Is there any specific order in which CAN messages are received by all participating CAN "controllers" on the multicast IP bus? For example, in a real life CAN bus, CAN messages have an arbitration function, so CAN messages with a lower ID have higher priority. Also, on real hardware CAN messages are received in the same order at all CAN controllers. This might seem details, so I'm not sure how important this is for simulation purpose. But it might trigger some unexpected situations when higher level protocols make some assumptions about this.

As a side note: I created [virtualcan](https://github.com/windelbouwman/virtualcan), in which CAN bus arbitration is not yet implemented, but since this project does take the central server approach, it is possible to add priority sorting, and messages will arrive at all ends in the same order.

It might be good to document how the priority / ordering of CAN message is handled so that people know what to expect.

can/interfaces/multicast_ip/bus.py
felixdivo: Yep. Actually, the `start` and `count` parameters were never really used, so I just deleted that option. Now its simpler too. :P

doc/interfaces/multicast_ip.rst
felixdivo: Sure, if it helps the users. Could you prepare the wording for it, in particular how it differs from the other virtual interfaces?

windelbouwman:
I made some comments across the PR, overall I believe this looks pretty good, and I'm happy with the work towards a networked virtualcan. Nice job! :+1:

doc/interfaces/multicast_ip.rst
felixdivo: Sure, at the very least this should be documented. I actually didn't think about that until right now. Huh.

But the implementation simply receives whatever is coming in that order, and does not care about overlapping timestamps or something like that. It does not entirely faithfully reproduce a physical network, that's correct. But if the bus isn't near capacity, that shouldn't be a problem, right?

This interface would have to implement throttling too in order to be close to realistic. Right now, we could theoretically send Gigabits of traffic over that "CAN" bus.

I'd say it's okay if the virtual interfaces are idealized in that way. The `python-can` *virtual* interface does not care either (and we should state that in the docs too).

can/interfaces/multicast_ip/bus.py
windelbouwman: Agreed to put a disclaimer for this in the docs. Also, I'm not sure how often this would happen, and if so, if it poses an issue. Maybe it does pose and issue, since the CAN bus is expected to be reliable, so sending a CAN message does not deal with re-transmits and all this stuff. You either send a message, or you do not. This is a good feature of the CAN protocol.

doc/interfaces/multicast_ip.rst
windelbouwman: Notable features:
- TCP/IP connections to a central central server approach
- custom CAN message serialization
- implementations in Python, C++ and rust (for cross language interoperability)
- a server to physical can bus written in rust (for speed reasons)

The reason why I created yet another binary serialization of a CAN message is that I could not get MessagePack to build easily on C++ with windows. This is a bit lame, but the C++ code now is fairly simple, and has few dependencies.

I'm not fully aware how the other virtual interfaces function, but I'm guessing it's more of the same?

doc/interfaces/multicast_ip.rst
windelbouwman: > But if the bus isn't near capacity, that shouldn't be a problem, right?

Let's hope so :wink: . This should be documented at the very least indeed. The thing with these kind of assumptions is that in 99.99% of the cases this poses no issue, and once a month, due to a glitch there is some weird failure. This might be troublesome to pinpoint.. I'd say, put the limitations in the docs, and maybe even issue a warning during opening of the interface?

> This interface would have to implement throttling too in order to be close to realistic. Right now, we could theoretically send Gigabits of traffic over that "CAN" bus.

Throttling would be a nice feature to add indeed! I'll keep it in mind.

> The `python-can` _virtual_ interface

The term virtualcan is a bit overloaded. Thanks for naming this plugin `virtualcan_multicast`!

doc/interfaces/multicast_ip.rst
felixdivo: That should suffice. Its somewhat different, the included one is internally just a global list that relays it within the same process. Dead simple. ;)

doc/interfaces/multicast_ip.rst
felixdivo: > The term virtualcan is a bit overloaded. 

Yeah. ;) It's actually `multicast_ip` to specify the carrying layer. Would you say `udp_multicast` would be better?

felixdivo:
I completed the documentation (If you don't have any more comments) and included this table:

Screenshot from 2020-11-22 13-48-44

I figured it is easier to read as an image than in source code.

can/interfaces/multicast_ip/bus.py
```diff
@@ -0,0 +1,327 @@
...
+    Both default addresses should allow for multi-host CAN networks in a normal local area network (LAN) where
+    multicast is enabled.
+
+    .. note::
+        The auto-detection of available interfaces (see) is implemented using heuristic that checks if the
+        required socket operations are available. It then returns two configurations, one based on
+        the :attr:`~MulticastIpBus.DEFAULT_GROUP_IPv6` address and another one based on
+        the :attr:`~MulticastIpBus.DEFAULT_GROUP_IPv4` address.
+
+    .. warning::
```
windelbouwman, line 45: It might be good to document here the known limitations:
- no CAN id prioritization
- no guarantees on delivery / order of delivery
- no rate limiting on messages

windelbouwman:
The table with comparison looks good!

felixdivo:
I think we should name it `udp_multicast` instead of `multicast_ip` as that is more descriptive of what it can achieve and what not.

felixdivo:
@hardbyte There seems to be [this](https://github.com/hardbyte/python-can/pull/644/checks?check_run_id=1456261689#step:6:245) error when the formatter in the cloud has to step in.

felixdivo:
I renamed to bus to `udp_multicast` and added some more tests to improve coverage a bit.

felixdivo:
Hm, I don't quite get why the black formatter fails on `pull_request` but succeds on `push` (as it should).

felixdivo:
Else, this PR is ready for a final review.

hardbyte (approved):
🤸🏼  nice! Everything looks good to merge from my perspective.

I've pushed a few minor tweaks to the docs - hopefully nothing controversial.

felixdivo:
Thanks Brian!

And the changes are not controversial. ;)

felixdivo:
Shall we merge (via squashing) or do we need another review? I thinks it's sufficient as is. @windelbouwman also had a look at it.

hardbyte:
Great effort @felixdivo. Let's push out another pre-release of 4.0 with this feature

PR #646, opened 2019-07-17
add test for vector interface

zariiii9003:
Add test for VectorBus to improve test coverage.


felixdivo:
It is executed as well: https://travis-ci.org/hardbyte/python-can/jobs/560133532#L453

test/test_vector.py
```diff
@@ -0,0 +1,158 @@
...
+        can.interfaces.vector.canlib.vxlapi.xlOpenPort.assert_called()
+        can.interfaces.vector.canlib.vxlapi.xlCanFdSetConfiguration.assert_not_called()
+        can.interfaces.vector.canlib.vxlapi.xlCanSetChannelBitrate.assert_called()
+
+    def test_bus_creation_fd(self) -> None:
+        self.bus = can.Bus(channel=0, bustype="vector", fd=True)
+        self.assertIsInstance(self.bus, canlib.VectorBus)
+        can.interfaces.vector.canlib.vxlapi.xlOpenDriver.assert_called()
+        can.interfaces.vector.canlib.vxlapi.xlGetApplConfig.assert_called()
+        can.interfaces.vector.canlib.vxlapi.xlOpenPort.assert_called()
```
karlding: Looking through the code, it seems like in certain cases, the same Vector API is called, but with a different parameter to distinguish between various cases (eg. for FD vs. non-FD). Like here, the difference is:

```python
        if fd:
            vxlapi.xlOpenPort(
                self.port_handle,
                self._app_name,
                self.mask,
                permission_mask,
                rx_queue_size,
                vxlapi.XL_INTERFACE_VERSION_V4, # <--
                vxlapi.XL_BUS_TYPE_CAN,         # <--
            )
        else:
            vxlapi.xlOpenPort(
                self.port_handle,
                self._app_name,
                self.mask,
                permission_mask,
                rx_queue_size,
                vxlapi.XL_INTERFACE_VERSION,    # <--
                vxlapi.XL_BUS_TYPE_CAN,         # <--
            )
```

Does it make sense to also assert on those arguments for those situations, rather than just that the function was called?

karlding:
In general, I'm not a _huge_ fan of extensively mocking, since it leaks details about the implementation, which makes tests brittle and code harder to refactor in the future.

However, until we get hardware testing or something, I don't think it's really feasible to do without, and so perhaps this serves as enough of a safeguard to make changes. With that being said, I'm okay with these, since I could see how this would improve confidence in changing code.

test/test_vector.py
zariiii9003: done :)

zariiii9003:
I do not like mocking this much either, but i think it is acceptable for this use case. I only mock the vxlapi.dll-functions and i am sure they won't change anytime soon.

So i rebuilt this a little bit. I catched the ctypes.windll exception in the vxlapi.py so i can ruse the ctype structures. I also added side_effect functions that mimic the vxlapi function behavior.

It probably makes sense to add more assertions for the sending and receiving functions, but i'll probably try that after my vacation is over. Then i can look at some real data.

test/test_vector.py
```diff
@@ -154,5 +215,156 @@ def test_reset(self):
         can.interfaces.vector.canlib.vxlapi.xlActivateChannel.assert_called()
 
 
+def xlGetApplConfig(
```
karlding: Ah, you're splitting these out because all these modify the reference and have a return value? Thus in order to return those values from the function, a `side_effect` function is used? That seems to be the main rationale for this change from my understanding.

This seems reasonable to do so like what you've done, and I suppose that any tests which need different behaviour can patch it as necessary.

However, I'm just wondering whether providing this as a default implementation is any different vs what you had before, since any time we want to exercise other cases the function will probably be patched in a way specific to the test case. @felixdivo do you have any thoughts?

zariiii9003: Some vxlapi functions modify the arguments (mutable ctypes) that are given to the function. That's why i added these side_effects. 
I think this is better than 'dumb' mocks.

When we add new functions to the vxlapi these tests will break as no mock exists for the new function. But i don't see this as a huge problem because it encourages looking at the tests. Solving it will be pretty simple too.

zariiii9003:
I tried something different: 

https://github.com/zariiii9003/python-can/tree/test_VectorBus_restructuring/can/interfaces/vector

I split the vxlapi.py into 
- XLDefine (enums of constants)
- XLClass (data types and structures)
- XLDriver (vxlapi functions).

That is the structure that vector uses for its .NET-Wrapper. I also copied the enum class names from that .NET wrapper. i think it really improves the readability of our wrapper. 
It also enhances the readability of the canlib.py itself because you can see now whether the object is a constant, a data type or a function.
It also shows which function those cryptic constant names belong to.

With this structure i can import the XLDefine.py and the XLClass.py into my test_vector.py without worrying about the ctypes.windll exception.

But i actually did this because this way i can add `can.interfaces.vector.canlib.XLDriver = Mock()` to my test_vector.py which makes the test more robust against new XLDriver functions being called in the canlib.py.

What do you think about this?

test/test_vector.py
felixdivo: I haven't done too much with mocking, but this seems reasonable the way it is now.

christiansandberg:
Sounds good to me. I just think we should strive to use PEP-8 module naming conventions like the rest of the files.

karlding:
Thanks for making this split @zariiii9003!

PR #650, opened 2019-07-20
Add BitTiming class compatibility to Vector interface

zariiii9003:
Implementation based on #625 

BitTiming class takes precedence over the bitrate argument.
I did not remove the other bit timing arguments from __init__() and i did not add any deprecation warnings.

i tested this on a VN1610


can/interfaces/vector/canlib.py
```diff
@@ -217,20 +224,36 @@ def __init__(
...
+                    self.canFdConf.arbitrationBitRate = ctypes.c_uint(timing.bitrate)
+                    self.canFdConf.sjwAbr = ctypes.c_uint(timing.sjw)
+                    self.canFdConf.tseg1Abr = ctypes.c_uint(timing.tseg1)
+                    self.canFdConf.tseg2Abr = ctypes.c_uint(timing.tseg2)
+                    if data_timing:
+                        self.canFdConf.dataBitRate = ctypes.c_uint(data_timing.bitrate)
+                        self.canFdConf.sjwDbr = ctypes.c_uint(data_timing.sjw)
+                        self.canFdConf.tseg1Dbr = ctypes.c_uint(data_timing.tseg1)
+                        self.canFdConf.tseg2Dbr = ctypes.c_uint(data_timing.tseg2)
+                    else:
```
hardbyte: This `else` clause seems unnecessary - instead why don't we make sure  `data_timing` falls back to the `timing` object

can/interfaces/vector/canlib.py
```diff
@@ -217,20 +224,36 @@ def __init__(
         if permission_mask.value == self.mask:
             if fd:
                 self.canFdConf = xlclass.XLcanFdConf()
-                if bitrate:
-                    self.canFdConf.arbitrationBitRate = ctypes.c_uint(bitrate)
+                if timing:
```
hardbyte: Can we refactor this out of the `init` and get back the `canFdConf` object? That should make the code more readible and make testing the corner cases a bit easier.

zariiii9003:
@hardbyte Thank you for your review! I will take care of it once the discussion in #625 is resolved.

can/interfaces/vector/canlib.py
zariiii9003: @hardbyte can you take another look? I implemented the changes you requested.

PR #652, opened 2019-07-23
Add typing annotations for functions in can.bus

karlding:
This adds typing annotations for use via mypy for all functions under
can.bus.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -4,7 +4,7 @@
 Contains the ABC bus implementation and its documentation.
 """
 
-from typing import Any, Dict, Iterable, Iterator, List, Optional, Sequence, Union
+from typing import Any, Iterable, Iterator, List, Optional, Sequence, Union
 
 import can.typechecking
 
--- a/can/bus.py
+++ b/can/bus.py
@@ -4,7 +4,7 @@
 Contains the ABC bus implementation and its documentation.
 """
 
-from typing import Any, Iterator, List, Optional, Sequence, Union
+from typing import Any, Iterator, List, Optional, Sequence, Tuple, Union
 
 import can.typechecking
 
@@ -56,7 +56,7 @@ class BusABC(metaclass=ABCMeta):
         :param channel:
             The can interface identifier. Expected type is backend dependent.
 
-        :param list can_filters:
+        :param can_filters:
             See :meth:`~can.BusABC.set_filters` for details.
 
         :param dict kwargs:
@@ -71,11 +71,9 @@ class BusABC(metaclass=ABCMeta):
     def recv(self, timeout: Optional[float] = None) -> Optional[can.Message]:
         """Block waiting for a message from the Bus.
 
-        :type timeout: float or None
         :param timeout:
             seconds to wait for a message or None to wait indefinitely
 
-        :rtype: can.Message or None
         :return:
             None on timeout or a :class:`can.Message` object.
         :raises can.CanError:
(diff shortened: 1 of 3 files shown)


karlding:
I wasn't sure how to handle annotating `stop_all_periodic_tasks`, since we wrap the original `stop` class method and then add support for the `remove_tasks` argument, so each implementation doesn't need to provide it. Thus, `wrapped_stop_method` is called first in the call stack before the original `stop` method.

Adding annotations to `stop_all_periodic_tasks` would result in that function being checked by `mypy`, and I wasn't sure how to resolve the following:

```
can/bus.py:291: error: Unexpected keyword argument "remove_task" for "stop" of "CyclicTask"
```

As such, I've skipped it for now, but any ideas or insights would be appreciated!

hardbyte:
@karlding given the change in #637 I think we can just remove the `remove_task=False` from `stop_all_periodic_tasks` altogether?

can/typechecking.py
```diff
@@ -2,5 +2,9 @@
 """
 import typing
 
-CanFilter = typing.Dict[str, typing.Union[bool, int, str]]
+import mypy_extensions
+
+CanFilter = mypy_extensions.TypedDict(
+    "CanFilter", {"can_id": int, "can_mask": int, "extended": bool}
```
karlding: Hmm, so looking at the CI failure, it seems like in order to take advantage of `TypedDict` we would either need to wait until it's included in `typing`, or add a runtime dependency on `mypy_extensions`?

hardbyte: Oh that is much nicer than the `Union` of possible values. I'm all for adding the dependency.

can/bus.py
```diff
@@ -164,7 +173,13 @@ def send(self, msg, timeout=None):
         """
         raise NotImplementedError("Trying to write to a readonly bus?")
 
-    def send_periodic(self, msgs, period, duration=None, store_task=True):
+    def send_periodic(
+        self,
+        msgs: Union[Sequence[can.Message], can.Message],
```
hardbyte: Would you mind removing the duplicate type signature from the doc string as you add them as annotations? 

We should be able to ensure the docs still provide all the info with https://github.com/agronholm/sphinx-autodoc-typehints

can/bus.py
```diff
@@ -223,14 +238,19 @@ def wrapped_stop_method(remove_task=True):
                     pass
             original_stop_method()
 
-        task.stop = wrapped_stop_method
+        setattr(task, "stop", wrapped_stop_method)
```
hardbyte: Interesting... was this to satisfy mypy?

can/bus.py
karlding: Yeah, I had that working, but I didn't want to add too much to this PR. I'll create a separate one that adds the new dependency.

can/bus.py
karlding: Yeah, I was getting the following:

```
can/bus.py:241: error: Cannot assign to a method
```

It appears that mypy doesn't [really](https://github.com/python/mypy/issues/2427) have good [support](https://github.com/python/mypy/issues/708) for this sort of patching, so the workaround is to use `setattr` or `type: ignore`.

felixdivo:
> @karlding given the change in #637 I think we can just remove the `remove_task=False` from `stop_all_periodic_tasks` altogether?

But it you then called `task.stop()` directly, it wouldn't be removed from the bus, right?

can/bus.py
felixdivo: Great work, man!

can/typechecking.py
felixdivo: I would opt for `mypy_extensions`, since it will be added to the standard library at some point an can the be removed/it is just an intermediate solution.

PR #655, opened 2019-07-24
Extend Ixxat Driver CyclicSendTask with base ModifiableCyclicTaskABC to have modify_data() available

dmothes:
Hi, 

The Python CanOpenlib uses the self._node.rpdo[1]['Name'].raw and calls the a .update() function to call modfiy_data() if driver has the attribute. The IXXAT driver didn't have the attribute for now. But if u don't stop() and .start() in between u can't update the data. Now the modify_data() function derived from ModifiableCyclicTaskABC takes this in account and deletes the IXXAT hardware message on the CAN2USB and creates a new one without task recreation. 

Regards

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,7 @@ sdist/
 var/
 *.egg-info/
 .installed.cfg
-*.egg
+# *.egg
 
 # PyInstaller
 #  Usually these files are written by a python script from a template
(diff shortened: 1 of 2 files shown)


christiansandberg:
Actually this is intentional. At least in my opinion a modifiable task should not affect the timing when modified. If you need to stop and start then it is not truly modifiable.

We might however question why we have this class. Maybe applications should always be able to call modify_data() and if the interface does not implement this then we should fall back to starting and stopping. Maybe also checking if the data has not changed so that timing is kept constant as much as possible.

felixdivo:
I think we should discuss this in an issue first (if there is need for it) before attempting an implementation.

PR #656, opened 2019-07-24
Make ThreadBasedCyclicSendTask event based

zariiii9003:
Since `time.sleep()`  is too inaccurate for automotive applications i suggest we make the cyclic task event based.  Maybe we should also add pywin32 to the requirements.

I also replaced `time.time()` with `time.perf_counter()` since it is platform independent.


felixdivo:
This looks like a nice addition. But can we somehow test this as well?

zariiii9003:
I'll take a look on the weekend

zariiii9003:
I added pywin32 to appveyor, now the new code lines are covered by the existing tests. Would that be sufficient?

felixdivo:
That looks good!

felixdivo:
@hardbyte is the dependency on `pywin32` handled correctly? If yes, this can be merged.

zariiii9003:
Alternatively we could add it as a requirement to `setup.py` 

```    
install_requires=[
   "wrapt~=1.10",
   "aenum",
   "filelock",
   "mypy_extensions >= 0.4.0, < 0.5.0",
   'pywin32;platform_system=="Windows"',
 ],
```

hardbyte:
@zariiii9003 once you add the dependency to `setup.py` this is okay by me.

PR #659, opened 2019-07-29
PCAN: Do not incorrectly reset CANMsg.MSGTYPE on remote frame.

craffert0:
In `PcanBus.send()`, we initially set `msgType` based on all the flags of `msg`, including RTR. In the if/else for `self.fd`, we are incorrectly resetting it if rtr. We should not, and so we are no longer doing it.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -426,9 +426,7 @@ class PcanBus(BusABC):
             CANMsg.MSGTYPE = msgType
 
             # if a remote frame will be sent, data bytes are not important.
-            if msg.is_remote_frame:
-                CANMsg.MSGTYPE = msgType.value | PCAN_MESSAGE_RTR.value
-            else:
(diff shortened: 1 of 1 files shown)


craffert0:
Note that the fix was needed because the bug was in code that wasn't covered in testing. Specifically, testing only uses fd-style can connections, and this is when it's NOT an fd.

That's why `codecov/patch` failed -- there's no testing.

hardbyte:
Thanks @craffert0 - and welcome to python-can.

Any additional testing for PCAN would be most welcome if you have the inclination :-)

schultetwin:
Can we backport this to the 3.x releases? This bug effects:
3.2.0
3.2.0-alpha
3.2.0-alpha.1
3.2.0-alpha.2
3.2.0-beta.0
3.2.1
3.2.1-alpha.0
3.2.1-alpha.1
3.2.1-alpha.2
3.3.0
3.3.1

This will fix #681

PR #664, opened 2019-08-01
Add typing annotations for can.message

karlding:
This adds typing annotations for functions in can.message. Currently the
typing annotation for the Message data is incomplete due to upstream not
implementing the buffer protocol. Since can.Message attempts to cast the
data to a bytearray via bytearray(), the Message data can be any type
that bytearray() supports.

signature.

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -8,6 +8,9 @@ This module contains the implementation of :class:`can.Message`.
     starting with Python 3.7.
 """
 
+from typing import Optional, Union
+
+from . import typechecking
 
 from copy import deepcopy
 from math import isinf, isnan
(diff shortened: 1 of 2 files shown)


can/message.py
```diff
@@ -202,7 +205,7 @@ def __copy__(self):
         )
         return new
 
-    def __deepcopy__(self, memo):
+    def __deepcopy__(self, memo: dict) -> "Message":
```
felixdivo: What does the annotation ` -> "Message"` do?

karlding: It's a [forward reference](https://mypy.readthedocs.io/en/latest/kinds_of_types.html#class-name-forward-references) for the return type.

You can check out [PEP 563](https://www.python.org/dev/peps/pep-0563/) for more about the proposed Python 4.0 change to address why this is needed.

PR #665, opened 2019-08-03
Add RedirectReader to the generated Listener docs

karlding:
This makes Sphinx aware of the RedirectReader class, which allows
CAN Messages to be gatewayed from one Bus to another Bus.

### DIFF
--- a/doc/listeners.rst
+++ b/doc/listeners.rst
@@ -41,6 +41,13 @@ BufferedReader
     :members:
 
 
+RedirectReader
+--------------
+
+.. autoclass:: can.RedirectReader
+    :members:
+
+
 Logger
 ------
 


karlding:
I think this was missed when the class was originally introduced? I don't believe there's any reason to exclude this, but do correct me if I'm wrong!

karlding:
Also, on a slightly unrelated note, it seems like the ReadTheDocs site is missing the docs for versions following `3.1.1`.. :thinking: Should `3.2.0`, `3.2.1`, `3.3.0` and `3.3.1` be available there as well?

felixdivo:
Yeah, I think it should be there as well. Here might be some more info (see other tabs as well): https://readthedocs.org/projects/python-can/versions/

hardbyte:
Hmm the GitHub -> readthedocs integration had stopped or been removed. I've fixed that now :-)

felixdivo:
I opened #666 :P

PR #673, opened 2019-08-06
Add typing to the generic IO module

felixdivo:
I added typing to:
- `can/io/generic.py`
- `can/io/logger.py`
- `can/io/player.py`

`Printer` now has a append mode. About 4 lines were changed and it made the `Logger` more uniform.

The API of `Logger` and `Player` was changed slightly.

I introduced new types in `can/typechecking.py`.

### DIFF
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -7,6 +7,7 @@ Contains a generic class for file IO.
 from abc import ABCMeta
 from typing import Optional, cast
 
+import can
 import can.typechecking
 
 
@@ -48,3 +49,15 @@ class BaseIOHandler(metaclass=ABCMeta):
         if self.file is not None:
             # this also implies a flush()
             self.file.close()
+
+
+# pylint: disable=abstract-method,too-few-public-methods
+class MessageWriter(
+    BaseIOHandler, can.Listener, metaclass=ABCMeta
+):
+    """The base class for all writers."""
+
+
+# pylint: disable=too-few-public-methods
+class MessageReader(BaseIOHandler, metaclass=ABCMeta):
+    """The base class for all readers."""
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -33,7 +33,7 @@ class BaseIOHandler(metaclass=ABCMeta):
             self.file = cast(Optional[can.typechecking.FileLike], file)
         else:
             # file is some path-like object
-            self.file = open(cast(can.typechecking.PathLike, file), mode)
+            self.file = open(cast(can.typechecking.StringPathLike, file), mode)
 
         # for multiple inheritance
         super().__init__()
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -52,9 +52,7 @@ class BaseIOHandler(metaclass=ABCMeta):
 
 
 # pylint: disable=abstract-method,too-few-public-methods
-class MessageWriter(
-    BaseIOHandler, can.Listener, metaclass=ABCMeta
-):
+class MessageWriter(BaseIOHandler, can.Listener, metaclass=ABCMeta):
     """The base class for all writers."""
 
 
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -7,7 +7,7 @@ Contains a generic class for file IO.
 from abc import ABCMeta
 from typing import Optional, Union, cast
 
-from can.typechecking import FileLike, PathLike
+from can.typechecking import FileLike, PathLike, AcceptedIOType
 
 
 class BaseIOHandler(metaclass=ABCMeta):
@@ -20,7 +20,7 @@ class BaseIOHandler(metaclass=ABCMeta):
         was opened
     """
 
-    def __init__(self, file: Optional[Union[FileLike, PathLike]], mode: str = "rt"):
+    def __init__(self, file: AcceptedIOType, mode: str = "rt"):
         """
         :param file: a path-like object to open a file, a file-like object
                      to be used as a file or `None` to not use a file at all
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -5,6 +5,9 @@ Contains a generic class for file IO.
 """
 
 from abc import ABCMeta
+from typing import Optional, Union, cast
+
+from can.typechecking import FileLike, PathLike
 
 
 class BaseIOHandler(metaclass=ABCMeta):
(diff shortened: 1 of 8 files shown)


felixdivo:
This needs a few other changes in order to be useful.

felixdivo:
Unfortunately this had to touch quite a few files. The rest of the `can/io/` will come later in order to not make this and larger.

can/io/generic.py
```diff
@@ -5,42 +5,57 @@
...
                      to be used as a file or `None` to not use a file at all
-        :param str mode: the mode that should be used to open the file, see
-                         :func:`open`, ignored if *file* is `None`
+        :param mode: the mode that should be used to open the file, see
+                     :func:`open`, ignored if *file* is `None`
         """
         if file is None or (hasattr(file, "read") and hasattr(file, "write")):
             # file is None or some file-like object
-            self.file = file
+            self.file = cast(Optional[can.typechecking.FileLike], file)
```
karlding: Can we can maybe refactor this to use `isinstance` instead and avoid the explicit `cast`? A series of `if...elif...` `isinstance` checks is completely type-checked, while a `cast` simply silences the type-checker.

can/io/player.py
```diff
@@ -6,8 +6,11 @@
 in the recorded order an time intervals.
 """
 
+import pathlib
 from time import time, sleep
-import logging
+import typing
+
+import can  # pylint: disable=unused-import
```
karlding: If this is for type-checking only, you can do something like:

```python
if TYPE_CHECKING:
    import can
```

karlding: If this is for type-checking only, you can do something like:

```python
if typing.TYPE_CHECKING:
    import can
```

can/typechecking.py
```diff
@@ -1,7 +1,10 @@
 """Types for mypy type-checking
 """
+
 import typing
 
+import os  # pylint: disable=unused-import
```
karlding: Same thing here, you can do:

```python
if typing.TYPE_CHECKING:
    import os
```

can/typechecking.py
```diff
@@ -18,3 +21,8 @@
 
 # Used for the Abstract Base Class
 Channel = typing.Union[int, str]
+
+# Used by the IO module
+FileLike = typing.IO[typing.Any]
+PathLike = typing.Union[str, "os.PathLike[str]"]
```
karlding: Technically, I believe `open` supports `str`, `bytes`, something that implements the `os.PathLike` interface, or an `int` corresponding to the file descriptor of the file. Are we just planning on ignoring `bytes` and `int` representations?

test/listener_test.py
```diff
@@ -137,13 +135,16 @@ def test_filetype_to_instance(extension, klass):
...
         test_filetype_to_instance(".txt", can.Printer)
 
-        # test file extensions that should use a fallback
-        test_filetype_to_instance("", can.Printer)
-        test_filetype_to_instance(".", can.Printer)
-        test_filetype_to_instance(".some_unknown_extention_42", can.Printer)
         with can.Logger(None) as logger:
             self.assertIsInstance(logger, can.Printer)
 
+        # test file extensions that should use a fallback
```
karlding: Can you update/remove this comment since we're changing the default behaviour? With your changes, there is no longer a fallback to the `Printer`, and instead we just raise the `ValueError` from the `Logger`'s `__init__` function.

can/io/player.py
```diff
@@ -45,45 +46,51 @@ class LogReader(BaseIOHandler):
     """
 
     @staticmethod
-    def __new__(cls, filename, *args, **kwargs):
+    def __new__(cls, filename: "can.typechecking.PathLike", *args, **kwargs):
         """
-        :param str filename: the filename/path the file to read from
+        :param filename: the filename/path the file to read from
```
karlding: Nit: Unrelated to your changes, but I think we're missing a word here in the docstring?

```suggestion
        :param filename: the filename/path of the file to read from
```

can/io/logger.py
```diff
@@ -28,36 +29,42 @@ class Logger(BaseIOHandler, Listener):  # pylint: disable=abstract-method
...
-                return CanutilsLogWriter(filename, *args, **kwargs)
-
-        # else:
-        log.warning('unknown file type "%s", falling pack to can.Printer', filename)
-        return Printer(filename, *args, **kwargs)
+        if filename is None:
+            return Printer(*args, **kwargs)
+
+        suffix = pathlib.PurePath(filename).suffix
+        if suffix == ".asc":
```
karlding: You can switch this to a table-driven method, which I think is cleaner and allows you to skip disabling the linter. Something like:

```python
        lookup = {
            ".asc": ASCWriter,
            ".blf": BLFWriter,
            ".csv": CSVWriter,
            ".db": SqliteWriter,
            ".log": CanutilsLogWriter,
            ".txt": Printer,
        }

        try:
            return lookup[suffix](filename, *args, **kwargs)
        except KeyError:
            pass

```

can/io/player.py
felixdivo: I was not able to do that because then sphinx complained about not knowing some targets (i.e. classes).

test/listener_test.py
felixdivo: Sure, must've forgotten that one.

can/io/generic.py
felixdivo: I didn't want to do that since the point of a file-*like* object is that it may reside in memory or whatever, and thus not inherit from the typical file class. Keeping duck-typing alive in this case seems more important for me. But if you can do that with `isinstance` as well somehow, we can surely change it.

can/typechecking.py
felixdivo: Yeah, that was roughly my thought as well. But I now think we should be able to support `byte` and `int` as well, since `PathLike` needs to implement `__fs_path__` (or something that) anyways, so we will always have a path to compare with.

can/io/logger.py
felixdivo: Sounds good, I'll have a look into this. :smile:

can/typechecking.py
felixdivo: You were right, only `str` works.

can/io/player.py
felixdivo: Okay, it worked.

can/io/logger.py
```diff
@@ -28,36 +29,40 @@ class Logger(BaseIOHandler, Listener):  # pylint: disable=abstract-method
...
+            ".csv": CSVWriter,
+            ".db": SqliteWriter,
+            ".log": CanutilsLogWriter,
+            ".txt": Printer,
+        }
+        suffix = pathlib.PurePath(filename).suffix
+        try:
+            return lookup[suffix](filename, *args, **kwargs)
+        except KeyError:
+            raise ValueError(f'No write support for this unknown log format "{suffix}"')
```
karlding: If we `raise` here, the previous exception (`KeyError`) is shown in the traceback. Do you think that's confusing for the user? Same for the case in `player.py`

We could potentially use the `raise from ...` syntax, or just put the `raise` in it's own block after the `try... except`.

felixdivo: I guess we should use `raise ValueError("...") from None`.

PR #674, opened 2019-08-07
Add typing annotations for can.util

karlding:
This adds typing annotations for functions in can.util.

signature.


can/typechecking.py
```diff
@@ -18,3 +23,18 @@
 
 # Used for the Abstract Base Class
 Channel = typing.Union[int, str]
+
+BusConfig = mypy_extensions.TypedDict(
```
karlding: So not really sure what the best way to go about this is.. On one hand, you get some additional type inference.

However, the problem with using a `TypedDict` is that it doesn't allow for optional types. `total=False` only allows omitting keys, and does not allow extra keys, so if various backends introduce their own config keys, they'd need to be specified here. This could potentially get really long.

felixdivo: I'm unsure about this too... The thing is that the "dispatcher class" Bus is else throwing the type information of the specific interfaces away right?

hardbyte: Requiring backends to add all optional config keys here seems to be a recipe for disaster. 

Annotating as a `NewType("BusConfig", dict)` might be a good comprise? It may offer a small advantage over a simple type alias to `Dict[str, any]` - only `dicts` created using a function annotated as returning a `typechecking.BusConfig` would pass mypy.

karlding:
Your proposed suggestion sounds reasonable to me!

Sorry for not responding earlier, I'm out of town until sometime next week, so I'll address that change once I get back.

felixdivo:
I guess this can be merged?

PR #678, opened 2019-08-08
Fix double codecov comments

felixdivo:


felixdivo:
I'm just fiddling around and will post a new & clean PR when it works.

PR #679, opened 2019-08-08
Add typing annotations for can.notifier

karlding:
This adds typing annotations for functions in can.notifier.

signature.

### DIFF
--- a/can/notifier.py
+++ b/can/notifier.py
@@ -4,6 +4,12 @@
 This module contains the implementation of :class:`~can.Notifier`.
 """
 
+from typing import Iterable, List, Optional, Union
+
+from can.bus import BusABC
+from can.listener import Listener
+from can.message import Message
+
 import threading
 import logging
 import time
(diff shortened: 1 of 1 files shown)


can/notifier.py
```diff
@@ -24,37 +36,40 @@ def __init__(self, bus, listeners, timeout=1.0, loop=None):
...
 
-        :param can.BusABC bus:
+        :param bus:
             CAN bus instance.
         """
-        if self._loop is not None and hasattr(bus, "fileno") and bus.fileno() >= 0:
+        if (
+            self._loop is not None
+            and hasattr(bus, "fileno")
+            and bus.fileno() >= 0  # type: ignore
```
karlding: This is disabled here because mypy doesn't really deal well with `hasattr` checks. See the discussion [here](https://github.com/python/mypy/issues/1424). We could either:

1. Disable the type checker (like is done here)
2. Perform the cast to `Any`
3. Refactor the code to have `fileno` on the base class return `None` by default (which we can check for), and then subclasses provide their own implementation as needed

I'm personally inclined to 3, but that's a bit more involved than the other 2 (and perhaps should be kept out of this change). Thoughts?

christiansandberg: I vote for 3, but it can return -1 instead of None.

felixdivo: I'd vote for 1. It's just a shortcoming of the type system, and no problem with our API.

hardbyte: I agree that one makes sense for this PR, but 3 would be the best long term solution.

PR #685, opened 2019-08-14
Add support for base 10 in CANReader

multiplor:
Change-Id: I61a04c53276ea5b74e81d77267ef5e063a211d8a

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -23,9 +23,11 @@ class ASCParseError(Exception):
     """ASC file could not be parsed correctly."""
 
 
-BASES = {'dec': 10, 'hex': 16}
+BASES = {"dec": 10, "hex": 16}
 CAN_MSG_EXT = 0x80000000
 CAN_ID_MASK = 0x1FFFFFFF
+BEGIN_TRIGGERBLOCK = "Begin Triggerblock"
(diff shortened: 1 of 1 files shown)


felixdivo:
And both linter and formatter complain, you can run them locally with `pylint --rcfile=.pylintrc-wip can/` and `black --verbose .` after installing the dependencies with `pip3 install -r requirements-lint.txt -U`.

can/io/asc.py
```diff
@@ -38,15 +44,32 @@ def __init__(self, file):
...
+            line = self.file.readline()
+            m = re.search(base_regex, line)
+            if m:
+                base = m.group('base')
+                if base not in BASES:
+                    raise ASCParseError("Support for base %s not implemented" % base)
+                self.base = BASES[base]
+                self.absolute_timestamps = m.group('timestamps') == 'absolute'
+                break
+        self.file.seek(0)
```
christiansandberg: Is it really necessary to seek? Couldn’t we just parse the header until we hit Begin Triggerblock and then continue with the actual log entries? The reason I’m asking is because there might be cases where one would like to stream the log over e.g. HTTP or something and then it would not be possible to seek.

felixdivo:
I think this is superseded by #820.

PR #687, opened 2019-08-15
Bugfix release v3.3.2

### LINKED ISSUE #686: Backport PCAN rtr fix to 3.x release
From #659 
> In PcanBus.send(), we initially set msgType based on all the flags of msg, including RTR. In the if/else for self.fd, we are incorrectly resetting it if rtr.

This issue is to create a 3.3.2 release with the fix back ported.

hardbyte:
Closes #686 by back porting @craffert0's fix from #659

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -1,3 +1,8 @@
+Version 3.3.2
+====
+
+Minor bug fix release addressing issue in PCAN RTR.
+
 Version 3.3.1
 ====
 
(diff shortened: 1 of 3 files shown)


christiansandberg:
Shouldn’t #684 also be included?

hardbyte:
It looks to me like that pcan bug didn't go into a release so doesn't need to be backported

christiansandberg:
Oh you’re right!

christiansandberg:
It doesn’t seem to have been published on PyPI.

craffert0:
Can you please publish this release on PyPi?

hardbyte:
Sorry I missed that - tracking in https://github.com/hardbyte/python-can/issues/698

PR #688, opened 2019-08-15
Allow message sequence in bus.send

pierreluctg:
I have only optimize the neovi bus to handle messages sequence. Other interfaces will have to be modified if their API support it.

- [x] Agree on the new API
- [ ] update doc/internal-api.rst#extending-the-busabc-class accordingly.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -3,7 +3,7 @@
 """
 Contains the ABC bus implementation and its documentation.
 """
-
+import collections
 from typing import Iterator, List, Optional, Sequence, Tuple, Union
 
 import can.typechecking
@@ -152,13 +152,37 @@ class BusABC(metaclass=ABCMeta):
         """
         raise NotImplementedError("Trying to read from a write only bus?")
 
-    @abstractmethod
-    def send(self, msg: Message, timeout: Optional[float] = None):
+    def send(
+        self, msgs: Union[Sequence[Message], Message], timeout: Optional[float] = None
+    ):
         """Transmit a message to the CAN bus.
 
         Override this method to enable the transmit path.
 
-        :param Message msg: A message object.
+        :param msgs: A message object of sequence of message objects.
+
+        :param timeout:
+            If > 0, wait up to this many seconds for message to be ACK'ed or
+            for transmit queue to be ready depending on driver implementation.
+            If timeout is exceeded, an exception will be raised.
+            Might not be supported by all interfaces.
+            None blocks indefinitely.
+
+        :raises can.CanError:
+            if the message could not be sent
+        """
+        for msg in self._get_messages_sequence(msgs):
+            self._send_internal(msg, timeout)
+
+    @abstractmethod
+    def _send_internal(
+        self, msgs: Union[Sequence[Message], Message], timeout: Optional[float] = None
+    ):
+        """Transmit a message to the CAN bus.
+
+        Override this method to enable the transmit path.
+
+        :param msgs: A message object of sequence of message objects.
 
         :param timeout:
             If > 0, wait up to this many seconds for message to be ACK'ed or
@@ -217,13 +241,7 @@ class BusABC(metaclass=ABCMeta):
             appropriate as the stopped tasks are still taking up memory as they
             are associated with the Bus instance.
         """
-        if not isinstance(msgs, (list, tuple)):
-            if isinstance(msgs, Message):
-                msgs = [msgs]
-            else:
-                raise ValueError("Must be either a list, tuple, or a Message")
-        if not msgs:
-            raise ValueError("Must be at least a list or tuple of length 1")
+        msgs = BusABC._get_messages_sequence(msgs)
         task = self._send_periodic_internal(msgs, period, duration)
         # we wrap the task's stop method to also remove it from the Bus's list of tasks
         original_stop_method = task.stop
@@ -242,6 +260,17 @@ class BusABC(metaclass=ABCMeta):
             self._periodic_tasks.append(task)
 
         return task
+
+    @staticmethod
+    def _get_messages_sequence(msgs: Union[Sequence[Message], Message]):
+        if not isinstance(msgs, collections.Sequence):
+            if isinstance(msgs, Message):
+                msgs = [msgs]
+            else:
+                raise ValueError("Must be either a Message sequence, or a Message")
+        if not msgs:
+            raise ValueError("Must be at least a sequence of length 1")
+        return msgs
 
     def _send_periodic_internal(
         self,
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -29,7 +29,6 @@ except ImportError as ie:
     )
     ics = None
 
-
 try:
     from filelock import FileLock
 except ImportError as ie:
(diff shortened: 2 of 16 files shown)


christiansandberg:
Is there are particular use-case when this is needed?

pierreluctg:
@christiansandberg the transmit/send method of some of the drivers used by the interfaces in python-can supports transmitting more than one message. The benefit of sending more than one messages in on call varies depending on the driver. For some this will allow sending frames back to back without the USB round time time delay, for other this may help improve the concurrency performance by allowing to release the GIL for a longer period of time, etc.

can/bus.py
```diff
@@ -217,13 +241,7 @@ def send_periodic(
...
             are associated with the Bus instance.
         """
-        if not isinstance(msgs, (list, tuple)):
-            if isinstance(msgs, Message):
-                msgs = [msgs]
-            else:
-                raise ValueError("Must be either a list, tuple, or a Message")
-        if not msgs:
-            raise ValueError("Must be at least a list or tuple of length 1")
+        msgs = self._get_messages_sequence(msgs)
```
felixdivo:
```suggestion
        msgs = BusABC._get_messages_sequence(msgs)
```

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -377,8 +371,17 @@ def send(self, msg, timeout=None):
             message.NetworkID = self.channels[0]
         else:
             raise ValueError("msg.channel must be set when using multiple channels.")
+        return message
+
+    def send(self, msgs, timeout=None):
```
felixdivo: Why is this overridden in here?

can/bus.py
```diff
@@ -152,13 +152,37 @@ def _recv_internal(
...
+        :raises can.CanError:
+            if the message could not be sent
+        """
+        for msg in self._get_messages_sequence(msgs):
+            self._send_internal(msg, timeout)
+
     @abstractmethod
-    def send(self, msg: Message, timeout: Optional[float] = None):
+    def _send_internal(
+        self, msgs: Union[Sequence[Message], Message], timeout: Optional[float] = None
```
felixdivo: Shouldn't the method `_send_internal()` take only a single message at a time? This would enable one to override `_send_internal()` for allow sending messages in general and `send()` to speed things up if supported by the driver/hardware/...

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -377,8 +371,17 @@ def send(self, msg, timeout=None):
...
+        self._send_internal(self._get_messages_sequence(msgs), timeout)
+
+    def _send_internal(self, msg, timeout=None):
+        if not ics.validate_hobject(self.dev):
+            raise CanError("bus not open")
+        messages = tuple(self._msg_to_spy_msg(_) for _ in msg)
```
felixdivo: Underscore is usually a placeholder for an unused argument, which it is not in this case, right?

can/bus.py
```diff
@@ -243,6 +261,17 @@ def wrapped_stop_method(remove_task=True):
 
         return task
 
+    @staticmethod
+    def _get_messages_sequence(msgs):
+        if not isinstance(msgs, (list, tuple)):
```
felixdivo: Can't we check for `typing.Sequence` somehow or simply `collections.Sequence`? Or check for `__iter__` + `__len__`?

can/bus.py
pierreluctg: That will means duplicating some code when the same driver API is use to send one or more messages.

```python
    def send(self, msgs, timeout=None):
        self._send_internal(self._get_messages_sequence(msgs), timeout)

    def _send_internal(self, msg, timeout=None):
        if not ics.validate_hobject(self.dev):
            raise CanError("bus not open")
        messages = tuple(self._msg_to_spy_msg(can_msg) for can_msg in msg)

        try:
            ics.transmit_messages(self.dev, messages)
        except ics.RuntimeError:
            raise ICSApiError(*ics.get_last_api_error(self.dev))
```

VS

```python
    def send(self, msgs, timeout=None):
        if not ics.validate_hobject(self.dev):
            raise CanError("bus not open")
        messages = tuple(self._msg_to_spy_msg(can_msg) for can_msg in msg)

        try:
            ics.transmit_messages(self.dev, messages)
        except ics.RuntimeError:
            raise ICSApiError(*ics.get_last_api_error(self.dev))

    def _send_internal(self, msg, timeout=None):
        if not ics.validate_hobject(self.dev):
            raise CanError("bus not open")

        try:
            ics.transmit_messages(self.dev, self._msg_to_spy_msg(msg))
        except ics.RuntimeError:
            raise ICSApiError(*ics.get_last_api_error(self.dev))
```

felixdivo: Well, the second one does not make sense, we would just not implement `_send_internal()` in that case, right?

Yet another approach, which might be a bit cleaner and more versatile:
`send()` receives some `Message` or sequence of them, and transforms them into a sequence, possibly containing just a single element. It is (pretty much?) never overridden. It takes away the (ugly/error prone) handling of sequences discussed somewhere else in this PR.
Then, `_send_internal()` handles only sequences, and in cases where the driver supports multiple messages, it does so directly, and in other cases it simply starts with a loop `for msg in msgs:`.

pierreluctg: @felixdivo, this is essentially what I did, except the loop is in the abstract class `send`method.

This is to reduce the the changes. A simple rename of `send` --> `_send_internal` in the derived class vs. rename of `send` --> `_send_internal` and adding wrapping code in a for loop.

can/bus.py
```diff
@@ -152,13 +152,37 @@ def _recv_internal(
...
+            self._send_internal(msg, timeout)
+
     @abstractmethod
-    def send(self, msg: Message, timeout: Optional[float] = None):
+    def _send_internal(
+        self, msgs: Union[Sequence[Message], Message], timeout: Optional[float] = None
+    ):
         """Transmit a message to the CAN bus.
 
         Override this method to enable the transmit path.
```
hardbyte, line 183: This comment can now be updated to be user facing, and we will have to update the [creating a new interface docs](https://python-can.readthedocs.io/en/master/development.html#creating-a-new-interface-backend)

can/interfaces/ics_neovi/neovi_bus.py
felixdivo: Ahh. I'd just add a comment.

```suggestion
    def send(self, msgs, timeout=None):
        # we always pass the entire sequence as a whole
```

can/bus.py
felixdivo: Seems like we all abandoned this at some point last year ...

Giving it another look now your changes seem perfectly reasonable.

felixdivo:
It seems like the merge conflicts aren't very extensive. And the only things to be done would be to update some documentation (and possibly change interfaces added by now), right?

@pierreluctg Do you have time to complete this PR?

pierreluctg:
@felixdivo, I will try to have a look this week.

felixdivo:
@pierreluctg Do you still plan on completing this?

felixdivo:
At this point, it is probably easier to re-do the PR. In any case, we should probably open an issue to discuss whether we even want this first.

pierreluctg:
I do not have time to work in this now, unfortunately.
Open to anyone to take over this work with a new PR.

PR #694, opened 2019-09-04
Cleanup the examples folder

felixdivo:
The examples were improved by switching to Python 3 features/style, adding some documentation, removing clutter, switching to the virtual interface such that they work everywhere and always closing the bus (by using a context manager) and other cleanups.

The folder is now checked by pylint and mypy on Travis as well.


felixdivo:
> Thanks for keeping the examples ✨

Sure, they're probably gold nuggets for newcomers! They have simply not gotten that much attention from us lately.

PR #701, opened 2019-09-11
ASCReader: Skip J1939TP messages

bessman:
.asc files can contain J1939 Transport Protocol messages, which look something like this:
`0.297743 1 J1939TP FEE3p 6 0 0 - Rx d 23 A0 0F A6 60 3B D1 40 1F DE 80 25 DF C0 2B E1 00 4B FF FF 3C 0F 00 4B FF FF FF FF FF FF FF FF FF FF FF FF`
This change makes ASCReader skip such messages, instead of breaking.


christiansandberg:
Could you also add a line in test/data/logfile.asc?

bessman:
Of course. In a real logfile the J1939TP entry would be preceded by a number of normal CAN messages containing parts of the J1939TP message. Should I also include those in the example file?

felixdivo:
We should probably document that behavior in the class. Can you do that?

Besides, is there some official documentation of the file format we can link to?

 > Of course. In a real logfile the J1939TP entry would be preceded by a number of normal CAN messages containing parts of the J1939TP message. Should I also include those in the example file?

Why not add the J1939TP in a few places, like at the start, in between and at the end? It shouldn't hurt.

felixdivo:
Than you @bessman for these improvements!

PR #703, opened 2019-09-28
Add auto-modifying cyclic tasks

bessman:
This is a proposal to add an optional callback function to ModifiableCyclicTaskABC which manipulates each message automatically before sending it. This feature can be used to, for example, increment a counter or compute a checksum, as shown in examples/cyclic_checksum.py.

If you think this functionality is appropriate for merging, I will of course add it to all interfaces which override the fallback ThreadBasedCyclicSendTask.


christiansandberg:
I think this is something that is much needed. Things like checksums and alive counters are getting more common as safety is becoming a major driving force. Unfortunately it's only going to work with the thread based software method. Passing the task entirely to the kernel or hardware also means you can't do any processing in user space.

Should we provide an interface to allow the user to choose the thread based transmission at the expense of poor timing?

bessman:
I've thought about this some more, and I'm no longer sure a callback is the way to go. At least for the issue of counters and checksums, a better solution is to pre-calculate all CRCs and send them using CyclicSendTask's already existing ability to send a sequence of messages, e.g.:
```
def crc(msg, counter):
    #  checksum calculation
    return checksum


msgs = []

for counter in range(16):
    m = Message(arbitration_id=0x123, data=[1, 2, 3, 4, 5, 6, 7, 0])
    m.data[7] = counter + (crc(m, counter) << 4)
    msgs.append(m)

bus.send_periodic(msgs)
```
This has the added benefit that the user does not have to keep track of the counter when the message data changes. With the callback method, the counter will reset when the data changes unless the user manually checks the counter of the most recently sent message and adds it to the message.

The callback is more flexible, and there could be applications where pre-calculating all messages is not possible (though I can't think of any at the moment). But if that flexibility comes at the cost of performance, perhaps it isn't worth it.

zariiii9003:
Maybe we could implement this in the `Message` class? The `data` attribute  could be a generator that calculates the next bytearray for the message.
The user could implement his calculation by subclassing `Message` and reimplementing the `data` attribute. Otherwise it would just return a private `_data` variable.

bessman:
> Maybe we could implement this in the `Message` class? The `data` attribute could be a generator that calculates the next bytearray for the message.
> The user could implement his calculation by subclassing `Message` and reimplementing the `data` attribute. Otherwise it would just return a private `_data` variable.

How would this work, exactly? If data is a generator, wouldn't we need to call next(data) all over the place? That seems like a big change.

hardbyte:
This is a good proposal, and a clean default implementation. I also don't see how a callback would interact nicely with the backends that already offload periodic messages (without falling back to the threading approach), which isn't a great user experience.

However if this is useful for a few people (e.g. #809) and we can implement it in a way that doesn't harm the performance when not being used then I'd be happy to merge it in. I'm thinking that in the socketcan case we make it very clear in the docs that giving a callback will mean looser timing of messages being sent based on Python threads rather than the kernel.

zorce:
Would love to see something like this, but if possible a kernel implementation to handle similar issues.
Also I have a use case where in an industry use case "CAN in Automation (CiA)" where you might want to read multiple muxed signals. And hence want to change the mux value between every frame, the frames have an inhibit time of let say 100ms between the messages.

hartkopp:
> Would love to see something like this, but if possible a kernel implementation to handle similar issues.
> Also I have a use case where in an industry use case "CAN in Automation (CiA)" where you might want to read multiple muxed signals. And hence want to change the mux value between every frame, the frames have an inhibit time of let say 100ms between the messages.

The CAN_BCM (in-kernel Broadcast Manager) can send a sequence of up to 256 CAN(FD) frames, e.g. every 100ms the next frame in the row is transmitted. The nframes element in the bcm_msg_head is usually 1 but can be extended to 2 .. 256 to send a sequence.
So you can send mux messages with just one TX_SETUP command:
https://www.kernel.org/doc/html/latest/networking/can.html#broadcast-manager-message-sequence-transmission

Same on the receiving side: When you want to filter on data changes on a specific MUX ID you define the mux mask (to define where the MUX ID sits in the CAN frame data) and put the MUX ID in that place in the RX_SETUP filter.
https://www.kernel.org/doc/html/latest/networking/can.html#broadcast-manager-multiplex-message-receive-filter

zorce:
> > Would love to see something like this, but if possible a kernel implementation to handle similar issues.
> > Also I have a use case where in an industry use case "CAN in Automation (CiA)" where you might want to read multiple muxed signals. And hence want to change the mux value between every frame, the frames have an inhibit time of let say 100ms between the messages.
> 
> The CAN_BCM (in-kernel Broadcast Manager) can send a sequence of up to 256 CAN(FD) frames, e.g. every 100ms the next frame in the row is transmitted. The nframes element in the bcm_msg_head is usually 1 but can be extended to 2 .. 256 to send a sequence.
> So you can send mux messages with just one TX_SETUP command:
> https://www.kernel.org/doc/html/latest/networking/can.html#broadcast-manager-message-sequence-transmission
> 
> Same on the receiving side: When you want to filter on data changes on a specific MUX ID you define the mux mask (to define where the MUX ID sits in the CAN frame data) and put the MUX ID in that place in the RX_SETUP filter.
> https://www.kernel.org/doc/html/latest/networking/can.html#broadcast-manager-multiplex-message-receive-filter

Didn't know that, thanks for the information. Is this also exposed with other backends, noticed there is a cyclic_multiple example added 2 years ago. but I was running on version 3.3.4. will try to update and try it out.

hartkopp:
> Didn't know that, thanks for the information. Is this also exposed with other backends, noticed there is a cyclic_multiple example added 2 years ago. but I was running on version 3.3.4. will try to update and try it out.

I've just programmed the CAN_BCM module and started to look into the API that python-can provides. Unfortunately I'm not that good in Python to get behind the current API concept for handling (and on-the-fly updating) a sequence of cyclic messages ;-)

leu1dy:
I am also interested about this feature. AliveCounter & Checksums are part of the daily work in automotive domain.

Any chance to see this feature at a release state ?

Thank you

james-goin (approved):

zariiii9003:
I resolved the merge conflicts. There are still a few problems though:

- `IXXATBus._send_periodic_internal` must be adapted
- IXXATBus and socketcan should warn, if modifier_callback is not None and use the thread based cyclic send task as fallback

Personally i would prefer a mutating callback with no return value though (`Callable[[Message], None]`)

zariiii9003:
I think this can be merged if everyone agrees with the changes.

kamaln16:
Hello,

I have tried implementing this auto-modifying task to my python code on my RPi 3B+ but when I switch from the virtual interface to socketcan, I get errors.
Before implementing these changes, the socketcan interface works fine.

Errors:
Cannot import module can.interfaces.socketcan for CAN interface 'socketcan'
Cannot import name 'CanProtocol' from 'can'

I have tried to look up the error online and fix it myself, but to no avail. Anyone have any idea why this is happening?

Issue #591, opened 2019-05-17
Look into CI testing on Arm with shippable


hardbyte:
http://docs.shippable.com/platform/tutorial/workflow/run-ci-builds-on-arm/

> It is the only hosted CI/CD platform to offer native CI/CD for Arm developers.
> Even better news? Building on Arm machines is absolutely free for open source projects!

marcel-kanter:
sounds nice. At our company we use some fpgas with arm hard cores and linux runs on it.
I thought about a file in the doc showing the  platform support for each interface some weeks ago. Some interfaces support Windows only. That should be changed. I hope I have enough time to implement linux support (x86 and arm) for the ixxat interface for the 4.0 release. My not so secret plan is to use this interface as an example for the exception handling and multiplatform support.

hardbyte:
Many of the windows only interfaces can be used on Linux via socketcan. Documenting the support would be a good idea though

PR #705, opened 2019-10-08
Add Travis CI support for arm64 architectures

karlding:
Add support for the arm64 architecture from Travis CI's alpha offering.
This enables the test suite to run within an arm64 LXD container (minus
the tests that exercise SocketCAN, as the image is missing the required
kernel modules).


karlding:
Proof that the worker is running on `arm64` from the [job log](https://travis-ci.org/hardbyte/python-can/jobs/594982587#L1-L5):

```
worker_info
Worker information
hostname: 29910ceb-cb30-4562-8175-bf4b6413bc34@1510.lxd-arm64-05-org
version: v6.2.1-20-g4ca60c6 https://github.com/travis-ci/worker/tree/4ca60c69640778fbc5c3203c4f4360ff72b969f1
instance: travis-job-hardbyte-python-can-594982587 b4a942b48ea3f45374c56e146cbe9b9c2c36d834426519af11d977b3cc5f7ff6 (via amqp)
startup: 3.534169152s
```

christiansandberg:
I don’t want to be boring and all, but what part of the library needs testing on specific architectures?

karlding:
Not really specific architectures, but there's code that relies on various sizes and alignments of types. The only way to really cover that besides mocking is to run the tests on different build environments.

There's also #591 in the backlog which calls out ARM testing.

christiansandberg:
I know SocketCAN and PCAN have some OS and architectural differences so if we could test against the actual drivers then there would definitely be a point. But if we can’t test the integration then I can’t see how that increases our confidence. It does consume more resources and requires future maintenance so I think we need to investigate which tests will benefit from this.

felixdivo:
How are we going to proceed on this? Letting this rot here isn't too helpful either.

hardbyte (approved):
I'm happy to add it.

PR #708, opened 2019-10-12
VectorBus - add methods to handle non message events

zariiii9003:
Add methods to handle vector driver events that are not caused by incoming messages. Can be implemented by subclassing.

I do not know if the docstrings are okay. `sphinx-build doc/ doc/_build` causes an error at `reading sources... [ 16%] bus`:
`AttributeError: 'property' object has no attribute 'expandtabs'`


zariiii9003:
Hello Christian,
the methods are not really meant to be called from outside. They are called from within `_recv_internal`.
The user has to implement the methods though. Would you remove the underscore anyway?

christiansandberg:
If they start with an underscore there is no guarantee that they will be called in future versions, which might break applications relying on their existence. If they are public, they need to be treated according to semantic versioning meaning backwards incompatible changes may only be introduces in a new major version.

PR #710, opened 2019-10-23
Add typing annotations for SocketCAN interface

karlding:
Add typing annotations for the SocketCAN interface.

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -5,6 +5,8 @@ along some internal methods.
 At the end of the file the usage of the internal methods is shown.
 """
 
+from typing import Dict, List, Optional, Sequence, Tuple, Type, Union
+
 import logging
 import ctypes
 import ctypes.util
@@ -34,11 +36,17 @@ from can.broadcastmanager import (
 from can.interfaces.socketcan.constants import *  # CAN_RAW, CAN_*_FLAG
 from can.interfaces.socketcan.utils import pack_filters, find_available_interfaces
 
-
 # Setup BCM struct
-def bcm_header_factory(fields, alignment=8):
+def bcm_header_factory(
+    fields: List[Tuple[str, Union[Type[ctypes.c_uint32], Type[ctypes.c_long]]]],
+    alignment: int = 8,
+):
     curr_stride = 0
-    results = []
+    results: List[
+        Tuple[
+            str, Union[Type[ctypes.c_uint8], Type[ctypes.c_uint32], Type[ctypes.c_long]]
+        ]
+    ] = []
     pad_index = 0
     for field in fields:
         field_alignment = ctypes.alignment(field[1])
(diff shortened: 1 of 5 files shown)


can/interfaces/socketcan/socketcan.py
```diff
@@ -300,14 +313,20 @@ class CyclicSendTask(
 
     """
 
-    def __init__(self, bcm_socket, messages, period, duration=None):
+    def __init__(
+        self,
+        bcm_socket: socket.socket,
+        messages: Union[Sequence[Message], Message],
```
hardbyte: sigh

can/interfaces/socketcan/socketcan.py
```diff
@@ -496,7 +522,7 @@ def capture_message(sock, get_channel=False):
 
     # Fetching the timestamp
     binary_structure = "@LL"
-    res = fcntl.ioctl(sock, SIOCGSTAMP, struct.pack(binary_structure, 0, 0))
+    res = fcntl.ioctl(sock.fileno(), SIOCGSTAMP, struct.pack(binary_structure, 0, 0))
```
hardbyte: 👍

can/interfaces/socketcan/socketcan.py
```diff
@@ -742,14 +782,14 @@ def _detect_available_configs():
     #
     log.setLevel(logging.DEBUG)
 
-    def receiver(event):
+    def receiver(event: threading.Event) -> None:
```
hardbyte: I have no idea anymore if this should also support `multiprocessing.Event`?

can/typechecking.py
```diff
@@ -8,10 +8,11 @@
 
 import mypy_extensions
 
-CanFilter = mypy_extensions.TypedDict(
-    "CanFilter", {"can_id": int, "can_mask": int, "extended": bool}
+CanFilter = mypy_extensions.TypedDict("CanFilter", {"can_id": int, "can_mask": int})
```
hardbyte: Is this correct? I'd have thought both `CanFilter` and `CanFilterExtended` have an `extended` bool attribute - just one is True and one is False?

karlding: I would've expected that also, but currently in the code we also accept cases where the `extended` key is not present. The SocketCAN interface is one such place:

```python
def pack_filters(can_filters=None):
    if can_filters is None:
        # Pass all messages
        can_filters = [{"can_id": 0, "can_mask": 0}] # <---- here

    can_filter_fmt = "={}I".format(2 * len(can_filters))
    filter_data = []
    for can_filter in can_filters:
        can_id = can_filter["can_id"]
        can_mask = can_filter["can_mask"]
        if "extended" in can_filter:
            # Match on either 11-bit OR 29-bit messages instead of both
            can_mask |= CAN_EFF_FLAG
            if can_filter["extended"]:
                can_id |= CAN_EFF_FLAG
        filter_data.append(can_id)
        filter_data.append(can_mask)

    return struct.pack(can_filter_fmt, *filter_data)
```

hardbyte: ok thanks for clarifying for me

PR #713, opened 2019-10-23
Fix crash in Canutils Log Reader when parsing RTR frames

koberbe:
Variable dataBin in CanutilsLogReader was not defined before it was accessed when reading remote frames.


karlding:
Thanks for the fix! I guess this is indicative of how many people use RTR frames, since it's been present for over a year :joy:

@hardbyte we probably will want to backport this to a `3.x.x` series release?

PR #724, opened 2019-11-14
Added keycheck for windows platform

tamenol:
PCANBasic will log an error when the PEAK-Driver key could not be found.
I figured this was the best place to add the check, so you don't need to keep a function on a more higher level.


tamenol:
Yes, I had problems when running my program on Windows PC's when the driver wasn't installed. This check lets me finish the program in exception when not present.

can/interfaces/pcan/basic.py
```diff
@@ -498,9 +500,16 @@ class PCANBasic:
...
+        # Loads the PCANBasic.dll and checks if driver is available
         if platform.system() == "Windows":
             self.__m_dllBasic = windll.LoadLibrary("PCANBasic")
+            aReg = winreg.ConnectRegistry(None, winreg.HKEY_LOCAL_MACHINE)
+            try:
+                aKey = winreg.OpenKey(aReg, r"SOFTWARE\PEAK-System\PEAK-Drivers")
+                winreg.CloseKey(aKey)
+            except WindowsError:
+                logger.error("Exception: The PEAK-driver couldn't be found!")
+            winreg.CloseKey(aReg)
```
felixdivo: This might be done in a finally block.

tamenol: will have a look aswell here

tamenol: See commit https://github.com/hardbyte/python-can/pull/724/commits/f542d8ce3a731521f862d72ab7a4a793a69c7902
Is this what you meant? Didn't see how else I'd use a finally block here

felixdivo: Yeah, just move the `winreg.CloseKey(aReg)` in line 512 into the finally block and then also remove the identical call in line 509.

tamenol: @felixdivo line 509 is different from the line 512 one. I almost also made the same assumption but the former is closing the key, the latter is closing the registry.

felixdivo: Whoops, yep. Then this is fine. Thanks for taking the time.

PR #725, opened 2019-11-14
Added status_string method to return simple status strings

tamenol:
status_string returns the current status of the bus in a single word.

Ok, so I have no idea this is warrented to get merged. I wrote this function as I needed a single word for the status of the bus (just like you have in PCAN-View). And I didn't want to import .basic in my project to get the needed values, hence why I put it in the `PcanBus` class


can/interfaces/pcan/pcan.py
```diff
@@ -528,6 +528,32 @@ def _detect_available_configs():
...
+        elif status == PCAN_ERROR_BUSHEAVY:
+            return "BUSHEAVY"
+        elif status == PCAN_ERROR_BUSWARNING:
+            return "BUSWARNING"
+        elif status == PCAN_ERROR_BUSPASSIVE:
+            return "BUSPASSIVE"
+        elif status == PCAN_ERROR_BUSOFF:
+            return "BUSOFF"
+        else:
+            return "NonImplementedError"
```
felixdivo: I' return `None` in this case, to make it more explicit that something went wrong.

can/interfaces/pcan/pcan.py
```diff
@@ -528,6 +528,32 @@ def _detect_available_configs():
             )
         return channels
 
+    def status_string(self):
```
felixdivo: I'd replace the `rtype` comment with an actual type hint: 

```python
def status_string(self) -> Optional[str]:
```

felixdivo:
Sorry for the long delay, do you still have tome to fix this?

tamenol:
@felixdivo No problem. And yes, I'll have a look this weekend/in the course of next week.

can/interfaces/pcan/pcan.py
tamenol: Yes, good idea. However, when looking back on this, wouldn't it be better if I made a dictonary with all the errors and their strings in `basic.py` and retrieve them with this function, so the class in `pcan.py` doesn't get cluttered much?

tamenol: @felixdivo see commit https://github.com/hardbyte/python-can/pull/725/commits/890128f40da623299f153d7e0fed1de028f2d417

felixdivo:
Looks good to me.

felixdivo:
Except [the formatter](https://black.readthedocs.io/en/stable/) is complaining [here](https://travis-ci.org/hardbyte/python-can/jobs/645876354#L321). You can fix it by executing `black .` in `python-can/can/`.

felixdivo:
What a massive change 👍

PR #731, opened 2019-12-01
Add support for "Robotell" USB-CAN interface

deonvdw:
This patch adds support for the Chinese USB-CAN interface made by Robotell and variously cloned. See https://www.amobbs.com/thread-4651667-1-1.html for some background on the original devices.

Clone devices typically make reference to EmbdedConfig and EmbededDebug.

This is the device I tested with: https://www.aliexpress.com/item/32878710246.html
Should also work with for example https://www.aliexpress.com/item/32992507201.html


can/interfaces/robotell.py
```diff
@@ -0,0 +1,404 @@
...
+        """
+        :param str channel:
+            port of underlying serial or usb device (e.g. /dev/ttyUSB0, COM8, ...)
+            Must not be empty.
+        :param int ttyBaudrate:
+            baudrate of underlying serial or usb device
+        :param int bitrate:
+            CAN Bitrate in bit/s. Value is stored in the adapter and will be used as default if no bitrate is specified
+        :param float poll_interval:
+            Poll interval in seconds when reading messages
```
christiansandberg: You can remove this.

can/interfaces/robotell.py
```diff
@@ -0,0 +1,404 @@
...
+            self.set_bitrate(bitrate)
+
+        self.channel_info = "Robotell USB-CAN s/n %s on %s" % (
+            self.get_serial_number(1),
+            channel,
+        )
+        logger.info("Using device: {}".format(self.channel_info))
+
+        super().__init__(
+            channel, ttyBaudrate=115200, bitrate=None, rtscts=False, **kwargs
```
christiansandberg: You don’t have to feed those specific keywords to the super class.

can/interfaces/robotell.py
```diff
@@ -0,0 +1,404 @@
...
+            Bitrate in bit/s
+        """
+        if bitrate <= self._MAX_CAN_BAUD:
+            self._writeconfig(self._CAN_BAUD_ID, bitrate)
+        else:
+            raise ValueError(
+                "Invalid bitrate, must be less than " + str(self._MAX_CAN_BAUD)
+            )
+
+    def _set_auto_retransmit(self, retrans_flag):
```
christiansandberg: Any methods that you want to be public should not be prefixed with an underscore.

can/interfaces/robotell.py
deonvdw: Good point! I used the slcan driver as template and blindly copied this parameter description even though I don't use it. Removing.

Think we should also remove the parameter description from slcan.py - it does not seem to be used there either.

can/interfaces/robotell.py
deonvdw: Updated to remove extra parameters.

Statement was copied from slcan.py - perhaps the original should be updated too.

can/interfaces/robotell.py
deonvdw: Cleaned up, thanks.

PR #733, opened 2019-12-02
fixes #732: add support for VN8900 xlGetChannelTime function

### LINKED ISSUE #732: XL_ERR_NOT_IMPLEMENTED when calling xlGetSyncTime on VN8900
Hello,

I'm trying to use a VN8900 device but get this error during the __init__ https://github.com/hardbyte/python-can/blob/release-3.3.2/can/interfaces/vector/canlib.py#L209 (using version 3.3.2). 

I have multiple VN devices connected (including a VN7600), but I don't use the serial number, I just assign a custom applciation name in the vector hardware configuration utility.

danielhrisca:


can/interfaces/vector/canlib.py
```diff
@@ -282,7 +282,10 @@ def __init__(
 
         # Calculate time offset for absolute timestamps
         offset = xlclass.XLuint64()
-        xldriver.xlGetSyncTime(self.port_handle, offset)
+        try:
+            xldriver.xlGetSyncTime(self.port_handle, offset)
+        except:
+            xldriver.xlGetChannelTime(self.port_handle, self.mask, offset)
```
christiansandberg: You could another level of try except so if both fails we can set `self._time_offset = 0` since it is not required to have absolute timestamps.

danielhrisca: done

danielhrisca:
@christiansandberg 

I think it's ready

Issue #738, opened 2019-12-06
Incorrect pypi wheel which might destroy other packages


chrisoro:
When installing your released pypi packaged it will install the following taken from RECORD:

> ../../Scripts/__pycache__/can_logger.cpython-37.pyc,,
> ../../Scripts/__pycache__/can_player.cpython-37.pyc,,
> ../../Scripts/__pycache__/can_viewer.cpython-37.pyc,,
> ../../Scripts/can_logger.py,sha256=mLcSR0czHD0kWfA-zRXy_VT64OsiJewRfvtcQB1lL38,166
> ../../Scripts/can_player.py,sha256=PW1zDei0tTGKpqRQLw1ZSyiMyuZM2R7p9Iwzks2EUYo,166
> ../../Scripts/can_viewer.py,sha256=2iiEVNNM6WBhiBOLwHWPU9qn4Sq0R38AJxTBtjjfZQA,166
> can/CAN.py,sha256=gm1FzhF3QVj_5W7HQkPrhIkZrqAb13vi9S6Rxicuuh8,941
> can/__init__.py,sha256=UaX_PSKHNZbHRg9visUW5k1wWj0CbjlC3CCgpqoBONo,1227
> can/__pycache__/CAN.cpython-37.pyc,,
> can/__pycache__/__init__.cpython-37.pyc,,
> can/__pycache__/broadcastmanager.cpython-37.pyc,,
> can/__pycache__/bus.cpython-37.pyc,,
> can/__pycache__/ctypesutil.cpython-37.pyc,,
> can/__pycache__/interface.cpython-37.pyc,,
> can/__pycache__/listener.cpython-37.pyc,,
> can/__pycache__/logger.cpython-37.pyc,,
> can/__pycache__/message.cpython-37.pyc,,
> can/__pycache__/notifier.cpython-37.pyc,,
> can/__pycache__/player.cpython-37.pyc,,
> can/__pycache__/thread_safe_bus.cpython-37.pyc,,
> can/__pycache__/util.cpython-37.pyc,,
> can/__pycache__/viewer.cpython-37.pyc,,
> can/broadcastmanager.py,sha256=aK7ACLJqXEijHfAp6HFZxhlQRQAe0V_y-lo6Sh2RAqk,5184
> can/bus.py,sha256=zRnIsqI8u42DoIN5Gk9KK3LN4O7tDrbwEvcnI8_P27w,13594
[...]

PR #740, opened 2019-12-06
Exclude all test packages in wheel, not just toplevel

chrisoro:
See https://github.com/hardbyte/python-can/issues/738


hardbyte:
Build is failing for an unrelated reason - fixed in #745

Issue #746, opened 2019-12-29
Tests fail on develop


marcel-kanter:
Using commit dcd694efb56d503ed9387edc5cfba9c523d559bf the tests fail.
Windows 10 Pro Version 1909
Python 3.7.6, pytest-5.3.2, py-1.8.1, pluggy-0.13.1
pytest plugins: timeout-1.3.3, cov-2.8.1, hypothesis-4.57.0

output of python setup.py test:

test/test_message_class.py::TestMessageClass::test_methods PASSED        [ 62%]
test/test_message_filtering.py::TestMessageFiltering::test_match_all PASSED [ 62%]
test/test_message_filtering.py::TestMessageFiltering::test_match_example_message PASSED [ 63%]
test/test_message_filtering.py::TestMessageFiltering::test_match_filters_is_empty PASSED [ 63%]
test/test_message_sync.py::TestMessageSync::test_general PASSED          [ 64%]
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++

After commenting out the pytest timeout detorator for the test, it results in:

test/simplecyclic_test.py::SimpleCyclicSendTaskTest::test_cycle_time FAILED [ 42%]
test/test_scripts.py::TestLoggerScript::test_do_commands_exist FAILED    [ 72%]
test/test_scripts.py::TestPlayerScript::test_do_commands_exist FAILED    [ 73%]
================= 3 failed, 194 passed, 29 skipped in 36.53s ==================
================================== FAILURES ===================================
__________________ SimpleCyclicSendTaskTest.test_cycle_time ___________________

self = <test.simplecyclic_test.SimpleCyclicSendTaskTest testMethod=test_cycle_time>

    @unittest.skipIf(
        IS_CI,
        "the timing sensitive behaviour cannot be reproduced reliably on a CI server",
    )
    def test_cycle_time(self):
        msg = can.Message(
            is_extended_id=False, arbitration_id=0x123, data=[0, 1, 2, 3, 4, 5, 6, 7]
        )
    
        with can.interface.Bus(bustype="virtual") as bus1:
            with can.interface.Bus(bustype="virtual") as bus2:
    
                # disabling the garbage collector makes the time readings more reliable
                gc.disable()
    
                task = bus1.send_periodic(msg, 0.01, 1)
                self.assertIsInstance(task, can.broadcastmanager.CyclicSendTaskABC)
    
                sleep(2)
                size = bus2.queue.qsize()
[...]

Issue #737, opened 2019-12-05
Absence of can.Message.__eq__


tysonite:
I noticed that ` can.Message.__eq__` has been removed in [this](https://github.com/hardbyte/python-can/commit/c37420d23e17f48067e7cc2de4778a88c743061e#diff-7080e7643f241026215c54f04aa5b7eaL132) revision. Is it really expected?

It was useful in verification of mocked calls in tests, as well as just comparing diffrerent messages. I understand that `equals()` provides the comparison mechanism now, but it looks to me that built-in equality function gives better flexibility, where not necessary to manage comparsion of timestamps.

christiansandberg:
The main issue is that there is no universal definition of when two messages are equal. Especially if the messages are different in time or appear on different channels.

tysonite:
Can those restrictions be relaxed while comparing two messages? The same issue applies to `equals()`, right? But it is still there.

christiansandberg:
With a method you can provide arguments controlling the behavior so it will be more flexible depending on your specific needs. You can now choose if you want to compare timestamps or not. The implementation of __eq__ will always be fixed and never be able to satisfy all use cases.

tysonite:
I do agree with what you are saying. On the other hand, invocation of a method requires specifying arguments if behaviour has to be changed. If you are fine with default behaviour, arguments are just omitted as they have default values.

I am wondering if two methods can co-exist. The **eq** calls **equals** by default without special arguments, but if anyone needs specialization, he calls **equals** specifically with a set of required arguments.

tysonite:
BTW, please feel free to close this issue if the project direction has been already confirmed in regards of that.

christiansandberg:
@felixdivo what do you think?

felixdivo:
Hey :) It's been a while ...

On a technical level, two methods could co-exist and call each other, sure. I think it would be okay (and I personally think also a good idea) to provide `__eq__` like that. It's the usual discussion on simple but possibly misleading vs. explicit but somewhat complicated. However, it is again a breaking change! (Although probably many will not have a problem with this change)

We should check that the documentation in the code explains how comparison works. And there should be some rationale on this in the "user guide" documentation as well.

felixdivo:
Done by #747.

PR #747, opened 2019-12-29
For equals the comparison of the channel is switchable. The operator …

marcel-kanter:
…overload checks two Message objects without comparison of timestamp or channel.

Attempt for #737

For test results be aware of issue #746. In test_message_sync.py the pytest timeout decorator needs to be commented and test/simplecyclic_test.py::SimpleCyclicSendTaskTest::test_cycle_time, test/test_scripts.py::TestLoggerScript::test_do_commands_exist and test/test_scripts.py::TestPlayerScript::test_do_commands_exist fail.

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -44,7 +44,7 @@ class Message:
         "is_fd",
         "bitrate_switch",
         "error_state_indicator",
-        "__weakref__"  # support weak references to messages
+        "__weakref__",  # support weak references to messages
     )
 
     def __init__(
@@ -60,7 +60,7 @@ class Message:
         is_fd: bool = False,
         bitrate_switch: bool = False,
         error_state_indicator: bool = False,
-        check: bool = False
+        check: bool = False,
     ):
         """
         To create a message object, simply provide any of the below attributes
@@ -118,7 +118,7 @@ class Message:
                 "R" if self.is_remote_frame else " ",
                 "F" if self.is_fd else " ",
                 "BS" if self.bitrate_switch else "  ",
-                "EI" if self.error_state_indicator else "  "
+                "EI" if self.error_state_indicator else "  ",
             ]
         )
 
@@ -156,7 +156,7 @@ class Message:
         args = [
             "timestamp={}".format(self.timestamp),
             "arbitration_id={:#x}".format(self.arbitration_id),
-            "is_extended_id={}".format(self.is_extended_id)
+            "is_extended_id={}".format(self.is_extended_id),
         ]
 
         if self.is_remote_frame:
@@ -199,7 +199,7 @@ class Message:
             data=self.data,
             is_fd=self.is_fd,
             bitrate_switch=self.bitrate_switch,
-            error_state_indicator=self.error_state_indicator
+            error_state_indicator=self.error_state_indicator,
         )
         return new
 
@@ -215,7 +215,7 @@ class Message:
             data=deepcopy(self.data, memo),
             is_fd=self.is_fd,
             bitrate_switch=self.bitrate_switch,
-            error_state_indicator=self.error_state_indicator
+            error_state_indicator=self.error_state_indicator,
         )
         return new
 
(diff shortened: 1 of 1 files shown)


marcel-kanter:
I'll fix this issue.

marcel-kanter:
Is there any way to tell pylint that after the lines
`if type(self) != type(other):
return False`

The type of other is the type of self?

And no, the following works only because of the quirks in the python implementations and is error-prone (I'll explain why if you ask):
`if not isinstance(other, Message):
return False`

can/message.py
```diff
@@ -279,8 +279,21 @@ def _check(self):
...
 
+    def __eq__(self, other: Any) -> bool:
+        """ Compares a given message with this one.
+        The result is the as same as calling ``equals`` without timestamp and channel check.
+
+        :param other: the message to compare with
+
+        :return: True if the given message equals this one
+        """
+        return self.equals(other, timestamp_delta=None, check_channel=False)
```
christiansandberg: No additional arguments should be passed here so the behavior is always the default.

can/message.py
```diff
@@ -290,29 +303,27 @@ def equals(
...
+                timestamp_delta is None
+                or abs(self.timestamp - other.timestamp) <= timestamp_delta
             )
+            and self.arbitration_id == other.arbitration_id
+            and self.is_extended_id == other.is_extended_id
+            and self.dlc == other.dlc
+            and self.data == other.data
+            and self.is_remote_frame == other.is_remote_frame
+            and self.is_error_frame == other.is_error_frame
+            and (check_channel or self.channel == other.channel)
```
christiansandberg: Should be “not check_channel”.

can/message.py
```diff
@@ -279,8 +279,21 @@ def _check(self):
...
+        :return: True if the given message equals this one
+        """
+        return self.equals(other, timestamp_delta=None, check_channel=False)
+
     def equals(
-        self, other: "Message", timestamp_delta: Optional[Union[float, int]] = 1.0e-6
+        self,
+        other: Any,
+        timestamp_delta: Optional[Union[float, int]] = 1.0e-6,
+        check_channel: bool = False,
```
christiansandberg: We should probably document backwards incompatible changes somewhere.

can/message.py
marcel-kanter: Oh yes.

can/message.py
marcel-kanter: Ok. I was not shure if checking the timestamp should be done or not. So I'll remove the both arguments, and make check_channel True. Then this is a full-property compare (with the recommended float-compare with an epsilon value)

can/message.py
marcel-kanter: See above and with check_channel = True this is full backwards compatible.

can/message.py
christiansandberg: There was a discussion earlier and the conclusion was to compare everything by default as that results in less surprises. In a real world application you might however want to relax some checks so that’s why the check method exists.

marcel-kanter: Yup. Please have a look on the last commit.

can/message.py
```diff
@@ -290,29 +303,29 @@ def equals(
...
+        :param check_channel: compare the channels of the messages
+
+        :return: True if the given message equals this one
         """
         # see https://github.com/hardbyte/python-can/pull/413 for a discussion
         # on why a delta of 1.0e-6 was chosen
+        if self is other:
+            return True
+        if type(self) != type(other):  # pylint: disable=unidiomatic-typecheck
+            return False
```
christiansandberg: It’s probably very unlikely but what if someone subclasses can.Message? Shouldn’t that also be possible to compare with?

marcel-kanter: NO.
Do you have experience on other OO high level languagey like Java, Go, Haskell?

christiansandberg: Allright, I guess it make sense.

PR #753, opened 2020-01-07
Add socket object to mimic the behaviour of SocketCAN for all interfaces

polybassa:
Hi, 
this PR tries to mimic the behaviour of Linux SocketCAN for any interface
I'm a contributor to the Scapy project, where I'm heavily using python-can for automotive related stuff.
During my work with python-can, two problems occur once a while. 
1. If I'm using a interface, for example Vector, I can not create multiple bus objects, since the access to the library is exclusive. SocketCAN differs here.
2. The handling of background threads for receiving asynchronous is inconvenient and for many use-cases I have to implement a select-like behaviour on my own.

This is my first PR to the python-can project. I hope I followed the coding guidelines. 
I only made one unit-test based on SocketCAN, for now. Do you have ideas for further unit tests?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -110,6 +110,7 @@ jobs:
           can/message.py
           can/notifier.py
           can/player.py
+          can/socket.py
           can/thread_safe_bus.py
           can/typechecking.py
           can/util.py
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -32,6 +32,7 @@ from .util import set_logging_level
 
 from .message import Message
 from .bus import BusABC, BusState
+from .socket import Socket
 from .thread_safe_bus import ThreadSafeBus
 from .notifier import Notifier
 from .interfaces import VALID_INTERFACES
--- a/can/socket.py
+++ b/can/socket.py
@@ -15,7 +15,8 @@ log = logging.getLogger("can.socket")
 log_autodetect = log.getChild("detect_available_configs")
 
 
-class SocketsThreadPool(object):
+class SocketsThreadPool:
+    # pylint: disable=no-member
     __instance = None
 
     def __new__(cls):
@@ -34,7 +35,7 @@ class SocketsThreadPool(object):
                     bus, _, _, sockets = self.buses[k]
                 except KeyError:
                     return
-            if len(sockets) == 0:
+            if not sockets:
                 return
             msg = bus.recv(timeout=0.01)
             if msg is None:
@@ -50,7 +51,7 @@ class SocketsThreadPool(object):
                     bus, tx_queue, tx_signal, sockets = self.buses[k]
                 except KeyError:
                     return
-            if len(sockets) == 0:
+            if not sockets:
                 return
             try:
                 tx_signal.acquire()
@@ -79,7 +80,7 @@ class SocketsThreadPool(object):
             bus, tx_queue, tx_signal, sockets = self.buses[k]
             sockets.append(socket)
             filters = [s.filters for s in sockets if s.filters is not None]
-            if len(filters):
+            if filters:
                 bus.set_filters(reduce(add, filters))
             socket.tx_queue = tx_queue
             socket.tx_signal = tx_signal
@@ -111,7 +112,7 @@ class SocketsThreadPool(object):
 
         for k, v in self.buses.copy().items():
             bus, _, tx_signal, sockets = v
-            if len(sockets) == 0:
+            if not sockets:
                 with self.buses_mutex:
                     del self.buses[k]
                 tx_signal.release()
--- a/can/socket.py
+++ b/can/socket.py
@@ -1,0 +1,149 @@
+import logging
+import queue
+import threading
+import time
+import copy
+
+from functools import reduce
+from operator import add
+
+from .interface import Bus
+from .bus import BusABC
+from . import CanError
+
+log = logging.getLogger("can.socket")
+log_autodetect = log.getChild("detect_available_configs")
+
+
+class SocketsThreadPool(object):
+    __instance = None
+
+    def __new__(cls):
+        if SocketsThreadPool.__instance is None:
+            SocketsThreadPool.__instance = object.__new__(cls)
+            SocketsThreadPool.__instance.buses = dict()
+            SocketsThreadPool.__instance.rx_threads = dict()
+            SocketsThreadPool.__instance.tx_threads = dict()
+            SocketsThreadPool.__instance.buses_mutex = threading.Lock()
+        return SocketsThreadPool.__instance
+
+    def recv_function(self, k):
+        while True:
+            with self.buses_mutex:
+                try:
+                    bus, _, _, sockets = self.buses[k]
+                except KeyError:
+                    return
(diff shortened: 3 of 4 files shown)


christiansandberg:
The Notifier class is not sufficient for your use case?

polybassa:
Hi Christian, unfortunately I need a select function. To realize select, I needed individual queues for each socket object.

This is the major difference to notifier class, in my opinion. 

I would be happy about feedback to my implementation, since I’m new to the python-can project.

christiansandberg:
I'm not yet entirely convinced this should go into this package for a number of reasons.

* The use case seems somewhat narrow. Someone else should pitch in if they think this generally useful. Keep in mind that it is also possible to make a separate Python package. We like to keep a fairly narrow focus on just interface abstraction and logging and allow other packages to build more specific functionality on top of it.
* The whole mimicking sockets API feels unintuitive and unnecessary to me. It will probably make more sense to have a different API to avoid confusion with real sockets.
* Since the select method you implemented is non blocking, the main thread that wants to wait for messages will just poll for non-empty queues as fast as the CPU can handle. Apart from wasting resources, just polling the buses directly with a timeout of 0 instead should work equally well or bad.

Please correct me if there is something I have misinterpreted.

christiansandberg:
What do you use the “socket” object for except for obtaining the message? Could it make sense to attach the originating Bus or log reader instance to each Message instead?

polybassa:
Hello Christian, thank you very much for your reply. I'm happy to discuss this PR. 

> I'm not yet entirely convinced this should go into this package for a number of reasons.
> 
>     * The use case seems somewhat narrow. Someone else should pitch in if they think this generally useful. Keep in mind that it is also possible to make a separate Python package. We like to keep a fairly narrow focus on just interface abstraction and logging and allow other packages to build more specific functionality on top of it.
I was considering to put this code into Scapy, first. After some considerations, I felt is more natural to have this socket objects in the python-can project. The main reason for this decision is the BusABC class. This sockets depend on individual filters for each socket. The BusABC provides all the filtering related features, already.
> 
>     * The whole mimicking sockets API feels unintuitive and unnecessary to me. It will probably make more sense to have a different API to avoid confusion with real sockets.
I tried to follow the same API as a `Bus` or `Interface` object of python-can. In my opinion, the socket is just a middle layer, between the real bus object and an application. I wanted to make the socket transparent to use it as drop in replacement of a Bus object. 
For further information about Hartkopps SocketCAN, I would like to point you to this slides: https://wiki.automotivelinux.org/_media/agl-distro/agl2017-socketcan-print.pdf
The first 10 slides show his ideas about sockets.
> 
>     * Since the select method you implemented is non blocking, the main thread that wants to wait for messages will just poll for non-empty queues as fast as the CPU can handle. Apart from wasting resources, just polling the buses directly with a timeout of 0 instead should work equally well or bad.
Thats true, but I consider different use cases. On Linux sockets, the call to `recv()` is blocking (usually). In order to receive from a socket, you usually call select first, to determine if data is available. Let's talk about a concrete use case: On Windows, I want to create a application with CAN which has two different "application" threads for different purposes. Each application thread is filtering for different CAN messages. I have one CAN interface, which is shared between my applications. With the concept of socket objects, each application thread can create its individual object, with filter settings. Whenever CAN messages arrive, the `SocketThreadPool` is receiving all messages and distributing them to the internal queues of the sockets of a CAN interface. Whenever a application wants to read some CAN messages, it can read the filtered messages from their sockets queue. 
There are multiple other ways, to achieve the same behaviour by implementing notifiers or listeners. Nevertheless I think this approach can simplify the application design for applications which share a CAN interface. 
> 
> 
> Please correct me if there is something I have misinterpreted.

polybassa:
> What do you use the “socket” object for except for obtaining the message?
I'm using the socket as some kind of middle layer between the CAN interface and an application thread. 
> Could it make sense to attach the originating Bus or log reader instance to each Message instead?
Could you please give me an example to this question? I don't get it completely.

christiansandberg:
I just think that naming it a socket makes it easy to confuse it with an actual socket and that you may use it as one. I don’t see it has anything to do with sockets so it would probably be better to call it something else, like a dispatcher or whatever.

How are you using the select method in your case? Where does your application block waiting for something to do?

I was wondering if the reason for a select method was to be able to determine the source of the messages. Otherwise the Notifier and custom Listeners could be used to accomplish the same thing. You can get the originating channel for each message but if you really needed to know the originating bus instance we could add that as well.

polybassa:
Hi Christian,

> On 10. Jan 2020, at 20:07, Christian Sandberg <notifications@github.com> wrote:
> 
> I just think that naming it a socket makes it easy to confuse it with an actual socket and that you may use it as one. I don’t see it has anything to do with sockets so it would probably be better to call it something else, like a dispatcher or whatever.
> 
Could you please explain me, why you don’t see it as a socket. What is the behavior/API of a socket, in your opinion?
> How are you using the select method in your case? Where does your application block waiting for something to do?
> 
I use the select to have my application not block waiting at all. Through the select, an application can determine if data is present in the sockets queue. Only if data is present (select returns readable sockets) recv will be called on this sockets.
> I was wondering if the reason for a select method was to be able to determine the source of the messages.
> 
This is not the case. This can’t be done on a real CAN bus either. I don’t want to implement non standard additions to CAN messages
> Otherwise the Notifier and custom Listeners could be used to accomplish the same thing. You can get the originating channel for each message but if you really needed to know the originating bus instance we could add that as well.
> 
Yes, that might be possible.

Again, my use case is to transparently share one driver instance of a CAN device (in python-can a Bus object) with multiple applications in a thread safe way (individual endpoints).

Thank you for this discussion.

christiansandberg:
A socket to me is something I can pass to the built in select.select() function or any other API that accepts socket-like objects. This will obviously not work because all sockets must have a file descriptor and your Socket class does not.

An application that does not block at any time will potentially use up all CPU resources while not doing anything at all. You could add some sleep time between the calls to your select method but you would always have to balance CPU usage and responsiveness. That’s why it’s best to rely on the recv() method of the bus to wait for a message and start processing it immediately, either in the same thread or waking up some other thread that is blocked waiting for a message.

polybassa:
Hi Christian,
thanks for your answer.
I agree on your opinion regarding the socket API. This is a fundamental difference and could lead to misinterpretation of the code. 

What terminology do you suggest?
Do you see a chance to get this addition into python-can?

polybassa:
I will implement this feature in scapy.

christiansandberg:
Thank you for your contribution anyway. Even if it would be useful for someone else, there is always a difficult balance for how many different ways of receiving messages we should provide without sacrificing simplicity. There are probably many more smart solutions out there so there will always be the question of where to draw the line. I hope you understand.

polybassa:
I fully understand. Thanks for the discussions we had. I changed the code to only act as a multiplexer, now. In this way, I could remove all threads. If you are interested, this is the current version: 
https://github.com/polybassa/scapy-1/blob/python-can-refactoring/scapy/contrib/cansocket_python_can.py

PR #754, opened 2020-01-07
Fix date format to show correct day of month

bessman:
ASCWriter currently writes the month twice (once as three-letter abbreviation and once as  zero-padded decimal number), but does not write the day of month. I don't think this is correct. For example, today's date (2020-01-07) is currently written Tue Jan 01 01:14:48.625694 PM 2020.

With this fix, today's date will instead be written Tue Jan 07 01:14:48.625694 PM 2020.


bessman:
For reference, CANalyzer writes today's date in .asc-files thus:
Tue Jan 07 13:14:48.626 am 2020

PR #760, opened 2020-01-28
Fixed Slcan fileno raising an exception on windows

bmeisels:
Slcan currently raises an exception when used with 'can.Notifier' on windows.

'can.Notifier.add_bus' calls 'bus.fileno' which calls 'serialPortOrig.fileno' which raises an exception on windows as it is not implemented.

The fix catches the exception raised by calling fileno instead of checking if it exists(as it can exist and still raise an error).


bessman:
Better specify the error type. Naked excepts can hide bugs.

can/interfaces/slcan.py
```diff
@@ -244,8 +244,10 @@ def shutdown(self):
         self.serialPortOrig.close()
 
     def fileno(self):
-        if hasattr(self.serialPortOrig, "fileno"):
+        try:
             return self.serialPortOrig.fileno()
+        except:
```
felixdivo: @bessman is right, we could for example use
```suggestion
        except (NotImplementedError, AttributeError):
```

bmeisels: The issue is that on windows fileno exists.
the pyserial code raises an exception.
the serial object inherits from io.IOBase which defines a fileno method.
according to the docs (https://docs.python.org/3/library/io.html#io.IOBase.fileno) it can raise an OSError. In practice in CPython io.UnsupportedOperation (https://docs.python.org/3/library/io.html#io.UnsupportedOperation) is raised.

Which of these should we catch?

bessman: In general, you want to be as specific as possible when catching exceptions. `except OSError:` would catch UnsupportedOperation, but it would also catch any other error which inherits from OSError, which we do not intend. Therefore, `except UnsupportedOperation:` is preferable.

bmeisels: @bessman the issue i have with catching UnsupportedOperation is that according to the documentation it could be any derivative of OSError.

felixdivo: I think UnsupportedOperation sound good.

bmeisels:
@felixdivo @bessman I made the changes. What do you think?

can/interfaces/slcan.py
```diff
@@ -244,9 +245,12 @@ def shutdown(self):
         self.serialPortOrig.close()
 
     def fileno(self):
-        if hasattr(self.serialPortOrig, "fileno"):
+        try:
             return self.serialPortOrig.fileno()
-        # Return an invalid file descriptor on Windows
+        except (NotImplementedError, AttributeError, io.UnsupportedOperation):
```
bessman: Can NotImplementedError or AttributeError actually be raised here? If not we shouldn't try to catch them. Based on the old code it looks like at least AttributeError could be raised. @christiansandberg, do you remember the reason for the `hasattr` check?

christiansandberg: That was probably a mistake from my end. I saw that pyserial did not implement the fileno() method but forgot to check if it was inherited instead.

bessman: In that case, if the only exception that has actually been observed in the wild here is UnsupportedOperation, I suggest only trying to catch that. If it turns out other exceptions (that we want to catch) can be raised here, add them later.

bmeisels:
Changed to only catch io.UnsupportedOperation.

christiansandberg:
I think an even better approach would be to implement the fileno() method in BusABC which would raise UnsupportedOperation. The slcan implementation would not need to handle the exception at all, but the Notifier would do that instead. Returning -1 instead of raising an exception was not very Pythonic I admit.

bmeisels:
@christiansandberg I like the general idea.
I think BusABC.fileno should raise NotImplementedError to be similar to other optional method in BusABC.

bmeisels:
Sorry for the inactivity. I thought we were leaning towards the solution offered by @christiansandberg .

hardbyte:
Ahh right you are. I've opened https://github.com/hardbyte/python-can/pull/877 if you'd like to check it out

PR #741, opened 2019-12-09
Fixing ASCII reader unable to read FD frames

sou1hacker:
CAN FD Frames in an ASCII log file could be logged in the following format:

```
0.045522 CANFD   62 Rx         7d  symbolic_name                    1 0 8  8 01 1b d6 d0 8a 1f d7 f6   104812  132   323000 a800f6c9 46500250 4b140250 20011736 2001040d
```

These frames are currently not read and skipped when using can.LogReader. This PR enables those frames to be read and processed.
Added unit tests for the same.

PR #761, opened 2020-01-29
adding absolute timestamp to ASC reader

cperkulator:
needed timestamps to be absolute so I added this. 

Probably needs some error handling.  Will get to it when I have the chance.

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -48,7 +48,7 @@ class ASCReader(BaseIOHandler):
         :param base: Select the base(hex or dec) of id and data.
                      If the header of the asc file contains base information,
                      this value will be overwritten. Default "hex".
-        :param relative_timestamp: Select whether the timestamps are 
+        :param relative_timestamp: Select whether the timestamps are
                      `relative` (starting at 0.0) or `absolute` (starting at
                      the system time). Default `True = relative`.
         """
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -31,8 +31,6 @@ class ASCReader(BaseIOHandler):
     """
     Iterator of CAN messages from a ASC logging file. Meta data (comments,
     bus statistics, J1939 Transport Protocol messages) is ignored.
-
-    TODO: turn relative timestamps back to absolute form
     """
 
     def __init__(
@@ -185,6 +183,13 @@ class ASCReader(BaseIOHandler):
 
         for line in self.file:
             temp = line.strip()
+            
+            #check for timestamp
+            if "begin triggerblock" in temp.lower():
+                _, _, start_time = temp.split(None, 2)
+                start_time = datetime.strptime(start_time, "%a %b %m %I:%M:%S.%f %p %Y").timestamp()
+                continue
+
             if not temp or not temp[0].isdigit():
                 # Could be a comment
                 continue
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -33,11 +33,13 @@ class ASCReader(BaseIOHandler):
     bus statistics, J1939 Transport Protocol messages) is ignored.
     """
 
+    FORMAT_START_OF_FILE_DATE = "%a %b %d %I:%M:%S.%f %p %Y"
+
     def __init__(
         self,
         file: Union[typechecking.FileLike, typechecking.StringPathLike],
         base: str = "hex",
-        relative: bool = True,
+        relative_timestamp: bool = True,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to read from
@@ -46,9 +48,9 @@ class ASCReader(BaseIOHandler):
         :param base: Select the base(hex or dec) of id and data.
                      If the header of the asc file contains base information,
                      this value will be overwritten. Default "hex".
-        :param relative: Select whether the timestamps are `relative` (starting
-                     at 0.0) of `absolute` (starting at the system time). 
-                     Default `True = relative`.
+        :param relative_timestamp: Select whether the timestamps are 
+                     `relative` (starting at 0.0) or `absolute` (starting at
+                     the system time). Default `True = relative`.
         """
         super().__init__(file, mode="r")
 
@@ -56,7 +58,7 @@ class ASCReader(BaseIOHandler):
             raise ValueError("The given file cannot be None")
         self.base = base
         self._converted_base = self._check_base(base)
-        self.relative_timestamp = relative
+        self.relative_timestamp = relative_timestamp
         self.date = None
         # TODO - what is this used for? The ASC Writer only prints `absolute`
         self.timestamps_format = None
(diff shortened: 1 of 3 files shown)


can/io/asc.py
```diff
@@ -49,10 +47,18 @@ def _extract_can_id(str_can_id):
...
+        start_time = 0.0
         for line in self.file:
             # logger.debug("ASCReader: parsing line: '%s'", line.splitlines()[0])
 
             temp = line.strip()
+            
+            #check for timestamp
+            if "begin triggerblock" in temp.lower():
+                _, _, start_time = temp.split(None, 2)
+                start_time = datetime.strptime(start_time, "%a %b %m %I:%M:%S.%f %p %Y").timestamp()
```
christiansandberg: What happens if the parsing fails? Should we catch some exception?

can/io/asc.py
```diff
@@ -49,10 +47,18 @@ def _extract_can_id(str_can_id):
         return can_id, is_extended
 
     def __iter__(self):
+        start_time = 0.0
```
bessman:
```suggestion
```

can/io/asc.py
```diff
@@ -49,10 +47,18 @@ def _extract_can_id(str_can_id):
...
         for line in self.file:
             # logger.debug("ASCReader: parsing line: '%s'", line.splitlines()[0])
 
             temp = line.strip()
+            
+            #check for timestamp
+            if "begin triggerblock" in temp.lower():
+                _, _, start_time = temp.split(None, 2)
```
bessman:
```suggestion
                try:
                    _, _, start_time = temp.split(None, 2)
                    start_time = datetime.strptime(start_time, "%a %b %m %I:%M:%S.%f %p %Y").timestamp()
                except ValueError:
                    start_time = None
```

can/io/asc.py
bessman:
```suggestion

```

can/io/asc.py
```diff
@@ -62,7 +68,7 @@ def __iter__(self):
             except ValueError:
                 # we parsed an empty comment
                 continue
-            timestamp = float(timestamp)
+            timestamp = float(timestamp) + start_time
```
bessman:
```suggestion
            timestamp = float(timestamp) + (start_time if start_time is not None else 0)
```

can/io/asc.py
bessman: Both of these lines could raise ValueError, if `temp` contains fewer than two whitespaces, or if it does not match the specified date format.

felixdivo: One idea would be to raise an error here, and maybe let the user suppress it via a param to the constructor.

felixdivo:
This might need a rebase due to #741 having been merged two days ago.

can/io/asc.py
christiansandberg: I think we could fall back to relative timestamps since absolute timestamps is not guaranteed in either log files or buses.

felixdivo: > I think we could fall back to relative timestamps since absolute timestamps is not guaranteed in either log files or buses.

Yeah, but maybe print a warning along the way.

sou1hacker:
Was wondering if we can make this optional by a flag passed to LogReader? In most of our use cases, we would actually prefer the current timestamp format which is easier to read and co-relate back with the ascii logs file. Maybe even make the existing implementation the default approach?

felixdivo:
@cperkulator Do you plan on having another look at this?

cperkulator:
> @cperkulator Do you plan on having another look at this?

oops I really apologize. I'll get this rebased and try and wrap it up today/

can/io/asc.py
```diff
@@ -47,14 +46,19 @@ def __init__(
         :param base: Select the base(hex or dec) of id and data.
                      If the header of the asc file contains base information,
                      this value will be overwritten. Default "hex".
+        :param relative: Select whether the timestamps are `relative` (starting
+                     at 0.0) of `absolute` (starting at the system time).
```
felixdivo, line 50:
```suggestion
                     at 0.0) or `absolute` (starting at the system time).
```

can/io/asc.py
```diff
@@ -47,14 +46,19 @@ def __init__(
...
+                     Default `True = relative`.
         """
         super().__init__(file, mode="r")
 
         if not self.file:
             raise ValueError("The given file cannot be None")
         self.base = base
         self._converted_base = self._check_base(base)
+        self.relative_timestamp = relative
```
felixdivo, line 59: I'd also name the argument `relative_timestamp` to keep it consistent with the attribute.

felixdivo (changes requested):
Would you mind adding one test for each of the modes `relative` and `absolute`? That would make your PR really great!

test/data/test_CanMessage.asc
```diff
@@ -2,7 +2,7 @@ date Sam Sep 30 15:06:13.191 2017
 base hex  timestamps absolute
 internal events logged
 // version 9.0.0
-Begin Triggerblock Sam Sep 30 15:06:13.191 2017
+Begin Triggerblock Sat Sep 30 10:06:13.191 PM 2017
```
cperkulator, line 5: looks like the tests files weren't made with the `ASCWriter` so I updated to make it right. Doesn't appear to mess with any tests.

test/logformats_test.py
```diff
@@ -374,11 +377,39 @@ def _setup_instance(self):
...
         )
 
-    def _read_log_file(self, filename):
+    def _read_log_file(self, filename, **kwargs):
         logfile = os.path.join(os.path.dirname(__file__), "data", filename)
-        with can.ASCReader(logfile) as reader:
+        with can.ASCReader(logfile, **kwargs) as reader:
             return list(reader)
 
+    def test_absolute_time(self):
```
cperkulator, line 385: @felixdivo Added just one test for absolute timestamp since relative is tested by the existing tests

test/data/test_CanMessage.asc
felixdivo: Looks good. Looks like it was generated with a German locale.

test/logformats_test.py
felixdivo: Cool! That should suffice.

felixdivo (approved):

felixdivo:
@Mergifyio update

cperkulator:
Not sure what is up with codecov but according to tox, everything I added is covered.

felixdivo:
> Not sure what is up with codecov but according to tox, everything I added is covered.

Yeah, just ignore it in this case. It's somewhat broken.

felixdivo:
@cperkulator Somehow, there seem to be sudden errors on the develop branch. See [here](https://github.com/hardbyte/python-can/actions/runs/779056159). Do you know what's happening? They are probably related to this PR, right?

cperkulator:
> @cperkulator Somehow, there seem to be sudden errors on the develop branch. See [here](https://github.com/hardbyte/python-can/actions/runs/779056159). Do you know what's happening? They are probably related to this PR, right?

They all passed before the merge right? I'll take a look.

cperkulator:
> > @cperkulator Somehow, there seem to be sudden errors on the develop branch. See [here](https://github.com/hardbyte/python-can/actions/runs/779056159). Do you know what's happening? They are probably related to this PR, right?
> 
> They all passed before the merge right? I'll take a look.

@felixdivo alright figured it out.  Those failing tests create a header in the temporary ASC file:
image

Since this time is before 1970 (and thus not representable by POSIX time) then we get an `OSError` exception.

I'll submit a new PR to catch that error

felixdivo:
Okay, sounds plausible. Thanks for figuring it out!

PR #762, opened 2020-02-05
'can.interface' NOT 'can.interfaces.interface'

auden-rovellequartz:
I was getting the following error (see below)

####################################################################
C:\Users\arovellequartz>python auden_scratchpad_002.py
  File "auden_scratchpad_002.py", line 18, in <module>
    from can.interfaces.interface import Bus
ModuleNotFoundError: No module named 'can.interfaces.interface'
####################################################################


felixdivo:
This looks good to me, but I'm not sure enough to merge this.

PR #764, opened 2020-02-06
Supports other base number(radix) at ASCReader.

typecprint:
This pull request changes the radix of data that can be handled by ASCReader.

The asc file that I am dealing with is made in decimal, so I needed to improve ASCReader. By using base as an argument of the constructor, other than hexadecimal numbers are supported.

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -30,22 +30,24 @@ class ASCReader(BaseIOHandler):
     TODO: turn relative timestamps back to absolute form
     """
 
-    def __init__(self, file):
+    def __init__(self, file, base=16):
         """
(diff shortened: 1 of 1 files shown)


christiansandberg:
It would be better if you parsed the base from header. I think it should say “base hex” or “base dec”.

typecprint:
@christiansandberg 
I thought about taking information from the header, but there are many opportunities to analyze only the middle lines. So I want to be able to choose the base. However, I think it is sufficient to support only hex and dec.

typecprint:
The function to read the base information from the header and the function to pass the base have been implemented to be compatible.

PR #766, opened 2020-02-11
Clean up socketcan

felixdivo:
Only small stuff: Adds documentation, simplifications and better error handling.

Should we keep the change to the coverage config? It allows the coverage artifacts to be viewed/downloaded [here](https://codecov.io/gh/hardbyte/python-can/commit/6de3d92787f56339cdf46da1d1637e0ea25fceb7/build).


can/interfaces/socketcan/utils.py
```diff
@@ -68,24 +65,15 @@ def find_available_interfaces() -> Iterable[str]:
...
     """
     Converts a given error code (errno) to a useful and human readable string.
 
-    :param int code: a possibly invalid/unknown error code
-    :rtype: str
+    :param code: a possibly invalid/unknown error code
     :returns: a string explaining and containing the given error code, or a string
               explaining that the errorcode is unknown if that is the case
     """
+    name = errno.errorcode.get(code, "UNKNOWN") if code is not None else "UNKNOWN"
```
karlding, line 76: Doesn't `get` already return the default when the key isn't in the dictionary? Something like:

```suggestion
    name = errno.errorcode.get(code, "UNKNOWN")
```

can/interfaces/socketcan/socketcan.py
```diff
@@ -595,18 +595,25 @@ def __init__(
...
+            An example channel would be 'vcan0' or 'can0'.
             An empty string '' will receive messages from all channels.
             In that case any sent messages must be explicitly addressed to a
             channel using :attr:`can.Message.channel`.
         :param receive_own_messages:
             If transmitted messages should also be received by this bus.
         :param fd:
             If CAN-FD frames should be supported.
-        :param list can_filters:
+        :param List can_filters:
```
karlding, line 616: I guess we might as well add this to the function declaration? Also I think there's a type  for filters in `typechecking.py` that can be extended, if it isn't sufficient.

can/interfaces/socketcan/socketcan.py
```diff
@@ -595,18 +595,25 @@ def __init__(
         fd: bool = False,
         **kwargs,
     ) -> None:
-        """
+        """Creates a new socketcan bus.
+
+        It setting some socket options fails, an error will be printed but no exception will be thrown.
```
karlding, line 600:
```suggestion
        If setting some socket options fails, an error will be printed but no exception will be thrown.
```

.codecov.yml
```diff
@@ -1,7 +1,7 @@
 # Validate with curl --data-binary @.codecov.yml https://codecov.io/validate
 codecov:
   archive:
-    uploads: no
+    uploads: yes
```
karlding, line 4: Probably separate this out into its own change?

felixdivo: Will do.

can/interfaces/socketcan/utils.py
felixdivo: Yep, that's right. But I added it for the type checker: It will complain when a `Mapping[int]` is queried with an `Optional[int]`. So either keep it as is or cast/ignore the type. And I thought this was the better approach. Probably should add a comment if that convinces you? ;)

.codecov.yml
felixdivo: Done in #825.

can/interfaces/socketcan/utils.py
karlding: Nit: I asked over at python/typing, and it seems like adding `type: ignore` is the suggested way to go about this.

But why is `None` even allowed as input? Doesn't this function just take an `int` and consider it an error if anything else is passed to avoid this in the first place? The test `test_error_code_to_str` seems to assume this...

felixdivo: I thought it might be useful where one is not sure whether an errno is present or not, though it is currently not used. Guess I'll add the `type: ignore` and change the test.

hardbyte:
:eyes:

PR #773, opened 2020-02-20
Changes to add direction to CAN messages

sou1hacker:
Adds direction to can.Message with the bool attribute is_rx
Adds corresponding changes to ASC & BLF io


christiansandberg:
Thanks for your contribution. I’ve been thinking that this would be a nice feature. There are some more places in the Message class where this needs to be added as well. And the documentation should be updated.

sou1hacker:
Added documentation and copy/deepcopy changes for message class

@christiansandberg Let me know if any more changes are needed. I deliberately left out checking Rx/Tx in the equals method for now.

christiansandberg:
You can also add it to __repr__().

We have been discussing a lot of what constitutes two equal messages. In the end we decided the default should be to check all attributes because that leads to less surprises. In some cases it is however possible to relax this by providing arguments to equals() and that seems appropriate for this since you might want to consider self reception of a sent message as being identical.

PR #774, opened 2020-02-20
add vector hardware config popup

zariiii9003:


can/interfaces/vector/canlib.py
```diff
@@ -577,6 +577,15 @@ def _detect_available_configs():
             )
         return configs
 
+    @staticmethod
+    def popup_vector_hw_configuration(wait_for_finish: int = 0) -> None:
+        """Open vector hardware configuration window.
+
+        :param int wait_for_finish:
+            Time to wait for user input.
```
christiansandberg: Specify that it should be in milliseconds.

zariiii9003: done

zariiii9003:
@christiansandberg Can we merge this?

PR #779, opened 2020-02-24
Adding direction to virutal bus messages

sou1hacker:
This adds can.Message direction (Rx/Tx) to virtual interface

### DIFF
--- a/can/interfaces/virtual.py
+++ b/can/interfaces/virtual.py
@@ -94,9 +94,7 @@ class VirtualBus(BusABC):
             msg_copy = deepcopy(msg)
             msg_copy.timestamp = timestamp
             msg_copy.channel = self.channel_id
-            msg_copy.is_rx = True
-            if bus_queue is self.queue:
-                msg_copy.is_rx = False
+            msg_copy.is_rx = bus_queue is not self.queue
             try:
                 bus_queue.put(msg_copy, block=True, timeout=timeout)
             except queue.Full:
--- a/can/interfaces/virtual.py
+++ b/can/interfaces/virtual.py
@@ -88,13 +88,18 @@ class VirtualBus(BusABC):
         msg_copy = deepcopy(msg)
         msg_copy.timestamp = time.time()
         msg_copy.channel = self.channel_id
+        msg_copy.is_rx = True
 
         # Add message to all listening on this channel
         all_sent = True
         for bus_queue in self.channel:
+            msg_to_send = msg_copy
+            if bus_queue is self.queue and self.receive_own_messages:
+                msg_to_send = deepcopy(msg_copy)
+                msg_to_send.is_rx = False
             if bus_queue is not self.queue or self.receive_own_messages:
                 try:
-                    bus_queue.put(msg_copy, block=True, timeout=timeout)
+                    bus_queue.put(msg_to_send, block=True, timeout=timeout)
                 except queue.Full:
                     all_sent = False
         if not all_sent:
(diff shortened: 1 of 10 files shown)


can/interfaces/virtual.py
```diff
@@ -88,13 +88,18 @@ def send(self, msg, timeout=None):
...
         msg_copy.timestamp = time.time()
         msg_copy.channel = self.channel_id
+        msg_copy.is_rx = True
 
         # Add message to all listening on this channel
         all_sent = True
         for bus_queue in self.channel:
+            msg_to_send = msg_copy
+            if bus_queue is self.queue and self.receive_own_messages:
+                msg_to_send = deepcopy(msg_copy)
```
christiansandberg: A regular copy would suffice here.

sou1hacker: @christiansandberg Isn't deepcopy better than copy? I mean doing msg.data = xx after receiving a message in one queue will modify the message for all other virtual interfaces. 

Also the original implementation sends the same message instance on every virtual bus. Should we change that to send a different instance since the same argument applies there.

christiansandberg:
Yes I know we pass the same msg instance to all interfaces (mostly for performance reasons) and that could be a risk if one of them mutates it for some reason. In general I think a message should be seen as immutable and especially the data part but perhaps it is better to deepcopy all messages instead just to be sure.

sou1hacker:
@christiansandberg  changed implementation to create a new deep copied instance of message for each virtual bus queue

can/interfaces/virtual.py
```diff
@@ -85,18 +85,22 @@ def _recv_internal(self, timeout):
...
-                except queue.Full:
-                    all_sent = False
+            if bus_queue is self.queue and not self.receive_own_messages:
+                continue
+            msg_copy = deepcopy(msg)
+            msg_copy.timestamp = timestamp
+            msg_copy.channel = self.channel_id
+            msg_copy.is_rx = True
+            if bus_queue is self.queue:
+                msg_copy.is_rx = False
```
pierreluctg:
```suggestion
            msg_copy.is_rx = bus_queue is not self.queue
```

pierreluctg:
Future improvement can be to implement a copy-on-write (CoW) for messages.

hardbyte:
Perhaps worth adding `is_rx` as an optional attribute to the module level `Message` class to coordinate with other backends.

pierreluctg:
@sou1hacker can you please add unit tests to cover the new features?

1. Test that is_rx flag is correctly set for virtual buses
1. Test that each queue receive it's own copy of the message

Thank you

test/back2back_test.py
```diff
@@ -125,6 +125,66 @@ def test_dlc_less_than_eight(self):
...
+                is_extended_id=False, arbitration_id=0x300, data=[2, 1, 3]
+            )
+            bus3.send(msg)
+            recv_msg_bus1 = self.bus1.recv(self.TIMEOUT)
+            recv_msg_bus2 = self.bus2.recv(self.TIMEOUT)
+            self_recv_msg_bus3 = bus3.recv(self.TIMEOUT)
+
+            self.assertEqual(recv_msg_bus1.is_rx, True)
+            self.assertEqual(recv_msg_bus2.is_rx, True)
+            self.assertEqual(self_recv_msg_bus3.is_rx, False)
```
pierreluctg: Please simplify by using assertTrue/assertFalse

```suggestion
            self.assertTrue(recv_msg_bus1.is_rx)
            self.assertTrue(recv_msg_bus2.is_rx)
            self.assertFalse(self_recv_msg_bus3.is_rx)
```

Issue #776, opened 2020-02-21
Supporting Tx/Rx message direction for each interface


sou1hacker:
Pull request #773 adds is_rx attribute to can.Message. Currently this attribute is set to True by default. Functionality needs to be added to each interface type to set this attribute appropriately when building CAN messages:

- [x] ics_neovi
- [ ] ixxat
- [ ] kvaser
- [ ] pcan
- [ ] seedstudio
- [ ] serial
- [ ] socketcan
- [ ] systec
- [ ] usb2can
- [ ] vector
- [ ] nican
- [ ] iscan
- [ ] virtual
- [ ] slcan
- [ ] robotell
- [ ] canalystii

bessman:
How should `receive_own_messages` behave with this change? Should messages transmitted by other applications have `is_rx=False`?

I'm looking at Kvaser's canlib, and I can see no obvious way to accomplish that.

tamenol:
I was looking to implement this for the PCAN bus. However I can't seem to find a variable in the PCANbasic library I could use to check if it's RX or TX.  Someone has an idea?

`TPCANMessageType` only has these values:
```python
# PCAN message types
PCAN_MESSAGE_STANDARD = TPCANMessageType(
    0x00
)  # The PCAN message is a CAN Standard Frame (11-bit identifier)
PCAN_MESSAGE_RTR = TPCANMessageType(
    0x01
)  # The PCAN message is a CAN Remote-Transfer-Request Frame
PCAN_MESSAGE_EXTENDED = TPCANMessageType(
    0x02
)  # The PCAN message is a CAN Extended Frame (29-bit identifier)
PCAN_MESSAGE_FD = TPCANMessageType(
    0x04
)  # The PCAN message represents a FD frame in terms of CiA Specs
PCAN_MESSAGE_BRS = TPCANMessageType(
    0x08
)  # The PCAN message represents a FD bit rate switch (CAN data at a higher bit rate)
PCAN_MESSAGE_ESI = TPCANMessageType(
    0x10
)  # The PCAN message represents a FD error state indicator(CAN FD transmitter was error active)
PCAN_MESSAGE_ERRFRAME = TPCANMessageType(
    0x40
)  # The PCAN message represents an error frame
PCAN_MESSAGE_STATUS = TPCANMessageType(
    0x80
)  # The PCAN message represents a PCAN status message
```

chrisoro:
@hardbyte @pierreluctg 
[...]

PR #780, opened 2020-02-26
Add TX/RX message direction to socketcan

tysonite:
References #773 and #776

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -509,10 +509,10 @@ def capture_message(
     # Fetching the Arb ID, DLC and Data
     try:
         if get_channel:
-            cf, addr = sock.recvfrom(CANFD_MTU)
+            cf, _, msg_flags, addr = sock.recvmsg(CANFD_MTU)
             channel = addr[0] if isinstance(addr, tuple) else addr
         else:
-            cf = sock.recv(CANFD_MTU)
+            cf, _, msg_flags, _ = sock.recvmsg(CANFD_MTU)
             channel = None
     except socket.error as exc:
         raise can.CanError("Error receiving: %s" % exc)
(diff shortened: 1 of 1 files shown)


tysonite:
Don't judge me too much, please. :) My first try to submit something there. I actually got a pointer to implement that looking into [candump tool](https://github.com/linux-can/can-utils/blob/da65fdfe0d1986625ee00af0b56ae17ec132e700/candump.c#L778) and [SocketCAN doc](https://www.kernel.org/doc/Documentation/networking/can.txt) 4.1.7 section.

can/interfaces/socketcan/socketcan.py
```diff
@@ -536,6 +536,7 @@ def capture_message(
     is_remote_transmission_request = bool(can_id & CAN_RTR_FLAG)
     is_error_frame = bool(can_id & CAN_ERR_FLAG)
     is_fd = len(cf) == CANFD_MTU
+    is_tx = bool(msg_flags & socket.MSG_DONTROUTE)
```
bessman: I would add a comment explaining the use of this flag:
`# MSG_DONTROUTE: set when the received frame was created on the local host.`

can/interfaces/socketcan/socketcan.py
```diff
@@ -555,6 +556,7 @@ def capture_message(
         is_remote_frame=is_remote_transmission_request,
         is_error_frame=is_error_frame,
         is_fd=is_fd,
+        is_rx=not is_tx,
```
bessman: I think `is_rx` should follow the same pattern as the other parameters, i.e. `is_rx=is_rx`. So move the `not` to L539 instead.

tysonite: Done

can/interfaces/socketcan/socketcan.py
tysonite: Done

Issue #743, opened 2019-12-12
ThreadBasedCyclicSendTask stops when any error happens on CAN bus


tysonite:
I would like to have a discussion over implementation of `ThreadBasedCyclicSendTask`. The issue there is that if any kind of error happens on the CAN bus, it terminate itself:

https://github.com/hardbyte/python-can/blob/b0d74607de0e411ea3b19347cee20cf8ccba0dcd/can/broadcastmanager.py#L251-L253

I see few issues there:
1. The user (application code actually) is not notified if error happens (`bus.send()` fires exception) during sending periodic messages in `ThreadBasedCyclicSendTask`, thus application can't handle it properly, for example, restart periodic task after some time. Do I miss something and is this possible?

2. It contradicts with a behavior of, for example, SocketCAN implementation of `send_periodic` as it moves responsibility of sending periodic messages to the kernel level, and it is expected that error handling will be done inside the kernel. As tests show, if error on CAN bus happens during usage of native periodic implementation, the periodic tasks are still there and it covers after some time.

Any thoughts on this?

christiansandberg:
It could certainly be discussed. I’m not sure how to handle that situation though. Letting the developer decide is probably better than trying to force some recovery mechanism that won’t suit everyone.

tysonite:
I would think about having a callback function passed as an argument to `bus.send_periodic`, it may accept an exception raised internally by `ThreadBasedCyclicSendTask` so application can be implement any custom logic based on that.

tysonite:
I could close this issue as fix is merged already. Or is there anything else in your minds?

PR #781, opened 2020-02-27
Add an error callback to ThreadBasedCyclicSendTask

tysonite:
References #743

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -212,7 +212,7 @@ class ThreadBasedCyclicSendTask(
         """Transmits `messages` with a `period` seconds for `duration` seconds on a `bus`.
 
         The `on_error` is called if any error happens on `bus` while sending `messages`.
-        If `on_error` present, and returns ``True`` when invoked, thread is
+        If `on_error` present, and returns ``False`` when invoked, thread is
         stopped immediately, otherwise, thread continuiously tries to send `messages`
         ignoring errors on a `bus`. Absence of `on_error` means that thread exits immediately
         on error.
@@ -266,7 +266,7 @@ class ThreadBasedCyclicSendTask(
                 except Exception as exc:
                     log.exception(exc)
                     if self.on_error:
-                        if self.on_error(exc) is True:
+                        if not self.on_error(exc):
                             break
                     else:
                         break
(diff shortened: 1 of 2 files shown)


can/broadcastmanager.py
```diff
@@ -250,6 +252,8 @@ def _run(self):
                     self.bus.send(self.messages[msg_index])
                 except Exception as exc:
                     log.exception(exc)
+                    if self.error_callback:
+                        self.error_callback(exc)
```
pierreluctg: Maybe you want to conditionally break on `error_callback` return?

This way, the error handler (`error_callback`) can decide if this is a catastrophic error and you need to stop sending or if this can be recovered automatically.

tysonite: I am not sure. I thought about that. In such case, resources are wasted by the thread responsible to send periodic messages while temporary error persists, personally, I prefer to handle that situation by other means outside of this loop, and restart periodic sender once issue resolves.

However if you want it to work that way, I can make such changes.

tysonite: @pierreluctg I did proposed changes. Can you please have a look?

pierreluctg: looks good to me

can/broadcastmanager.py
```diff
@@ -207,13 +207,28 @@ def __init__(
         messages: Union[Sequence[Message], Message],
         period: float,
         duration: Optional[float] = None,
+        on_error: Optional[Callable[[Exception], bool]] = None,
     ):
+        """Transmits `messages` with a `period` seconds for `duration` seconds on a `bus`.
+
+        The `on_error` is called if any error happens on `bus` while sending `messages`.
+        If `on_error` present, and returns ``True`` when invoked, thread is
```
christiansandberg: We will probably add the same functionality to the Notifier as well, but to keep backwards compatibility it will probably be implemented the other way around, i.e. return True to keep thread alive, otherwise it will stop like today. So to make the behavior consistent I think you should switch too.

can/broadcastmanager.py
```diff
@@ -250,7 +265,11 @@ def _run(self):
                     self.bus.send(self.messages[msg_index])
                 except Exception as exc:
                     log.exception(exc)
-                    break
+                    if self.on_error:
+                        if self.on_error(exc) is True:
```
christiansandberg: You shouldn't make an explicit comparison with booleans. From PEP-8:

```
Yes:   if greeting:
No:    if greeting == True:
Worse: if greeting is True:
```

tysonite: Got it. Updated to `not ...` as per your other comment.

can/broadcastmanager.py
tysonite: Done

tysonite:
Any other comments, guys?

felixdivo:
Sorry @tysonite for the delay

tysonite:
Thanks! Appreciate it

PR #783, opened 2020-02-28
Adding plugin support to can.io Reader/Writer

sou1hacker:
Allows users to plug in their own reader/writer implementation to can.io functions. Similar to how we have plugin support for can interfaces (BACKENDS)

### DIFF
--- a/can/io/logger.py
+++ b/can/io/logger.py
@@ -60,10 +60,12 @@ class Logger(BaseIOHandler, Listener):  # pylint: disable=abstract-method
             ".log": CanutilsLogWriter,
             ".txt": Printer,
         }
-        lookup.update({
-            writer.name: writer.load()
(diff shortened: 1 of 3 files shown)


can/io/logger.py
```diff
@@ -51,17 +62,18 @@ def __new__(
...
-            ".blf": BLFWriter,
-            ".csv": CSVWriter,
-            ".db": SqliteWriter,
-            ".log": CanutilsLogWriter,
-            ".txt": Printer,
-        }
+        if not Logger.fetched_plugins:
+            Logger.message_writers.update(
+                {
+                    writer.name: writer.load()
```
zariiii9003, line 68: @felixdivo @sou1hacker 
This is quite old, but is this `load` method actually documented anywhere?

PR #784, opened 2020-02-29
Update asc.py

wamalbi:
Fixes the bug where the date in start of file header has the month repeated in lieu of day number.
Friday February 26 would log as Fri Feb 2 instead of Fri Feb 26. 
This change fixes that issue.


christiansandberg:
Thanks, but this was already fixed in #754. The develop branch is where you can find the latest code.

Issue #721, opened 2019-11-06
Calling send_periodic several times doesn't work with SocketCan


random-username1:
Hi,
```python3
task = bus.send_periodic(msg, period)
task2 = bus.send_periodic(msg2, period2)
```
After launching several send_periodic tasks, with SocketCan, only the last one appears to be working although the same script works fine with PeakCan on another device.

> ***EDIT***: both messages have the same arbitration ID but different datas

Plugging this other device on the CAN Bus to intercept frames confirms that the last send_periodic task to be started is the only one working.

I'm using python-can 3.3.2 on a 32 bits Raspbian stretch (v9.4)

Here is the output of uname -a:
```
pi@raspberrypi:~ $ uname -a
Linux raspberrypi 4.14.71-v7+ #1145 SMP Fri Sep 21 15:38:35 BST 2018 armv7l GNU/Linux
```

christiansandberg:
Why do you need to send multiple messages with the same CAN-ID? SocketCAN does not support that and should generally never be needed. If it is a multiplexed message then you can provide a list of messages to `send_periodic` instead.

```python
bus.send_periodic([msg, msg2], period)
```

random-username1:
In my case, several CAN messages have the same CAN-ID and the first byte of data is used to 
differentiate them.
These messages have to be sent at different periods so I cannot give a list of messages to the `send_periodic()` method.

Then, wouldn't it be better to just call `bus.send()` in a threaded loop (like it's been done with the `class ThreadBasedCyclicSendTask:` of  python-can's `broadcastmanager.py` file) ?

Also I this this issue is related to #720

christiansandberg:
If you really need to do that then you should probably implement the periodic sending yourself like you say or use ThreadBasedCyclicSendTask directly somehow. The timing will be worse since it is done in Python instead of directly by the kernel so it is best avoided if possible.

karlding:
> Why do you need to send multiple messages with the same CAN-ID? SocketCAN does not support that and should generally never be needed.

[...]

PR #785, opened 2020-03-02
Relax restriction of arbitration ID uniqueness for SocketCAN

tysonite:
References #721 
I am unsure about the correct way to make a unique identifier. Please, suggest.

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -323,6 +323,8 @@ class CyclicSendTask(
     ):
         """
         :param bcm_socket: An open BCM socket on the desired CAN channel.
+        :param task_id:
+            The identifier used to uniquely reference particular cyclic send task.
         :param messages:
             The messages to be sent periodically.
         :param period:
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -7,6 +7,7 @@ At the end of the file the usage of the internal methods is shown.
 
 from typing import Dict, List, Optional, Sequence, Tuple, Type, Union
 
+import threading
 import logging
 import ctypes
 import ctypes.util
@@ -162,7 +163,7 @@ def build_can_frame(msg: Message) -> bytes:
         __u8    data[CANFD_MAX_DLEN] __attribute__((aligned(8)));
     };
     """
-    can_id = _add_flags_to_can_id(msg)
+    can_id = _compose_arbitration_id(msg)
     flags = 0
     if msg.bitrate_switch:
         flags |= CANFD_BRS
@@ -286,7 +287,18 @@ def send_bcm(bcm_socket: socket.socket, data: bytes) -> int:
             raise e
 
 
-def _add_flags_to_can_id(message: Message) -> int:
+GLOBAL_TASK_ID = 0
+TASK_LOCK = threading.Lock()
+
+
+def _compose_next_task_id() -> int:
+    with TASK_LOCK:
+        global GLOBAL_TASK_ID
+        GLOBAL_TASK_ID += 1
+        return GLOBAL_TASK_ID
+
+
+def _compose_arbitration_id(message: Message) -> int:
     can_id = message.arbitration_id
     if message.is_extended_id:
         log.debug("sending an extended id type message")
@@ -297,7 +309,6 @@ def _add_flags_to_can_id(message: Message) -> int:
     if message.is_error_frame:
         log.debug("sending error frame")
         can_id |= CAN_ERR_FLAG
-
     return can_id
 
 
@@ -341,7 +352,7 @@ class CyclicSendTask(
     def _tx_setup(self, messages: Sequence[Message]) -> None:
         # Create a low level packed frame to pass to the kernel
         body = bytearray()
-        self.can_id_with_flags = _add_flags_to_can_id(messages[0])
+        self.task_id = _compose_next_task_id()
         self.flags = CAN_FD_FRAME if messages[0].is_fd else 0
 
         if self.duration:
@@ -353,39 +364,8 @@ class CyclicSendTask(
             ival1 = 0.0
             ival2 = self.period
 
-        # First do a TX_READ before creating a new task, and check if we get
-        # EINVAL. If so, then we are referring to a CAN message with the same
-        # ID
-        check_header = build_bcm_header(
-            opcode=CAN_BCM_TX_READ,
-            flags=0,
-            count=0,
-            ival1_seconds=0,
-            ival1_usec=0,
-            ival2_seconds=0,
-            ival2_usec=0,
-            can_id=self.can_id_with_flags,
-            nframes=0,
-        )
-        try:
-            self.bcm_socket.send(check_header)
-        except OSError as e:
-            if e.errno != errno.EINVAL:
-                raise e
-        else:
-            raise ValueError(
-                "A periodic Task for Arbitration ID {} has already been created".format(
-                    messages[0].arbitration_id
-                )
-            )
-
         header = build_bcm_transmit_header(
-            self.can_id_with_flags,
-            count,
-            ival1,
-            ival2,
-            self.flags,
-            nframes=len(messages),
+            self.task_id, count, ival1, ival2, self.flags, nframes=len(messages)
         )
         for message in messages:
             body += build_can_frame(message)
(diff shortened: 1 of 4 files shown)


can/interfaces/socketcan/socketcan.py
```diff
@@ -286,7 +287,18 @@ def send_bcm(bcm_socket: socket.socket, data: bytes) -> int:
             raise e
 
 
-def _add_flags_to_can_id(message: Message) -> int:
+GLOBAL_TASK_ID = 0
+TASK_LOCK = threading.Lock()
```
karlding: Instead of making this global, perhaps consider storing this on the `Bus` context since we map 1:1 to a SocketCAN bus. This is because the kernel code matches based on the combination of:

1. Task ID (`op->can_id`)
2. Interface (`op->ifindex`)
3. Flags (`op->flags`)

```c
/*
 * helpers for bcm_op handling: find & delete bcm [rx|tx] op elements
 */
static struct bcm_op *bcm_find_op(struct list_head *ops,
				  struct bcm_msg_head *mh, int ifindex)
{
	struct bcm_op *op;

	list_for_each_entry(op, ops, list) {
		if ((op->can_id == mh->can_id) && (op->ifindex == ifindex) &&
		    (op->flags & CAN_FD_FRAME) == (mh->flags & CAN_FD_FRAME))
			return op;
	}

	return NULL;
}
```

(Taken from the [5.5.7 kernel](https://elixir.bootlin.com/linux/v5.5.7/source/net/can/bcm.c#L700))

I think this should be doable if we expand the `socketcan.CyclicTask` to take an ID argument in `__init__`. Thoughts?

can/interfaces/socketcan/socketcan.py
```diff
@@ -286,7 +287,18 @@ def send_bcm(bcm_socket: socket.socket, data: bytes) -> int:
...
+
+
+def _compose_next_task_id() -> int:
+    with TASK_LOCK:
+        global GLOBAL_TASK_ID
+        GLOBAL_TASK_ID += 1
+        return GLOBAL_TASK_ID
```
karlding: We will want to also ensure that we don't overflow the `canid_t` type, on the off-chance that someone is running this for extended periods of time and repeatedly starting/stopping periodic TX tasks.

```c
typedef __u32 canid_t;
```

Ensuring that we wrap at `2^11 - 1` is probably sufficient for now.

test/test_cyclic_socketcan.py
```diff
@@ -244,27 +244,6 @@ def test_cyclic_initializer_different_arbitration_ids(self):
...
-            data=[0x22, 0x22, 0x22, 0x22, 0x22, 0x22],
-            is_extended_id=False,
-        )
-
-        task_a = self._send_bus.send_periodic(messages_a, 1)
-        self.assertIsInstance(task_a, can.broadcastmanager.CyclicSendTaskABC)
-
-        # The second one raises a ValueError when we attempt to create a new
-        # Task, since it has the same arbitration ID.
-        with self.assertRaises(ValueError):
```
karlding: I think this test could probably be repurposed to ensure that we receive both messages, rather than throwing an exception.

can/interfaces/socketcan/socketcan.py
tysonite: Thanks, good catch. It looks to be an unsigned integer though, thus wrapping is required at `2^32 - 1`, isn't it?

Considering this accidental case, some of the previous tasks IDs can be used already, wouldn't it better still implement usage of `CAN_BCM_TX_READ` to check if they are busy or not?

can/interfaces/socketcan/socketcan.py
tysonite: This looks good to me. I am thinking about adding a member function to `SocketcanBus` class that returns next task ID, incrementing old one. The result of that function passed to `CyclicSendTask::__init__` as an additional argument. I will try that.

test/test_cyclic_socketcan.py
tysonite: Modified this test.

can/interfaces/socketcan/socketcan.py
```diff
@@ -734,12 +705,14 @@ def _send_periodic_internal(
...
-        # TODO: The SocketCAN BCM interface treats all cyclic tasks sharing an
-        # Arbitration ID as the same Cyclic group. We should probably warn the
-        # user instead of overwriting the old group?
-        task = CyclicSendTask(bcm_socket, msgs, period, duration)
+        task_id = self._get_next_task_id()
+        task = CyclicSendTask(bcm_socket, task_id, msgs, period, duration)
         return task
 
+    def _get_next_task_id(self) -> int:
+        self._task_id = (self._task_id + 1) % (2 ** 11 - 1)
```
karlding: I believe you still need mutual exclusion around the `task_id` since access to the `Bus` via `send_periodic` isn't serialized. I think we assume that `send_periodic` can be called concurrently.

can/interfaces/socketcan/socketcan.py
karlding: When I commented, I wasn't sure if the kernel code was doing anything with anything beyond the 11-bits (ie. checking other bits for various flags in the 32-bits) so I just said 11-bits to be safe to avoid any potential unintended side effects.

Looking more closely through the kernel code now, it seems like for TX Tasks, this value is only copied (for things other than the TX Task ID) when `TX_CP_CAN_ID` is set as part of `flags` on the BCM message. The only places it's used for determining behaviour is when dealing with RX tasks, but they're stored in separate lists so they won't affect each other (and `python-can` currently doesn't support this).

But yes, you're right, it should be safe to increase this to 32-bits.

karlding: > Considering this accidental case, some of the previous tasks IDs can be used already, wouldn't it better still implement usage of CAN_BCM_TX_READ to check if they are busy or not?

Yes, let's do that. If we detect that the TX ID has been used, let's just raise an Exception, and make it explicit for the user to decide whether to retry calling `bus.send_periodic`.

can/interfaces/socketcan/socketcan.py
tysonite: I am not really sure in that. There are ThreadSafeBus, but I don't really understand whether it serializes access to periodic messages sending. Anyway, I've added a guard.

can/interfaces/socketcan/socketcan.py
tysonite: Done

can/interfaces/socketcan/socketcan.py
```diff
@@ -316,6 +316,7 @@ class CyclicSendTask(
     def __init__(
         self,
         bcm_socket: socket.socket,
+        task_id: int,
```
karlding, line 319: Please make sure a comment is added below for this parameter.

tysonite: Done.

tysonite:
I thought about that too. No problem.

I would be happy to [update](https://python-can.readthedocs.io/en/master/interfaces/socketcan.html#) if you point out what you want to add into existing documentation. Note please that those changes are not user-visible, public API hasn't changed.

But I could add some description that clarifies internals, though. Is that what you want?

hardbyte:
Awesome thanks, clarifying the internal behavior is what I was thinking. But you are now in the best position to check that the docstrings are still accurate e.g. [`socketcan.CyclicSendTask`](https://python-can.readthedocs.io/en/master/interfaces/socketcan.html#can.interfaces.socketcan.CyclicSendTask).

tysonite:
I did few more changes to Sphinx docs if you don't mind along with other updates to docstrings.

doc/interfaces/socketcan.rst
```diff
@@ -1,16 +1,26 @@
...
 
-The full documentation for socketcan can be found in the kernel docs at
-`networking/can.txt <https://www.kernel.org/doc/Documentation/networking/can.txt>`_.
-
-
-.. note::
-
-    Versions before 2.2 had two different implementations named
+The `SocketCAN`_ documentation can be found in the Linux kernel docs at
+``networking`` directory. Hereafter is quote from SocketCAN Linux document::
```
hardbyte, line 5:
```suggestion
``networking`` directory. Quoting from the SocketCAN Linux documentation::
```

tysonite:
@karlding, @hardbyte, thank you for review and letting that in.

Issue #786, opened 2020-03-04
BLF padding/alignment incorrect for CAN_FD_MESSAGE_64


MartinStellmacher:
In blf.py all messages are padded by obj_size % 4. This assumption doesn't hold for all types of messages. I experienced a parsing error while reading a file which contained CAN_FD_MESSAGE_64 items. They are not padded at all, as you can see in Toby Lorenz C++ implementation.

Issue #459 is closed, but the problem still persists. Solution might be as simple as:

                    if header[4] == CAN_FD_MESSAGE_64 or header[4] == ETHERNET_FRAME_EX: 
                        next_pos = pos + obj_size #no padding
                    else:
                        next_pos = pos + obj_size + (obj_size % 4)  #weird padding

In Toby Lorenz implementation the padding is handled separately for each frame type...

christiansandberg:
This fix (for FD frames only) is not released, you need to use develop.

JuicyPussy:
faced the same problem, how could I contribute to fix it?

christiansandberg:
A log file would be helpful. You can e-mail it to me if you don’t want to share it publicly: christiansandberg at me dot com.

MartinStellmacher:
Hi Christian, thank you, but I think this fix isn't enough. In the BLF files I have at hand I found three different kinds of padding: 
1. none, like for the CAN_FD, 
2. crazy padding with `length%4` (mostly used) and 
3. correct padding with `(4-length%4)%4` (I think these were Ethernet-Frames). Even if you don't wan't to read them, your parser will fail, if they are contained within the file.

My final solution is to assume no padding and correct it by searching the LOBJ-mark with offsets from 0-3. Doing it this way I can separate the container format completely from the payload formats. I couldn't measure any increase in parsing times.

christiansandberg:
Yeah that’s probably the best solution. Just need to make sure it handles padding spanning over two containers.

MartinStellmacher:
Yes, I would do the check before reading the header, after fixing the start again for the header and than again to ensure that the payload is in the same container.

christiansandberg:
If you could make a pull request with your solution that would be awesome.

PR #790, opened 2020-03-07
Replaced assumption about padding by just searching for the start of …

MartinStellmacher:
…the next frame by the help of the 'LOBJ' mark.

#786

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -220,6 +220,14 @@ class BLFReader(BaseIOHandler):
         # Loop until a struct unpack raises an exception
         while True:
             self._pos = pos
+            #fix padding
+            if pos + obj_header_base_size > max_pos:
+                # This object continues in the next container
+                return
+            for i in range(4):
+                if data[pos+i:pos+i+4] == b"LOBJ":
+                    break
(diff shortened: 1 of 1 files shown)


can/io/blf.py
```diff
@@ -220,6 +220,14 @@ def _parse_data(self, data):
...
         while True:
             self._pos = pos
+            #fix padding
+            if pos + obj_header_base_size > max_pos:
+                # This object continues in the next container
+                return
+            for i in range(4):
+                if data[pos+i:pos+i+4] == b"LOBJ":
+                    break
+            pos += i
```
christiansandberg: I would suggest using the `index` method instead since it is implemented in optimized C. It should also handle the case when LOBJ is split up in two containers.

```suggestion
            # Find next object after padding (depends on object type)
            try:
                pos = data.index(b"LOBJ", pos, pos + 8)
            except ValueError:
                if pos + 8 > max_pos:
                    # Not enough data in container
                    return
                raise BLFParseError("Could not find next object")
```

You would then also need to change line 178 to remove `memoryview` but that does not seem affect performance much.

PR #792, opened 2020-03-12
Add CAN filtering example

aaknitt:


bessman:
Since the filtering API is shared between all interfaces, this piece of documentation would fit better in either the [set_filters docstring](https://github.com/hardbyte/python-can/blob/c001ef274316b015f7b068414a541a573ce4ef5c/can/bus.py#L320) or in https://github.com/hardbyte/python-can/blob/develop/doc/bus.rst#filtering. I would suggest the former, since it is currently more detailed.

As a side note, perhaps the "can_mask" key should be optional like the "extended" key, with default values of 0x7FF or 0x1FFFFFFF depending on the value of "extended".

aaknitt:
> Since the filtering API is shared between all interfaces, this piece of documentation would fit better in either the [set_filters docstring](https://github.com/hardbyte/python-can/blob/c001ef274316b015f7b068414a541a573ce4ef5c/can/bus.py#L320) or in https://github.com/hardbyte/python-can/blob/develop/doc/bus.rst#filtering. I would suggest the former, since it is currently more detailed.
> 
> As a side note, perhaps the "can_mask" key should be optional like the "extended" key, with default values of 0x7FF or 0x1FFFFFFF depending on the value of "extended".

Makes sense.  I created #795 to put it into /bus.rst.  I happened to come across that Filtering section in the docs in the socketcan module section...wonder if it would make sense to remove it from there or add a link to the Filtering section in /bus.rst.

Default values for can_mask probably also makes sense and would make life a lot easier for simple implementations that are just doing one CAN ID per filter.  I'll look into a PR for that.

PR #806, opened 2020-04-10
Fix BLFParseError

### LINKED ISSUE #786: BLF padding/alignment incorrect for CAN_FD_MESSAGE_64

### LINKED ISSUE #803: BLF Parse error
hello, I get a BLF parse error when opening a normal BLF file, is there some workaround? can I email the file to somebody to check out?

christiansandberg:
Seek for start of object instead of calculating it

Fixes #803
Fixes #786

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -175,7 +175,7 @@ class BLFReader(BaseIOHandler):
 
             if obj_type == LOG_CONTAINER:
                 method, uncompressed_size = LOG_CONTAINER_STRUCT.unpack_from(obj_data)
-                container_data = memoryview(obj_data)[LOG_CONTAINER_STRUCT.size :]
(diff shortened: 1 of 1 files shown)


hardbyte:
Assume merging this closes #790 @christiansandberg ?

PR #811, opened 2020-04-17
Add _detect_available_configs to serial_can

josko7452:
Hi I am using serial_can interface and I was missing the serial port detection.

This is very simple PR, it just adds pySerial `list_ports.comports()` call non-intrusive way. 
I.e. if the detection fails (or `serial.tools.list_ports` is not available) for one reason or another it just gives an empty list.
It does not attempt to get more detailed configuration (e.g. baudrates) besides the list of serial ports.

Tested on Linux and Windows:
Linux:
```
>>> can.detect_available_configs('serial')
[{'interface': 'serial', 'channel': '/dev/ttyUSB0'}]
```
Window:
```
>>> can.detect_available_configs('serial')
[{'interface': 'serial', 'channel': 'COM4'}]
```
As you'd expect. These configs can be directly used to connect to the interface (with the default baudrate of the serial interface).

### DIFF
--- a/can/interfaces/serial/serial_can.py
+++ b/can/interfaces/serial/serial_can.py
@@ -21,6 +21,11 @@ except ImportError:
     )
     serial = None
 
+try:
+    from serial.tools import list_ports
+except ImportError:
+    list_ports = None
+
 
 class SerialBus(BusABC):
     """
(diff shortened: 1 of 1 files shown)


felixdivo:
I'll just wait until the CI tests pass and merge it afterwards.

PR #815, opened 2020-04-20
Fix two typos in the BufferedReader docstring

leventerevesz:


felixdivo:
Thanks for your PR. However, the second "iff" is intentional as it stands for ["if and only if"](https://en.wiktionary.org/wiki/iff). We can also delete it if it confuses people I guess, since it is mostly (only?) used in mathematical contexts. I have no strong feelings about this. @hardbyte What do you think?

hardbyte:
Yeah lets aim to be more inclusive in our language :-)

PR #817, opened 2020-04-22
Asc IO rework + enhancements

sou1hacker:
Changed existing implementation for AscReader to rely more heavily on regex
Added support for CANFD remote frames
Enhanced/fixed parsing on some frames to support new ASC format versions


sou1hacker:
This implementation while cleaner is 30% slower

hardbyte:
That's a shame! Thanks all the same

PR #818, opened 2020-04-23
Add typing annotations for ASC module

karlding:

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -6,6 +6,9 @@ Example .asc files:
     - under `test/data/logfile.asc`
 """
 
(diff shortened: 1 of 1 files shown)


can/io/asc.py
```diff
@@ -213,24 +232,29 @@ def __init__(self, file, channel=1):
...
         # the last part is written with the timestamp of the first message
         self.header_written = False
-        self.last_timestamp = None
-        self.started = None
+        self.last_timestamp = 0.0
+        self.started = 0.0
 
-    def stop(self):
+    def stop(self) -> None:
+        # This is guaranteed to not be None since we raise ValueError in __init__
```
felixdivo, line 239: Then maybe it should be private?

PR #819, opened 2020-04-23
ASCII header

sou1hacker:
Parses the ASCII file header before parsing the file for messages. 
Extracts the attributes present in the header.


sou1hacker:
Will be commited via https://github.com/hardbyte/python-can/pull/820

PR #820, opened 2020-04-27
ASCII Reader/Writer enhancements

sou1hacker:
### Changes
* Added support for new formats for CAN remote frames
* Reworked and cleaned existing implementation
* Some minor fixes (e.g. dlc != len(data) for FD frames)
* Added parsing ASCII header parsing

### Time analysis for both new & old implementations
We tested with an ASCII logs file with ~ 7 million messages (both classic CAN & FD) across 8 channels.
**Old implementation**: Took an average of 128 seconds across 3 runs
**New implementation**: Took an average of 117 seconds across 3 runs


can/io/asc.py
```diff
@@ -43,120 +43,164 @@ def __init__(self, file, base="hex"):
...
+                    _, base, _, timestamp_format = line.split()
+                except ValueError:
+                    raise Exception("Unsupported header string format: {}".format(line))
+                self.base = base
+                self.timestamps_format = timestamp_format
+            elif lower_case.endswith("internal events logged"):
+                if lower_case.startswith("no"):
+                    self.internal_events_logged = False
+                else:
+                    self.internal_events_logged = True
```
pierreluctg, line 68:
```suggestion
                self.internal_events_logged = not lower_case.startswith("no")

```

can/io/asc.py
```diff
@@ -43,120 +43,160 @@ def __init__(self, file, base="hex"):
...
+            )
+
+            if frame_name_or_brs.isdigit():
+                brs = frame_name_or_brs
+                esi, dlc, data_length, data = rest_of_message.split(None, 3)
+            else:
+                brs, esi, dlc, data_length, data = rest_of_message.split(None, 4)
+
+            self._extract_can_id(can_id_str, msg_kwargs)
+            msg_kwargs["bitrate_switch"] = True if brs == "1" else False
```
pierreluctg, line 157:
```suggestion
            msg_kwargs["bitrate_switch"] = brs == "1"
```

can/io/asc.py
```diff
@@ -43,120 +43,160 @@ def __init__(self, file, base="hex"):
...
+            msg_kwargs["error_state_indicator"] = True if esi == "1" else False
```
pierreluctg, line 158:
```suggestion
            msg_kwargs["error_state_indicator"] = esi == "1"
```

can/io/asc.py
```diff
@@ -43,120 +43,160 @@ def __init__(self, file, base="hex"):
...
+            dlc = int(dlc, self._converted_base)
+            msg_kwargs["dlc"] = dlc
+            data_length = int(data_length)
+
+            # CAN remote Frame
+            msg_kwargs["is_remote_frame"] = True if data_length == 0 else False
```
pierreluctg, line 164:
```suggestion
            msg_kwargs["is_remote_frame"] = data_length == 0
```

can/io/asc.py
```diff
@@ -43,120 +43,160 @@ def __init__(self, file, base="hex"):
...
+                dlc = int(dlc, self._converted_base)
+                msg_kwargs["dlc"] = dlc
+                self._process_data_string(data, dlc, msg_kwargs)
+
+        return Message(**msg_kwargs)
+
+    def _process_fd_can_frame(self, line, msg_kwargs):
+        channel, dir, rest_of_message = line.split(None, 2)
+        msg_kwargs["channel"] = int(channel) - 1
+        msg_kwargs["is_rx"] = False if dir != "Rx" else True
```
pierreluctg, line 135:
```suggestion
        msg_kwargs["is_rx"] = not (dir != "Rx")
```

can/io/asc.py
```diff
@@ -43,120 +43,160 @@ def __init__(self, file, base="hex"):
...
+
+        # CAN error frame
+        if line.strip()[0:10].lower() == "errorframe":
+            # Error Frame
+            msg_kwargs["is_error_frame"] = True
+            msg_found = True
+
+        if not msg_found:
+            abr_id_str, dir, rest_of_message = line.split(None, 2)
+            msg_kwargs["is_rx"] = False if dir != "Rx" else True
```
pierreluctg, line 105:
```suggestion
            msg_kwargs["is_rx"] = not (dir != "Rx")
```

can/io/asc.py
```diff
@@ -43,120 +43,156 @@ def __init__(self, file, base="hex"):
         """
         super().__init__(file, mode="r")
         self.base = base
+        self._converted_base = self._check_base(base)
+        self.date = None
+        self.timestamps_format = None
+        self.internal_events_logged = None
```
karlding, line 49: This doesn't seem to be used anywhere. It seems to just indicate the presence of the following line:

```
internal events logged
```

Can you help me understand why this needs to be exposed?

can/io/asc.py
```diff
@@ -43,120 +43,156 @@ def __init__(self, file, base="hex"):
...
+            line = line.strip()
+            lower_case = line.lower()
+            if lower_case.startswith("date"):
+                self.date = line[5:]
+            elif lower_case.startswith("base"):
+                try:
+                    _, base, _, timestamp_format = line.split()
+                except ValueError:
+                    raise Exception("Unsupported header string format: {}".format(line))
+                self.base = base
```
karlding, line 62: The `__init__` takes a given `base`, whereas here right after that we're potentially overwriting it. What happens if the constructor specified `hex`, but the file is `dec`?

If you were intending to use this value from the `.asc` to validate, I don't think you should've overwritten `self.base` on this line.

karlding:
Out of curiosity, do you have a sample file with some of the missing functionality that the current samples don't have? For example, I don't think the current ASC has any examples of error frames (and I think RTR frames? not sure if there's anything else). If so, could you include them?

It would be nice to have specific tests for each type of frame using "golden files" to test parsing :slightly_smiling_face:

can/io/asc.py
sou1hacker: @karlding The line can be either "internal events logged" or "no internal events logged". It may be useful in the future as we add support for more kinds of frames/events in the AscReader.

can/io/asc.py
sou1hacker: @karlding  In both the current and previous implementations, we were ignoring the base passed on to the constructor and using the base defined in the ASCII logs file to parse all the messages. It seems accurate to set base to the value actually used during parsing so I think overwriting it is fine.

sou1hacker:
@karlding Added some examples for CAN / CANFD error, remote frames, etc to the current ASC logs file.

can/io/asc.py
karlding: Hmm, if the idea is for the `ASCReader` to parse the value from the file, then we should probably just deprecate those parameters in `__init__` and raise a `DeprecationWarning` or something.

I don't have access to Vector's tools so I can't check, but are there cases where an ASC file is considered valid but it doesn't specify a base?

karlding:
Thanks for adding the additional frames!

I think it might make test-writing easier if we split out the different frames into their own separate files, similar to what exists for the BLF format? That way we can also write tests that check that the parser is working by asserting on the type of frame that is converted to a `Message`, instead of one giant file (ie. `is_remote_frame`, `is_fd`). What do you think?

can/io/asc.py
sou1hacker: Perhaps the only use case I can think of is when we have a logs file split across two ASCII files. Not completely sure if the header is present in files other than the first file. If the intention is perhaps to read only the second file, then having a manually specified base can be useful. Your call.

sou1hacker:
> Thanks for adding the additional frames!
> 
> I think it might make test-writing easier if we split out the different frames into their own separate files, similar to what exists for the BLF format? That way we can also write tests that check that the parser is working by asserting on the type of frame that is converted to a `Message`, instead of one giant file (ie. `is_remote_frame`, `is_fd`). What do you think?

Will look at BLF and replicate here.

can/io/asc.py
karlding: Let's just leave this for now then. If someone has a use case for this, we can revisit it at that point.

Issue #805, opened 2020-04-02
TestLoggerScript and TestPlayerScript test suites fail


Apteryks:
They are trying to launch "can_logger.py" or "can_player.py --help" (IS_UNIX == True), expecting these are in my PATH, but the program is not yet installed, so that fails.

Are these tests expected to be run after the software has been installed? That'd be weird, making those the only 2 out of 152 to have this requirement.

The error messages look like:
```
                else:
                    err_filename = orig_executable
                if errno_num != 0:
                    err_msg = os.strerror(errno_num)
                    if errno_num == errno.ENOENT:
                        err_msg += ': ' + repr(err_filename)
>               raise child_exception_type(errno_num, err_msg, err_filename)
E               FileNotFoundError: [Errno 2] No such file or directory: 'can_player.py': 'can_player.py'

/gnu/store/78w7y0lxar70j512iqw8x3nimzj10yga-python-3.7.4/lib/python3.7/subprocess.py:1522: FileNotFoundError
```

felixdivo:
This is still the case (on my local machines and the current `develop` branch).

felixdivo:
I can't reproduce this any more.

I did this in Docker (`docker run -it python:3.10 bash`) for the current `develop` branch:

```bash
cd home/
git clone https://github.com/hardbyte/python-can.git
cd python-can/
python setup.py install
pip install tox
tox -e gh  # github actions test suite
```

Only IPv6 stuff of the interface `udp_multicast` fails.

Also, skipping `python setup.py install` works.

In any case:

```
test/test_scripts.py::TestLoggerScript::test_do_commands_exist PASSED                          [ 80%]
test/test_scripts.py::TestLoggerScript::test_does_not_crash PASSED                             [ 81%]
test/test_scripts.py::TestPlayerScript::test_do_commands_exist PASSED                          [ 81%]
test/test_scripts.py::TestPlayerScript::test_does_not_crash PASSED                             [ 81%]
test/test_scripts.py::TestLogconvertScript::test_do_commands_exist PASSED                      [ 81%]
test/test_scripts.py::TestLogconvertScript::test_does_not_crash PASSED                         [ 82%]
```

Thus, this seems to be resolved.

PR #824, opened 2020-05-03
DNI: Add pyproject.toml (and PEP 518 build system) support

karlding:
This adds support for `pyproject.toml`, now that the Status of PEP 518
is `Final`. The idea would be to remove the `setuptools`/`distutils`
files and switch over to `pyproject.toml`. I've been testing these
changes locally with poetry, and things _seem_ to be working correctly,
but I haven't gone through the process of deploying to PyPI (via the
sandbox).

One change here that is perhaps a bit controversial is regarding what
to do about the tests that invoke subprocess to test the installation
of the `can_{logger,player,viewer}` scripts. Currently as-is, these
basically test whether these scripts have been packaged/installed
correctly (as noted in [this issue](https://github.com/hardbyte/python-can/issues/805)). This doesn't _really_ make sense. If we
want to test the output of these scripts, I think we should just
directly invoke the package. I've opted to skip these tests for now
to get CI green.

I can split up these changes into separate PRs if that makes things
easier to review, but am submitting this for now so you can see how
the gist of things will look like (hence the work-in-progress label),
and to provide any comments in general about the approach. The changes
are basically done, minus an issue with poetry that prevents 3.9
nightlies from passing, updating the AppVeyor build, and adding the
deploy section in the Travis CI config.


felixdivo:
I'd agree that this is an overall good change. Thanks for the effort!

Some comments:
- Skipping `CanScriptTest` sounds reasonable, they are more problematic than initially expected.
- Why did you abandon the `setup.cfg`? If it's still supported, I'd suggest keeping it sine it prevents the source folder from getting cluttered with all those config files.

I don't see the need for factoring this into multiple PRs.

karlding:
> Why did you abandon the `setup.cfg`? If it's still supported, I'd suggest keeping it sine it prevents the source folder from getting cluttered with all those config files.

It is still supported, but I figured since `setuptools` was being stripped out it would make sense to also remove this. I don't really have strong feelings either way. :shrug:

Also, I suppose we could go less drastic and continue using the `setuptools` backend while specifying the backend in `pyproject.toml`, and revisit the migration to Poetry at a later time, if that makes the changes more reasonable. After all, right now there's no PEP covering the use case for editable installs, so if this is something we want to support, we would still need to provide a `setup.py` shim layer anyways.

felixdivo:
@karlding

I read up on this a bit more, and agree with you on:

> it would make sense to also remove this

or at least gradually shift to `pyproject.toml` as the tools start supporting it. Using Poetry sounds good too.

pyproject.toml
```diff
@@ -0,0 +1,81 @@
+[tool.poetry]
+name = "python-can"
+version = "3.2.0"
+description="Controller Area Network interface module for Python"
+license = "LGPL-3.0-only"
+authors = ["Brian Thorne <brian@thorne.link>"]
```
hardbyte, line 6: It wouldn't be controversial to remove the `authors` and perhaps just list the mailing address email `python-can@googlegroups.com` under maintainers

hardbyte:
What does "DNI" stand for in this context @karlding ?

Keen to hear any feedback from anyone before making this change, cc @christiansandberg @felixdivo @pierreluctg @zariiii9003

karlding:
DNI: Do Not Integrate, since this is a somewhat controversial change, so I wanted to make it explicit that we shouldn't merge this as is unless it got some more eyes on the change.

zariiii9003:
Pip 20.3 introduced a dependency resolver. What are the remaining benefits of switching to poetry?
I'm not against it, i use poetry in one of my own projects. Just curious.

hardbyte:
The key change will be removing `setup.py` so the build isn't platform dependent but rather fully declarative.
 
IMHO adding poetry is just a nice to have for us library developers. Most end users shouldn't need to know about it.

felixdivo:
Moving stuff to the `pyproject.toml` file (but not force using to poetry) sounds like a good idea! Now, [coveragepy also supports being configured via that file](https://coverage.readthedocs.io/en/coverage-5.5/config.html#configuration-reference):

> If coverage.py is installed with the toml extra (pip install coverage[toml]), it will automatically read from “pyproject.toml”. Configuration must be within the [tool.coverage] section, for example, [tool.coverage.run].

felixdivo:
Is someone willing to open a new PR without the conflicts? I is probably quite straightforward and mainly the CI will check whether it worked.

PR #827, opened 2020-05-06
Enable GitHub Actions runners to run tests

karlding:
GitHub announced Actions at GitHub Universe in 2018, however this was
somewhat limited, and support for running CI/CD pipelines was only
released in August 2019. This was in beta for a long time, and finally
has been rolled out to the general public.

This adds a minimal GitHub Actions workflow that runs all the current
tests on a build matrix consisting of {Ubuntu,macOS,Windows} and Python
3.{6,7,8}. This also seems to run much faster than the current Travis CI
pipelines, and actually lets us cover multiple versions on macOS without
a series of hacks.

Unfortunately, the kernel that GitHub Actions uses (5.0.0-1035-azure)
doesn't include the vcan kernel module, so we still need Travis CI to
run SocketCAN tests.


karlding:
This adds better coverage on macOS, so I think those changes are pretty reasonable, unless someone has a preference for Azure Pipelines (although it's all [Microsoft](https://twitter.com/natfriedman/status/1159526215658561536) :joy:) or some other provider (Semaphore CI)?

The parts that are potentially controversial are whether we want to migrate off Travis CI for their Linux builds (and potentially releases) and AppVeyor for their Windows builds, and how far we want to go with that, or whether we want to just run these in parallel. I've deferred that from this PR for the most part, but I can also revert the additions of Ubuntu and Windows if we don't want the duplication.

hardbyte:
@karlding if you update this branch it should merge in automatically. I can't update it or push the changes to develop myself without disabling the branch protection which I don't want to do.

karlding:
Hmm, seems like Mergify/GitHub has some issues when adding a new file that adds additional steps to the workflow..

hardbyte:
@Mergifyio refresh

PR #828, opened 2020-05-06
Possible fix for tests sometimes failing

sou1hacker:
Fixed a minor issue in tests where I think previously we were sometimes getting some left over messages from the first cyclic thread, causing the tests to fail. 

One example PR where the tests were failing due to this: https://github.com/hardbyte/python-can/pull/825
Travis link: https://travis-ci.org/github/hardbyte/python-can/jobs/683606723

### DIFF
--- a/test/simplecyclic_test.py
+++ b/test/simplecyclic_test.py
@@ -168,6 +168,7 @@ class SimpleCyclicSendTaskTest(unittest.TestCase, ComparingMessagesTestCase):
         on_error_mock.assert_not_called()
         task.stop()
         bus.shutdown()
+        sleep(.5)
 
         # bus has been shutted down
         on_error_mock.reset_mock()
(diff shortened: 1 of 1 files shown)


sou1hacker:
Original PR: https://github.com/hardbyte/python-can/pull/781 (@tysonite )

tysonite:
Ohh, I see the issue, thanks for fixing it!

karlding:
Maybe this obvious and I'm missing something, but I don't see how this solves the problem if you're receiving messages due to the thread loop condition check racing against the flag being set when `stop` is called...

Calling `reset_mock` resets all the call attributes on the mock, which imo is correct. And constructing a new mock doesn't provide a synchronization point or barrier, which is what's needed in the first place.

I would have expected the solution to either create a barrier before starting the next set of assertions, or prevent the race from affecting the test by explicitly synchronizing on the last message being delivered. You can't really create a barrier since the send thread isn't controlled by the test (`stop` would have to do so, but those aren't the semantics it provides), so your options are either:

1. Teardown the bus and start a new one on a separate channel
2. Wait for `>= 1` period so the messages get flushed before continuing the test

sou1hacker:
The first approach I committed in this PR actually had a sleep between testing the first scenario and the second. However, this approach is slightly better since we don't waste time on sleeping and I don't think we are compromising on the existing test itself in any way.

PR #829, opened 2020-05-06
Add callback for updating Message data

zariiii9003:
Add callback to `Message` class. Make the `data` attribute a property with getter and setter.
Every call to `Message.data` will return the updated data if a callback function is present.

I added an alternative property `current_data` to retrieve the data without triggering the update.

### DIFF
--- a/can/message.py
+++ b/can/message.py
@@ -77,11 +77,6 @@ class Message:
             Possible problems include the `dlc` field not matching the length of `data`
             or creating a message with both `is_remote_frame` and `is_error_frame` set to `True`.
 
-        :param data_callback:
-            This optional function is called every time the `data` property is accessed. The function
-            `data_callback` must accept `arbitration_id: int, data: CanData` as arguments and return
-            the new `data`.
-
         :raises ValueError: iff `check` is set to `True` and one or more arguments were invalid
         """
         self.timestamp = timestamp
--- a/can/message.py
+++ b/can/message.py
@@ -64,7 +64,7 @@ class Message:
         bitrate_switch: bool = False,
         error_state_indicator: bool = False,
         check: bool = False,
-        data_callback: Optional[Callable[[int, CanData], CanData]] = None,
+        data_callback: Optional[Callable[[int, bytearray], bytearray]] = None,
     ):
         """
         To create a message object, simply provide any of the below attributes
@@ -91,10 +91,15 @@ class Message:
         self.error_state_indicator = error_state_indicator
         self.data_callback = data_callback
 
+        self._data: bytearray
         if data is None or is_remote_frame:
             self._data = bytearray()
         else:
-            self.data = data
+            try:
+                self.data = bytearray(data)
+            except TypeError:
+                err = "Unable to set message data {} ({})".format(data, type(data))
+                raise TypeError(err)
 
         if dlc is None:
             self.dlc = len(self._data)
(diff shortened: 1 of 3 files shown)


can/message.py
```diff
@@ -55,24 +56,26 @@ def __init__(
...
+        channel: Optional[Channel] = None,
         dlc: Optional[int] = None,
-        data: Optional[typechecking.CanData] = None,
+        data: Optional[CanData] = None,
         is_fd: bool = False,
         is_rx: bool = True,
         bitrate_switch: bool = False,
         error_state_indicator: bool = False,
         check: bool = False,
+        data_callback: Optional[Callable[[int, bytearray], bytearray]] = None,
```
karlding, line 67: Before we get too deep with this review, I just want to clarify a few things about the context behind this change. This is motivated because you want support for periodic transmissions where the CAN message data can be arbitrarily modified, right?

First of all, do any interfaces support arbitrary modifications when doing periodic transmissions natively, or will every interface need to use a userspace implementation? I'm not arguing that there's no desire for this feature—there's clearly a desire for support, since this comes up every few months whenever someone uses the `send_periodic` API. But as `python-can` primarily is providing an API that abstracts over CAN interfaces, if no interface supports it, then all we're doing is providing a `Thread` or scheduling mechanism that can be implemented by any user in their own application.

Also, I'm curious why you chose to go about this by making `data` a generator as opposed to introducing a new API that takes a `Message` generator. This means that this wouldn't support a use case where someone wants messages sent ever `x` seconds where the Arbitration ID _and_ the data can vary between Messages (instead of just modifying the data), which seems like a valid use case if we're already supporting arbitrary modifications of the data when doing periodic transmissions. Was there a reason why you decided to implement it this way?

zariiii9003: Hello Karl, thank you for your review!

I only develop for vector tools so i can not say anything of substance about native periodic transmissions. This probably should be mentioned in the docs if this proposal is acceptable.

I chose this implementation because it requires the least amount of changes and should not break compatibility. 

Yes this can easily be implemented by the user (as everything else can be) but i noticed that this topic comes up quite often so why not provide a simple solution?

I'm not sure how a message generator api would work for this. 
Here is my use case: i have 30 messages with cycle times from 10ms to 1s. Most of these have CRC and Rolling Counters. Also i want the signal values to update as quickly as possible. So for my use case updating just the data seems like a reasonable solution. I create the messages once with a callback and register them with my BroadcastManager that takes care of sending the messages at the correct times.

karlding: I haven't had much time to look at this, so sorry if I haven't been responding with any sort of SLA.

> I'm not sure how a message generator api would work for this.

Basically from a 5,000 foot view, I think it's similar to this data generator, but instead of yielding new data, you would be yielding new Messages. This way we would keep messages 1:1 with what's actually being sent, instead of having a separate overload for the "current" data vs the actual data. You can pass in any context variables to the generator that are needed, and the generator returns the next message. This way support for different Arbitration IDs is baked in for "free".

I don't really like the idea of overloading the fact the message data is not immutable. Having 2 fields denoting what the data represents seems non-ideal, since the message no longer represents a CAN message that's being sent/received (but rather, some higher-level protocol notion of a message).

You should be able to hack something together by essentially doing what the default `ThreadBasedCyclicSendTask` is doing as a Proof of Concept, but instead of incrementing the message index, you call the infinite generator. This might result in slightly more boilerplate, but I think it makes things more explicit.

hardbyte:
I think generating new messages would be a better approach than adding this proposed callback. Appreciate the proposal and the implementation though! 

I'll close this PR now, but feel free to open an issue to discuss the proposal or to submit another PR!

Issue #582, opened 2019-05-13
Use TOX


felixdivo:

zariiii9003:
Closing this because it's implemented in develop with #833 .

PR #833, opened 2020-05-10
use tox instead of 'setup.py test'

zariiii9003:
See here: 
- https://github.com/pypa/setuptools/pull/1878
- https://github.com/hardbyte/python-can/issues/582


zariiii9003:
`test_thread_based_cyclic_send_task` failed and i can not rerun it. [All tests passed in my fork](https://travis-ci.org/github/zariiii9003/python-can/builds/685404482).

I guess https://github.com/hardbyte/python-can/pull/828 did not fix it.

setup.py
```diff
@@ -29,30 +28,9 @@
 extras_require = {
     "seeedstudio": ["pyserial>=3.0"],
     "serial": ["pyserial~=3.0"],
-    "neovi": ["python-ics>=2.12", "filelock"],
+    "neovi": ["python-ics>=2.12"],
```
hardbyte, line 31: Should this change be here?

zariiii9003: Not necessarily, but i noticed that filelock is mentioned in install_requires already.

PR #835, opened 2020-05-12
Configure automatic merging 

hardbyte:
By adding a mergify config we should get some automation benefits.


felixdivo:
@hardbyte This branch can be deleted too, right?

PR #839, opened 2020-05-19
Support PCAN interface on cygwin

henzef:
this is the pull request for the 3.x branch (see #840 for the 4.x branch)


felixdivo:
@hardbyte Do we still add new features to version 3.x?

henzef:
Does that even count as a "feature"?

felixdivo:
> Does that even count as a "feature"?

Yes, a new platform being supported does not seem to be a bugfix for me. (See [semver](https://semver.org/#semantic-versioning-200))

henzef:
I would add it in the 3.x branch as its a tiny change that cannot break existing functionality, but it's up to you to decide. (it's also questionable if cygwin counts as a "platform", it's still windows after all)

Issue #736, opened 2019-12-04
Meaning of can.BusABC.state


onvav:
Hello,
please help me understand the meaning of `class BusState(Enum):`. I would like to show status of the interface (Kvaser Leaf and Vector VN1610 are on my desk) as specified by CAN specification (error active, error passive and bus-off state).
But the @property `def state(self):` of the classes `class KvaserBus(BusABC):` and `class VectorBus(BusABC):` are not implemented.

I would like to implement them but I am not sure what it shall returns
- state of the CAN bus as defined by standard (active, passive, error=busoff)
- state of the Python class BusABC (active, passive=silent_mode, error=some_driver)

The documentation says that it returns ["current state of the hardware"](https://python-can.readthedocs.io/en/master/bus.html#can.BusABC.state). Does it mean state of driver connection with HW interface or state of CAN bus controller (active->passive->busoff)?

christiansandberg:
Unfortunately it is a mix of both. I’m guessing the idea was that when reading it you would get the first and when writing to it you change the second. You could look at another implementation, like pcan for inspiration.

bggardner:
I was writing a post to the message board, but found this issue and decided to post it here instead:

Could we consider renaming the BusState enum options according to the CAN specification (ERROR_ACTIVE, ERROR_PASSIVE, BUS_DOWN)?

Also, as I primarily use socketcan, there is a fourth state ("STOPPED"), which means the interface (not just the "bus") is down and will not automatically recover (as in Bus-Off) until brought back up.  I'm not sure how applicable this is to the other interfaces, but maybe add this fourth state?  **Knowing the difference between Bus-Off and down/stopped in an application can be important.**  For those who are interested, the SocketCAN state can be found using "ip -d link show <interface>", then finding the string on this line: "can state <STATE> restart-ms <TIME-MS>".

Additionally, defaulting BusABC's state property to be ACTIVE seems to be bad practice.  It seems the better practice would be to either make the state property throw a NotImplementedError() by default or default to a new UNKNOWN enum value.  If there is no way to detect the state for an interface and it actually needs the state property to return ACTIVE (and not UNKNOWN) or an exception, then only that interface's Bus class would override the state property.  This may solve the "mix of both" issue by simply defining more states, or making them different class properties (hardware state vs. software state).  I would prefer a single property (state) with more enum options.

I can work on a PR if pursuing this is worthwhile.

bggardner:
[...]

PR #847, opened 2020-06-01
Renamed BusState members and added SocketcanBus.state override

bggardner:
RE: #736


zariiii9003:
I agree that this should be fixed as the naming is very confusing. 
But i think you did not consider that BusState.Active and BusState.Passive was used to distinguish active and silent mode on some (or all?) interfaces.

It might be better to create a new Enum `ErrorState` and rename `BusState` to `BusMode` or something like that.

felixdivo:
#736 should first reach a consensus. Afterward, this PR might be reopened again.

Issue #836, opened 2020-05-12
View virtual CAN messages on CANoe


gcShyam:
I am trying to implement a rest bus simulation using the library.
To debug if the code is correct, I use the trace window in CANoe application to check the messages.
Currently, I am trying to send CAN messages on the virtual CAN bus.

The following is the starting code which is supposed to send a message on the virtual CAN.
```
import can
def send_one():
    with can.interface.Bus() as bus:
        bus = can.interface.Bus(bustype='virtual', app_name='SelfApp', channel=[0])
        msg = can.Message(arbitration_id=0xC0FFEE, data=[0, 25, 0, 1, 3, 1, 4, 1], is_extended_id=True)
        try:
            bus.send(msg)
            print(f"Message sent on {bus.channel_info}")
        except can.CanError:
            print("Message NOT sent")
if __name__ == "__main__":
    send_one()
```

I get the following error:
```Traceback (most recent call last):
  File ".../SendMsg.py", line 51, in <module>
    send_one()
  File ".../SendMsg.py", line 30, in send_one
    with can.interface.Bus() as bus:
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\site-packages\python_can-3.3.2-py3.6.egg\can\interface.py", line 108, in __new__
  File "C:\Program Files (x86)\Microsoft Visual Studio\Shared\Python36_64\lib\site-packages\python_can-3.3.2-py3.6.egg\can\util.py", line 197, in load_config
NotImplementedError: Invalid CAN Bus Type - None```

Can someone please help me to remove this error?

**Comment by bessman** [comment]
You should create the bus when you open the `with` block, not inside it:
```
import can
def send_one():
    with can.interface.Bus(bustype='virtual', app_name='SelfApp', channel=0) as bus:
        msg = can.Message(arbitration_id=0xC0FFEE, data=[0, 25, 0, 1, 3, 1, 4, 1], is_extended_id=True)
        try:
            bus.send(msg)
            print(f"Message sent on {bus.channel_info}")
        except can.CanError:
            print("Message NOT sent")
if __name__ == "__main__":
    send_one()
```
Also, note that `channel` cannot be a list.

gcShyam:
Now I get the following message.
`Message sent on Virtual bus channel 0`
But still, I am unable to see the message on the trace window of CANoe application.

bessman:
[...]

PR #849, opened 2020-06-02
Add methods get_application_config(), set_application_config() and set_timer_rate() to VectorBus

zariiii9003:
As promised [here](https://github.com/hardbyte/python-can/issues/836#issuecomment-628878221)


can/interfaces/vector/canlib.py
```diff
@@ -586,15 +576,99 @@ def popup_vector_hw_configuration(wait_for_finish: int = 0) -> None:
...
+        xldriver.xlSetApplConfig(
+            app_name.encode(),
+            app_channel,
+            hw_type.value,
+            hw_index,
+            hw_channel,
+            bus_type.value,
+        )
+
+    def set_timer_rate(self, timer_rate_ms: int) -> None:
```
christiansandberg, line 649: How would you use this method?

zariiii9003: You run `set_timer_rate(10)` once and the interface will generate an event with the tag `XL_TIMER` every 10ms. If you have an active notifier this will cause the method `handle_can_event()` or `handle_canfd_event` to be called.

It should be a relatively precise timer since it is hardware generated. But since i am in home office i did not really have an opportunity to test how reliable it is.

zariiii9003: XL_TIMER:
XL_TIMER

Windows Multimedia Timer:
MultimediaTimer

Tested with a full restbus simulation on virtual can with a VN1610 attached. The timing looks good but the XL_TIMER events just get dropped if another event occurs at the same time which is quite disappointing.

Even on my slow Notebook the Windows MM Timer works better.

So it really is only useful to tell if the bus is still alive.

zariiii9003: Update: it is actually quite decent, there was something else hogging my event loop.

XL_TIMER

christiansandberg: I think I read about it in the API documentation thinking it would be a good candidate for sending cycling messages, but if I remember correctly the timer was reset each time some other event occurred so it didn't feel very useful. Since we usually have a timeout when waiting for events to keep the application alive even when there are no messages I'm not sure what the use case is, but on the other hand it doesn't hurt to have more options.

PR #851, opened 2020-06-05
Refactor VectorBus.send()

zariiii9003:
I made the `send()` method a little cleaner.

The `send()` method now calls `_send_can_msg_sequence()` which can send multiple messages  with a single vxlapi call. But i did not add it to the public API since there is another PR #688 for that.

(Also i fixed a mistake i made in the `handle_can_event`/`handle_canfd_event` docstrings)

### DIFF
--- a/can/interfaces/vector/canlib.py
+++ b/can/interfaces/vector/canlib.py
@@ -11,6 +11,8 @@ import logging
 import time
 import os
 
+import typing
+
 try:
     # Try builtin Python 3 Windows API
     from _winapi import WaitForSingleObject, INFINITE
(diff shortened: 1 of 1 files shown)


can/interfaces/vector/canlib.py
```diff
@@ -474,65 +475,101 @@ def handle_canfd_event(self, event: xlclass.XLcanRxEvent) -> None:
...
 
-        # If channel has been specified, try to send only to that one.
-        # Otherwise send to all channels
-        mask = self.channel_masks.get(msg.channel, self.mask)
+        xl_event = xlclass.XLevent()
+        xl_event.tag = xldefine.XL_EventTags.XL_TRANSMIT_MSG.value
+        xl_event.tagData.msg.id = msg_id
+        xl_event.tagData.msg.dlc = msg.dlc
+        xl_event.tagData.msg.flags = flags
+        xl_event.tagData.msg.data = tuple(msg.data)
```
christiansandberg, line 530: Have you verified that this works for DLC < 8?

zariiii9003: Just tried it, no problems here.

PR #853, opened 2020-06-06
CANtact Support

ericevenchick:
Hi, this PR adds support for the CANtact devices, including the original device running [candleLight firmware](https://github.com/candle-usb/candleLight_fw), the CANable/CANable Pro, and the upcoming CANtact Pro.

Users can install the driver and supporting packages using pip: `pip install cantact`. This is tested across Mac, Windows and Linux.

The CANtact Pro will support CAN-FD as well, but currently none of these devices support FD. I will provide another PR once FD support is ready!

Thanks,
Eric


zariiii9003:
You could add `cantact` to `extras_require` in `setup.py`

ericevenchick:
I've added the package to `extras_require` and fixed formatting.

ericevenchick:
Added testing and resolved issues with running them on Travis.

Also, SimpleCyclicSendTaskTest was randomly causing failures in CI, which also seems to be true of the develop branch. I've disabled this test in CI to resolve the issue. I'm not sure if this is the best place to make that fix, but it was needed for CI to run reliably.

hardbyte:
Eric it would be great to add CANtact support directly to python-can. 🚀 

I was under the impression that CANtact is supported via slcan and socketcan? If so might be worth adding a note on what advantages one method might have or what operating systems require what backend etc.

Can you please add an interface documentation page and link to it from [interfaces](https://python-can.readthedocs.io/en/master/interfaces.html). I'd like to see upstream links so people can easily find the right place to look if/when they run into trouble.

Is hardware filtering supported?

ericevenchick:
> I was under the impression that CANtact is supported via slcan and socketcan? If so might be worth adding a note on what advantages one method might have or what operating systems require what backend etc.

CANtact was originally slcan + SocketCAN, but now has a driver based on gs_usb. Of course, this only works well on Linux. The main benefit of adding direct support will be easy cross-platform support.

With this PR, getting setup with CANtact + python-can consists of:
```
python3 -m pip install cantact git+https://github.com/ericevenchick/python-can@cantact
can_logger.py -i cantact -c 0 -b 500000
```
This works across Linux, macOS, and Windows.

> Can you please add an interface documentation page and link to it from [interfaces](https://python-can.readthedocs.io/en/master/interfaces.html). I'd like to see upstream links so people can easily find the right place to look if/when they run into trouble.

Will do.

> Is hardware filtering supported?

Not right now. This is because gs_usb doesn't have that feature. However, CANtact is going to branch from the gs_usb driver very shortly to add features such as CAN-FD and single wire CAN for the new CANtact Pro. The goal is to continue reverse compatibility with gs_usb, at least until I can get a driver mainlined into the kernel (no idea how long that'll take), but probably forever since having support on older kernels is nice.

When I add FD support, I'll be adding the other non-gs_usb features as well. I anticipate another PR for the update, but the driver code and firmware are still in the works.

ericevenchick:
Just checking in to see if there's anything missing from this PR. Happy to made whatever edits are needed to get it merged :)

can/interfaces/cantact.py
```diff
@@ -0,0 +1,135 @@
...
+        for i in range(0, interface.channel_count()):
+            channels.append({"interface": "cantact", "channel": "ch:%d" % i})
+        return channels
+
+    def __init__(
+        self,
+        channel,
+        bitrate=500000,
+        poll_interval=0.01,
+        monitor=False,
```
zariiii9003, line 42: The BusABC class has the `state` property for setting active/silent mode. Although i personally prefer your solution

can/interfaces/cantact.py
```diff
@@ -0,0 +1,135 @@
...
+        _testing=False,
+        **kwargs
+    ):
+        """
+        :param int channel:
+            Device number
```
zariiii9003, line 48: A short description on where to find that number would be nice. Also: Are channel lists supported?

can/interfaces/cantact.py
```diff
@@ -0,0 +1,135 @@
...
+        return channels
+
+    def __init__(
+        self,
+        channel,
+        bitrate=500000,
+        poll_interval=0.01,
+        monitor=False,
+        _testing=False,
+        **kwargs
```
zariiii9003, line 44: Does the hardware support custom bit timings?

can/interfaces/cantact.py
```diff
@@ -0,0 +1,135 @@
...
+    def _recv_internal(self, timeout):
+        frame = self.interface.recv(int(timeout * 1000))
+        if frame is None:
+            # timeout occured
+            return None, False
+
+        msg = Message(
+            arbitration_id=frame["id"],
+            is_extended_id=frame["extended"],
+            timestamp=time.time(),  # Better than nothing...
```
zariiii9003, line 82: Is there no hardware timestamp?

can/interfaces/cantact.py
```diff
@@ -0,0 +1,135 @@
...
+            is_remote_frame=frame["rtr"],
+            dlc=frame["dlc"],
+            data=frame["data"][: frame["dlc"]],
+            channel=frame["channel"],
+        )
```
zariiii9003, line 87: Can the interface distinguish between RX and TX messages? See #776

doc/installation.rst
```diff
@@ -90,6 +90,14 @@ To install ``python-can`` using the XL Driver Library as the backend:
 
 3. Use Vector Hardware Configuration to assign a channel to your application.
 
+CANtact
+~~~~~~~
+
+To install ``python-can`` using the CANtact driver backend:
+
+1. Install the Python ``cantact`` package: ``python3 -m pip install cantact``
```
zariiii9003, line 98: Remove the `1.` if there is no second step. It might be worth mentioning that you can install via `pip install "python-can[cantact]"`. You could also provide a link to the cantact docs

can/interfaces/cantact.py
ericevenchick: I think this is a bit more explicit, so I'll leave it as is.

can/interfaces/cantact.py
ericevenchick: I'll improve the documentation for this one.

can/interfaces/cantact.py
ericevenchick: This varies by device, but I think it makes sense to move that to the driver and have it determine the timestamp if available (or use system time if not).

can/interfaces/cantact.py
ericevenchick: Yes, I need to expose that to Python though. I'll work on that change!

ericevenchick:
Made some improvements based on your feedback @zariiii9003. Timestamp and `is_rx` are now provided by the driver. There is now an option to use custom bit timings. Updated documentation based on feedback and added additional test for bit timing.

PR #854, opened 2020-06-09
Filter out Tx error messages

pierreluctg:
Filter out Tx message with global error flag set.

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -261,9 +261,15 @@ class NeoViBus(BusABC):
         for ics_msg in messages:
             if ics_msg.NetworkID not in self.channels:
                 continue
+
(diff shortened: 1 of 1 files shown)


can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -261,9 +261,16 @@ def _process_msg_queue(self, timeout=0.1):
         for ics_msg in messages:
             if ics_msg.NetworkID not in self.channels:
                 continue
+
             is_tx = bool(ics_msg.StatusBitField & ics.SPY_STATUS_TX_MSG)
-            if not self._receive_own_messages and is_tx:
-                continue
+            global_err = bool(ics_msg.StatusBitField & ics.SPY_STATUS_GLOBAL_ERR)
```
felixdivo, line 266: Could move this into `if is_tx:` since it is only needed there. Not really a performance issue though, since the difference is probably immeasurable. ;)

PR #856, opened 2020-06-11
ixxat: Raise exception on busoff in recv()

henzef:
I recently ran into a problem where the ixxat would switch into a busoff state and my application was not notified about that. This patch should address that problem.


henzef:
So can this be merged? :-)

PR #858, opened 2020-06-16
Converting vector interface argument to lowercase

pierreluctg:
Variable name should be lowercase, with words separated by underscores as necessary to improve readability. (PEP8)

Current CamelCase arguments cannot be set from the configuration file since the variables are read as lowercase.


pierreluctg:
@felixdivo deprecation warning (and aliases) added

pierreluctg:
AppVeyor unit test failure is unrelated to this change.

can/interfaces/vector/canlib.py
```diff
@@ -49,9 +52,50 @@
...
+def deprecated_args_alias(**aliases):
+    def deco(f):
+        @functools.wraps(f)
+        def wrapper(*args, **kwargs):
+            rename_kwargs(f.__name__, kwargs, aliases)
+            return f(*args, **kwargs)
+
+        return wrapper
+
+    return deco
```
hardbyte, line 64: Not that there is anything wrong with this (in fact it is really neat) but these helpers shouldn't be here.

Either it should be moved into `can.util` or perhaps consider adding a dependency that could do it?
https://github.com/tantale/deprecated or equivalent.

can/interfaces/vector/canlib.py
```diff
@@ -64,12 +108,12 @@ def __init__(
...
-        tseg2Abr=3,
-        sjwDbr=2,
-        tseg1Dbr=6,
-        tseg2Dbr=3,
+        sjw_abr=2,
+        tseg1_abr=6,
+        tseg2_abr=3,
+        sjw_dbr=2,
+        tseg1_dbr=6,
+        tseg2_dbr=3,
```
hardbyte, line 116: 👍

can/interfaces/vector/canlib.py
```diff
@@ -49,9 +52,50 @@
...
+                raise TypeError(
+                    "{} received both {} (deprecated) and {}".format(
+                        func_name, alias, new
+                    )
+                )
+            kwargs[new] = kwargs.pop(alias)
+
+
+def arg_to_c_uint(value):
+    return ctypes.c_uint(int(value))
```
hardbyte, line 83: I'd prefix this function with an underscore to mark it internal.

A better place for it would be `can.ctypesutil`

can/interfaces/vector/canlib.py
```diff
@@ -217,19 +242,19 @@ def __init__(
...
                 self.canFdConf = xlclass.XLcanFdConf()
                 if bitrate:
-                    self.canFdConf.arbitrationBitRate = ctypes.c_uint(bitrate)
+                    self.canFdConf.arbitrationBitRate = arg_to_c_uint(bitrate)
                 else:
-                    self.canFdConf.arbitrationBitRate = ctypes.c_uint(500000)
-                self.canFdConf.sjwAbr = ctypes.c_uint(sjwAbr)
-                self.canFdConf.tseg1Abr = ctypes.c_uint(tseg1Abr)
-                self.canFdConf.tseg2Abr = ctypes.c_uint(tseg2Abr)
+                    self.canFdConf.arbitrationBitRate = arg_to_c_uint(500000)
```
christiansandberg, line 247: I don’t think you need to convert it to a ctypes type explicitly since the type of the attribute has already been specified so ctypes will do it for you. Converting to int would suffice.

pierreluctg: agree

can/util.py
```diff
@@ -279,6 +280,38 @@ def channel2int(channel: Optional[Union[typechecking.Channel]]) -> Optional[int]
     return None
 
 
+def deprecated_args_alias(**aliases):
+    """Allows to rename/deprecate a function kwarg(s) and
+    have the deprecated kwarg(s) set as alias(es)
+    """
```
hardbyte, line 286: Lets add an example to the docstring:

```suggestion
def deprecated_args_alias(**aliases):
    """Allows to rename/deprecate a function kwarg(s) and
    have the deprecated kwarg(s) set as alias(es)
    
    Example:
    
        @deprecated_args_alias(oldArg="new_arg", anotherOldArg="another_new_arg")
        def library_function(new_arg, another_new_arg):
            pass
    """
```

A test would be nice, as would supporting removed arguments.

pierreluctg: Both are in now ;)

pierreluctg:
Travis keeps getting random test_thread_based_cyclic_send_task error unrelated to this change.

PR #860, opened 2020-06-19
Updates for 4.0.0

hardbyte:
A bunch of changelog entries were missing from the backported releases.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,11 +5,11 @@
 # run arbitrary code.
 extension-pkg-whitelist=
 
-# Add files or directories to the blacklist. They should be base names, not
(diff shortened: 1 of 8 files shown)


hardbyte:
Oh good catch, hopefully there aren't too many changes to docs and things that went into the 3.x series that didn't make it into develop!

Yeah I've been a bit uncomfortable being the "author" of python-can while it has so clearly been developed by so many people.

hardbyte:
First failure looks like pylint has made a new release. 
Now it looks like `SimpleCyclicSendTaskTest::test_thread_based_cyclic_send_task` is flaky at least on Python 3.8.

PR #865, opened 2020-07-09
Raising more precise API error when set bitrate fails

pierreluctg:
Also calling shutdown before raising exception from the init if the device is opened

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -158,15 +158,25 @@ class NeoViBus(BusABC):
         with open_lock:
             ics.open_device(self.dev)
 
-        if "bitrate" in kwargs:
-            for channel in self.channels:
-                ics.set_bit_rate(self.dev, kwargs.get("bitrate"), channel)
-
-        fd = kwargs.get("fd", False)
-        if fd:
-            if "data_bitrate" in kwargs:
+        try:
+            if "bitrate" in kwargs:
                 for channel in self.channels:
-                    ics.set_fd_bit_rate(self.dev, kwargs.get("data_bitrate"), channel)
+                    ics.set_bit_rate(self.dev, kwargs.get("bitrate"), channel)
+
+            fd = kwargs.get("fd", False)
+            if fd:
+                if "data_bitrate" in kwargs:
+                    for channel in self.channels:
+                        ics.set_fd_bit_rate(
+                            self.dev, kwargs.get("data_bitrate"), channel
+                        )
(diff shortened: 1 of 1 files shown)


can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -158,15 +158,23 @@ def __init__(self, channel, can_filters=None, **kwargs):
...
+            if fd:
+                if "data_bitrate" in kwargs:
+                    for channel in self.channels:
+                        ics.set_fd_bit_rate(
+                            self.dev, kwargs.get("data_bitrate"), channel
+                        )
+        except ics.RuntimeError as re:
+            logger.error(re)
+            err = ICSApiError(*ics.get_last_api_error(self.dev))
+            self.shutdown()
```
felixdivo, line 176: Just a thought: can this throw an exception again? Would we also want to catch that?

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -158,15 +158,23 @@ def __init__(self, channel, can_filters=None, **kwargs):
...
-        fd = kwargs.get("fd", False)
-        if fd:
-            if "data_bitrate" in kwargs:
+        try:
+            if "bitrate" in kwargs:
                 for channel in self.channels:
-                    ics.set_fd_bit_rate(self.dev, kwargs.get("data_bitrate"), channel)
+                    ics.set_bit_rate(self.dev, kwargs.get("bitrate"), channel)
+
+            fd = kwargs.get("fd", False)
```
felixdivo, line 166: Could be inlined into the if statement's condition. (Does not need to be a variable.) That's nothing too important though. ;)

pierreluctg: Agree, but will not touch it in this PR

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Yes, `shutdown` (which call `ics.close_device` can raise a `ics.RuntimeError`. However, in this case we do not want to call `ics.get_last_api_error`.

Will fix

pierreluctg:
@hardbyte can you please merge. The `test_thread_based_cyclic_send_task` keep failing either on Trusty or OSX.

can/interfaces/ics_neovi/neovi_bus.py
felixdivo: Cool!

felixdivo:
They randomly worked when I restarted them a minute ago.

PR #869, opened 2020-07-15
Simplified the serial number decode conditions

pierreluctg:
Also fixing to the correct range.


pierreluctg:
@hardbyte can you please merge. Thank you

PR #877, opened 2020-07-28
Add a default fileno function to the BusABC class. 

hardbyte:
Modify Notifier to handle the exception raised if the bus doesn't override it.

### DIFF
--- a/can/notifier.py
+++ b/can/notifier.py
@@ -61,24 +61,26 @@ class Notifier:
         :param bus:
             CAN bus instance.
         """
+        reader: int = -1
         try:
             reader = bus.fileno()
         except NotImplementedError:
-            # Bus doesn't support fileno
-            reader = None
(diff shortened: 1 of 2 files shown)


can/notifier.py
```diff
@@ -61,13 +61,15 @@ def add_bus(self, bus: BusABC):
...
+        try:
+            reader = bus.fileno()
+        except NotImplementedError as e:
+            # Bus doesn't support fileno
+            reader = None
+
+        if self._loop is not None and reader is not None and reader >= 0:
             # Use file descriptor to watch for messages
-            reader = bus.fileno()  # type: ignore
+            reader = bus.fileno()
```
zariiii9003, line 72: This line seems redundant now

bmeisels:
Looks good. I would like to make a followup change to slcan which will raise NotImplementedError on windows (which was the original motivation for this change).

Issue #874, opened 2020-07-23
How to record CAN log with several shorter pieces


DanielTian90:
If run python-can for a long time to record CAN trace with ASCWriter, size of asc file is extremely large(more than 5GB), and hard to open and analyze by other tools.
Is there any easier way to record long term asc file with several files with specific size like 1GB, and not missing any CAN messages.

hardbyte:
It should be relatively straightforward to write something like a `RotatingFileWriter` - it could be a good addition to the `can.io` module if you wanted to open a PR. 

Suggest it implements the [`Listener`](https://github.com/hardbyte/python-can/blob/develop/can/listener.py#L22) interface and just passes the messages on to an [`ASCWriter`](https://github.com/hardbyte/python-can/blob/develop/can/io/asc.py#L218) instance closing and starting a new ASCWriter when your wrap condition is met.

Let us know how you get on

zariiii9003:
Closing this because of merge to the develop branch. Feel free to reopen if there are any issues.

PR #881, opened 2020-08-01
Implement rotating file logger

zariiii9003:
See #874 

Any API or testing suggestions are very welcome.

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -24,7 +24,7 @@ class CanError(IOError):
 
 from .listener import Listener, BufferedReader, RedirectReader, AsyncBufferedReader
 
-from .io import Logger, RotatingFileLogger, Printer, LogReader, MessageSync
+from .io import Logger, SizedRotatingCanLogger, Printer, LogReader, MessageSync
 from .io import ASCWriter, ASCReader
 from .io import BLFReader, BLFWriter
 from .io import CanutilsLogReader, CanutilsLogWriter
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -24,7 +24,7 @@ class CanError(IOError):
 
 from .listener import Listener, BufferedReader, RedirectReader, AsyncBufferedReader
 
-from .io import Logger, SizedRotatingCanLogger, Printer, LogReader, MessageSync
+from .io import Logger, SizedRotatingLogger, Printer, LogReader, MessageSync
 from .io import ASCWriter, ASCReader
 from .io import BLFReader, BLFWriter
 from .io import CanutilsLogReader, CanutilsLogWriter
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -24,7 +24,7 @@ class CanError(IOError):
 
 from .listener import Listener, BufferedReader, RedirectReader, AsyncBufferedReader
 
-from .io import Logger, Printer, LogReader, MessageSync
+from .io import Logger, RotatingFileLogger, Printer, LogReader, MessageSync
 from .io import ASCWriter, ASCReader
 from .io import BLFReader, BLFWriter
 from .io import CanutilsLogReader, CanutilsLogWriter
--- a/can/io/__init__.py
+++ b/can/io/__init__.py
@@ -4,7 +4,7 @@ and Writers based off the file extension.
 """
 
 # Generic
-from .logger import Logger
+from .logger import Logger, RotatingFileLogger
 from .player import LogReader, MessageSync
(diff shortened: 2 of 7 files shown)


christiansandberg:
I remember we discussed this topic once and the challenge is that there are an infinite number of log rotation strategies so if we want to add support for one, we have to be prepared to include support for a large number of strategies, like size based, date and time based, or some external trigger like current test case or something. There will also be a need to add support to the logger CLI.

I'm not totally against it, it just needs to be carefully designed so it won't come back and bite us in the ass later on. :)

zariiii9003:
Here is an alternative. This one is similar to the python builtins logger. I copied some code from [here](https://github.com/python/cpython/blob/46e448abbf35c051e5306a4695670d7ec7abc4e9/Lib/logging/handlers.py#L45). I thought a familiar API might be beneficial.

What do you think?

```python
class BaseRotatingCanLogger(Listener, ABC):
    supported_writers = {
        ".asc": ASCWriter,
        ".blf": BLFWriter,
        ".csv": CSVWriter,
        ".log": CanutilsLogWriter,
        ".txt": Printer,
    }
    namer = None
    rotator = None
    writer: FileIOMessageWriter

    def __init__(self, *args, **kwargs):
        self.writer_args = args
        self.writer_kwargs = kwargs

        self.rollover_count = 0

    def rotation_filename(self, default_name):
        if not callable(self.namer):
            result = default_name
        else:
            result = self.namer(default_name)
        return result

    def rotate(self, source: StringPathLike, dest: StringPathLike):
        if not callable(self.rotator):
            if os.path.exists(source):
                os.rename(source, dest)
        else:
            self.rotator(source, dest)

    def on_message_received(self, msg: Message):
        if self.should_rollover(msg):
            self.do_rollover()
            self.rollover_count += 1

        self.writer.on_message_received(msg)

    def get_new_writer(self, filename: StringPathLike) -> FileIOMessageWriter:
        suffix = pathlib.Path(filename).suffix.lower()
        try:
            writer_class = self.supported_writers[suffix]
        except KeyError:
            raise ValueError(
                f'Log format with suffix"{suffix}" is '
                f"not supported by {self.__class__.__name__}."
            )
        return writer_class(filename, *self.writer_args, **self.writer_kwargs)

    def on_error(self, exc: Exception):
        ...

    @abstractmethod
    def should_rollover(self, msg: Message) -> bool:
        ...

    @abstractmethod
    def do_rollover(self):
        ...


class SizedRotatingCanLogger(BaseRotatingCanLogger):
    def __init__(self, filename: StringPathLike, max_bytes: int = 0, *args, **kwargs):
        super(SizedRotatingCanLogger, self).__init__(*args, **kwargs)

        self.base_filename = os.path.abspath(filename)
        self.max_bytes = max_bytes

    def should_rollover(self, msg: Message) -> bool:
        if self.writer is None:
            self.writer = self.get_new_writer(self.base_filename)

        if self.writer.file.tell() >= self.max_bytes:
            return True

        return False

    def do_rollover(self):
        if self.writer:
            self.writer.stop()

        sfn = self.base_filename
        dfn = self.rotation_filename(self._default_name())
        self.rotate(sfn, dfn)

        self.writer = self.get_new_writer(self.base_filename)

    def _default_name(self) -> StringPathLike:
        path = pathlib.Path(self.base_filename)
        new_name = (
            path.stem
            + datetime.now().strftime("_%Y-%m-%d_%H-%M-%S")
            + f"_#{self.rollover_count:03}"
            + path.suffix
        )
        return str(path.parent / new_name)

```

christiansandberg:
I like that!

christiansandberg:
I like that!

felixdivo (approved):
Hey @zariiii9003, that's a great implementation! Nice code, documentation and tests.

The only thing that I might discuss is the naming of the `*RotatingCanLogger`. Do we need the "Can" in there, as the plain `Logger` also does not contain it? I'd leave it out.

Oh, and thanks for showing me the ellipsis object of Python ^^. It's quite rare (and I think usually `pass` is used instead) but we can keep it.

zariiii9003:
Thank you for your review and your kind words @felixdivo 
I updated the class names. Having CAN in the class names indeed seems a bit redundant in a can library.

felixdivo (approved):
Looks good :+1:

felixdivo:
@christiansandberg Do you also approve these changes? If yes, we can merge.

christiansandberg (approved):

PR #884, opened 2020-08-03
Update installation.rst

johanbrus:
For the IXXAT drivers, also did a successful test with the V4 drivers on Win10 so maybe this can be added to the documentation?


hardbyte:
I'll push this change to develop. Thanks

PR #887, opened 2020-08-06
Release 3.3.4

hardbyte:
Would appreciate a few people testing out the beta before we release the (hopefully) final 3.x release.

Happy for any last features or fixes to make it in too.

```
pip install python-can==3.3.4b0
```

PR #892, opened 2020-08-09
Fix ixxat ascii Python 2.7

wkemps:
I used the IXXAT interface with Python 2.7. But his gave me the following error:

> File "C:\Python27\lib\site-packages\can\interfaces\ixxat\canlib.py", line 336, in __init__
>    if (UniqueHardwareId is None) or (self._device_info.UniqueHardwareId.AsChar == bytes(UniqueHardwareId, 'ascii')):
>TypeError: str() takes at most 1 argument (2 given)

I found a few sources which suggested using string.encode() instead since this is compatible with Python 2 and 3.
https://stackoverflow.com/questions/14472650/python-3-encode-decode-vs-bytes-str

The included solution fixed the problem and allowed me to use CAN communication with IXXAT USB-to-CAN device. 

P.S.: This is my first time contributing to an open source project. I tried following the guidelines, please let me know if something should be done differently.

### DIFF
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -462,7 +462,7 @@ class IXXATBus(BusABC):
             else:
                 if (UniqueHardwareId is None) or (
                     self._device_info.UniqueHardwareId.AsChar
-                    == UniqueHardwareId.encode('ascii')
+                    == UniqueHardwareId.encode("ascii")
                 ):
                     break
                 else:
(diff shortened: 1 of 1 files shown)


christiansandberg:
We (like almost everyone else) have dropped support for Python 2 so even if this change looks OK to me you will likely encounter more issues now or in the future that won't be fixed.

hardbyte:
Thanks for the contribution @wkemps, I've applied your patch to the 3.3.4 release in this commit: https://github.com/hardbyte/python-can/pull/887/commits/5c7810f65e2935ffea30054a3b4ac350da1e844e

PR #902, opened 2020-08-26
seeed interface: fix fileno issue with windows

Gussy:
**Environment:**

* Python: 3.8.2
* Windows-10-10.0.18362-SP0

**Issue**
When using the seeedstudio interface on Linux or MacOS, I'm able to connect to the interface and send/receive messages.
When running the same script on Windows 10 the following error is thrown:
```
  File "<redacted>.py", line 264, in <module>
  File "<redacted>.py", line 44, in __init__
  File "can\notifier.py", line 56, in __init__
  File "can\notifier.py", line 66, in add_bus
  File "can\interfaces\seeedstudio\seeedstudio.py", line 266, in fileno
io.UnsupportedOperation: fileno
```

From the [python-serial docs](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.fileno), `fileno()` is only supported on Posix platforms.

The commit below fixes the issue for me, without breaking functionality on Linux, although I'm not sure if this is correct solution.

### DIFF
--- a/can/interfaces/seeedstudio/seeedstudio.py
+++ b/can/interfaces/seeedstudio/seeedstudio.py
@@ -8,6 +8,7 @@ SKU 114991193
 
 import logging
 import struct
+import io
 from time import time
 from can import BusABC, Message
 
@@ -262,7 +263,7 @@ class SeeedBus(BusABC):
         return None, None
 
     def fileno(self):
-        if hasattr(self.ser, "fileno"):
+        try:
             return self.ser.fileno()
-        # Return an invalid file descriptor on Windows
-        return -1
+        except io.UnsupportedOperation:
+            raise NotImplementedError("fileno is not implemented using current CAN bus")


bmeisels:
This has come up also with with slcan (https://github.com/hardbyte/python-can/pull/760).
and the solution chosen was to raise an error which will be caught by the notifier(https://github.com/hardbyte/python-can/pull/877).
Suggest you revise your solution in line with the change to notifier as i believe your current patch changes the behavior on Linux.
I would consider detecting explicitly if you are on windows.

Gussy:
Thanks for the feedback @bmeisels.

I think that I understood what you're asking for based on those two PRs you linked to, so I went ahead and amended the commit.

can/interfaces/seeedstudio/seeedstudio.py
```diff
@@ -262,7 +263,9 @@ def _recv_internal(self, timeout):
...
 
     def fileno(self):
-        if hasattr(self.ser, "fileno"):
+        try:
             return self.ser.fileno()
-        # Return an invalid file descriptor on Windows
+        except io.UnsupportedOperation:
+            raise NotImplementedError("fileno is not implemented using current CAN bus")
+
         return -1
```
hardbyte, line 271: This line now looks unreachable, and should be removed.

Gussy: @hardbyte I've amended the commit to remote that line 👍

hardbyte (approved):
Thanks! 

Sorry it took me a long time to get back to this

Gussy:
Is there anything I can do to help get this merged into develop?

karlding (approved):

PR #903, opened 2020-08-27
Backport Neovi features for release 3.3.4

pierreluctg:
Includes https://github.com/hardbyte/python-can/pull/865 and https://github.com/hardbyte/python-can/pull/621

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -40,7 +40,7 @@ except ImportError as ie:
     )
 
     class FileLock:
-        """Dummy file lock that do not actually do anything"""
+        """Dummy file lock that does not actually do anything"""
 
         def __init__(self, lock_file, timeout=-1):
             self._lock_file = lock_file
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -15,8 +15,9 @@ import os
 import tempfile
 from collections import deque
 
-from can import Message, CanError, BusABC
 from filelock import FileLock
+
+from can import Message, CanError, BusABC
 
 logger = logging.getLogger(__name__)
 
@@ -28,6 +29,11 @@ except ImportError as ie:
         "python-ics module installed!: %s", ie
     )
     ics = None
+
+
+# Use inter-process mutex to prevent concurrent device open.
+# When neoVI server is enabled, there is an issue with concurrent device open.
+open_lock = FileLock(os.path.join(tempfile.gettempdir(), "neovi.lock"))
 
 
 class ICSApiError(CanError):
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -115,6 +115,7 @@
         'aenum',
         'typing;python_version<"3.5"',
         'windows-curses;platform_system=="Windows"',
+        'filelock'
```
zariiii9003, line 118: filelock should be in neovi under `extras_require`, right?

pierreluctg: Trying to keep this change as close as possible to the original PR https://github.com/hardbyte/python-can/pull/621.

Please open an issue if you think this should be move under `extras_require` and we can address it in 4.x.

pierreluctg: And by the way you are the one that removed it from `extras_require` on develop.
https://github.com/hardbyte/python-can/pull/833

zariiii9003: Yes, i'm aware of that :)

It was both in `extras_require` and `install_requires` so i wrongly assumed it was used elsewhere without checking it. We should avoid installing unnecessary packages though so i would prefer it in `neovi`.

pierreluctg: Done

can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -118,18 +149,28 @@ def __init__(self, channel, can_filters=None, **kwargs):
...
+                if "data_bitrate" in kwargs:
+                    for channel in self.channels:
+                        ics.set_fd_bit_rate(
+                            self.dev, kwargs.get("data_bitrate"), channel
+                        )
+        except ics.RuntimeError as re:
+            logger.error(re)
+            err = ICSApiError(*ics.get_last_api_error(self.dev))
+            try:
+                self.shutdown()
```
zariiii9003, line 171: If there is no `except` clause then you can remove the `try` and `finally`.

zariiii9003 (commented):
Is there a way to release the lock in case the application crashes?

can/interfaces/ics_neovi/neovi_bus.py
pierreluctg: Without the try/finally the exception `err` from line 169 would be lost if there is a exception in `self.shutdown(). Here the interesting exception is the one from line 169.

zariiii9003: If an exception happens in `bus.shutdown()` it is raised immediately since there is no `except` clause. Am i missing something?

pierreluctg: You can try this code bellow to see how this works:

```python
class ErrorOne(Exception):
    pass

class ErrorTwo(Exception):
    pass


def method_one():
    try:
        raise ErrorOne()
    finally:
        raise ErrorTwo()

def method_two():
    raise ErrorOne()

if __name__ == '__main__':
    try:
        method_one()
    except ErrorOne:
        print("Got Error One")
    except ErrorTwo:
        print("Got Error Two")

    try:
        method_two()
    except ErrorOne:
        print("Got Error One")
    except ErrorTwo:
        print("Got Error Two")
```

zariiii9003: Thank you for the example, i have never seen that before. I just found the [documentation](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions) for it too.

hardbyte (approved):

pierreluctg:
> Is there a way to release the lock in case the application crashes?

The answer depends on the platform. I suggest that you take a look at filock ducumentation and source code https://filelock.readthedocs.io/en/latest/.

PR #905, opened 2020-09-12
The gs_usb interface support

jxltom:
The ```gs_usb``` is the interface for Geschwister Schneider USB/CAN devices and bytewerk.org candleLight USB CAN devices such as candlelight, canable, cantact, etc. Refer to https://github.com/torvalds/linux/blob/master/drivers/net/can/usb/gs_usb.c and https://github.com/candle-usb/candleLight_fw

This PR implements the general driver for ```gs_usb``` devices based on ```pyusb```, tested on Windows/Linux/Mac.

It looks like one of the ```gs_usb``` device ```cantact``` is supported in https://github.com/hardbyte/python-can/pull/853. However, that interface might diverge from ```gs_usb``` since it is going to add CAN FD support. This PR is more universal for ```gs_usb``` devices. Since this interface is using pyusb as backend, it has better crossplatform compatibility.

### DIFF
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -10,6 +10,7 @@ import logging
 
 logger = logging.getLogger(__name__)
 
+
 class GsUsbBus(can.BusABC):
     def __init__(self, channel, bus, address, bitrate, can_filters=None, **kwargs):
         """
@@ -53,7 +54,7 @@ class GsUsbBus(can.BusABC):
 
         # Pad message data
         msg.data.extend([0x00] * (CAN_MAX_DLC - len(msg.data)))
-        
+
         frame = GsUsbFrame()
         frame.can_id = can_id
         frame.can_dlc = msg.dlc
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -90,7 +90,7 @@ class GsUsbBus(can.BusABC):
         msg = can.Message()
         msg.arbitration_id = frame.arbitration_id
         msg.dlc = frame.can_dlc
-        msg.data = bytearray(frame.data)[0:msg.dlc]
+        msg.data = bytearray(frame.data)[0 : msg.dlc]
         msg.timestamp = frame.timestamp
         msg.channel = self.channel_info
         msg.is_extended_id = frame.is_extended_id
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -34,7 +34,8 @@ class GsUsbBus(can.BusABC):
         """Transmit a message to the CAN bus.
 
         :param Message msg: A message object.
-        :param timeout: timeout is not used here
+        :param timeout: timeout is not supported.
+            The function won't return until message is sent or exception is raised.
 
         :raises can.CanError:
             if the message could not be sent
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -86,16 +86,18 @@ class GsUsbBus(can.BusABC):
         if not self.gs_usb.read(frame=frame, timeout_ms=timeout_ms):
             return None, False
 
-        msg = can.Message()
-        msg.arbitration_id = frame.arbitration_id
-        msg.dlc = frame.can_dlc
-        msg.data = bytearray(frame.data)[0 : msg.dlc]
-        msg.timestamp = frame.timestamp
-        msg.channel = self.channel_info
-        msg.is_extended_id = frame.is_extended_id
-        msg.is_remote_frame = frame.is_remote_frame
-        msg.is_error_frame = frame.is_error_frame
-        msg.is_rx = True
+        msg = can.Message(
+            timestamp=frame.timestamp,
+            arbitration_id=frame.arbitration_id,
+            is_extended_id=frame.can_dlc,
+            is_remote_frame=frame.is_remote_frame,
+            is_error_frame=frame.is_error_frame,
+            channel=self.channel_info,
+            dlc=frame.can_dlc,
+            data=bytearray(frame.data)[0 : msg.dlc],
+            is_rx=True,
+        )
+
         return msg, False
 
     def shutdown(self):
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -69,7 +69,6 @@ class GsUsbBus(can.BusABC):
         :param float timeout: seconds to wait for a message,
                               see :meth:`~can.BusABC.send`
                               0 and None will be converted to minimum value 1ms.
-                              float will be finally converted to integer.
 
         :return:
             1.  a message that was read or None on timeout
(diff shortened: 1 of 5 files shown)


can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
...
+        msg = can.Message()
+        msg.arbitration_id = frame.arbitration_id
+        msg.dlc = frame.can_dlc
+        msg.data = bytearray(frame.data)[0 : msg.dlc]
+        msg.timestamp = frame.timestamp
+        msg.channel = self.channel_info
+        msg.is_extended_id = frame.is_extended_id
+        msg.is_remote_frame = frame.is_remote_frame
+        msg.is_error_frame = frame.is_error_frame
+        msg.is_rx = True
```
hardbyte, line 99: Can you please pass these as named arguments to the `Message` constructor

doc/interfaces/gs_usb.rst
```diff
@@ -0,0 +1,30 @@
+.. _gs_usb:
+
+CAN driver based on WCID for Geschwister Schneider USB/CAN devices and bytewerk.org candleLight USB CAN interfaces
+==================================================================================================================
+
+Windows/Linux/Mac CAN driver based on WCID for Geschwister Schneider USB/CAN devices and candleLight USB CAN interfaces.
+
+Install: ``pip install "python-can[gs_usb]"``
+
+Usage: pass ``bus`` and ``address`` to open the device. The parameters can be got by ``pyusb``
```
hardbyte, line 10: Please document the complete setup for a first time user?

Assume someone has found the python-can docs and hasn't had any direct experience with `pyusb`.

doc/interfaces/gs_usb.rst
```diff
@@ -0,0 +1,30 @@
+.. _gs_usb:
+
+CAN driver based on WCID for Geschwister Schneider USB/CAN devices and bytewerk.org candleLight USB CAN interfaces
+==================================================================================================================
```
hardbyte, line 4: 💯  for providing docs!

Is there any plan to support message filtering via this interface?
Are there any internal methods that users should know about? E.g. might be worth documenting the `gs_usb` instance attribute.

Please include a link to the upstream gs_usb  GitHub page (or docs)

can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if msg.is_error_frame:
+            can_id = can_id | CAN_ERR_FLAG
+
+        frame = GsUsbFrame()
+        frame.can_id = can_id
+        frame.can_dlc = msg.dlc
+        frame.timestamp_us = int(msg.timestamp * 1000000)
+        msg.data.extend([0x00] * (CAN_MAX_DLC - len(msg.data)))
+        frame.data = list(msg.data)
```
hardbyte, line 55: I'd finish the data processing and then construct your frame:

```suggestion
        # Pad message data
        msg.data.extend([0x00] * (CAN_MAX_DLC - len(msg.data)))
        
        frame = GsUsbFrame()
        frame.can_id = can_id
        frame.can_dlc = msg.dlc
        frame.timestamp_us = int(msg.timestamp * 1000000)
        frame.data = list(msg.data)
```

You are modifying the message which the user may still be using (e.g. storing/displaying etc) I'd like to avoid that, or at the very least clearly mention in the interface docs.

can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self.gs_usb.send(frame)
```
hardbyte, line 57: I imagine this can fail... please catch any usb specific exceptions and raise `can.CanError` (or a subclass)

can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
+from typing import cast, Any, Iterator, List, Optional, Sequence, Tuple, Union
+
+from gs_usb.gs_usb import GsUsb
+from gs_usb.gs_usb_frame import GsUsbFrame
+from gs_usb.constants import CAN_ERR_FLAG, CAN_RTR_FLAG, CAN_EFF_FLAG, CAN_MAX_DLC
+import can
+import usb
+
+
```
hardbyte, line 9: Optionally add logging

can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
...
+        self.gs_usb.set_bitrate(bitrate)
+        self.gs_usb.start()
+
+        super().__init__(channel=channel, can_filters=can_filters, **kwargs)
+
+    def send(self, msg: can.Message, timeout: Optional[float] = None):
+        """Transmit a message to the CAN bus.
+
+        :param Message msg: A message object.
+        :param timeout: timeout is not used here
```
hardbyte, line 34: Ok to ignore the timeout but please specify what the behaviour is

hardbyte (changes requested):
Overall looks very clean, just minor comments mostly around documentation.

Bit of a shame to not have support for periodic messages or filtering.

hardbyte:
@ericevenchick any chance you could provide your thoughts?

ericevenchick:
@jxltom is right, CANtact is diverging a bit from gs_usb to add FD support. There's already a gs_usb_fd Linux kernel module for this: https://github.com/linklayer/gs_usb_fd

The goal was to keep support with the original gs_usb, but that's not my standard to modify. The gs_usb_fd module supports normal gs_usb devices, and FD devices provide a flag to show they have support.

can/interfaces/gs_usb.py
```diff
@@ -0,0 +1,107 @@
...
+        Read a message from the bus and tell whether it was filtered.
+        This methods may be called by :meth:`~can.BusABC.recv`
+        to read a message multiple times if the filters set by
+        :meth:`~can.BusABC.set_filters` do not match and the call has
+        not yet timed out.
+
+        :param float timeout: seconds to wait for a message,
+                              see :meth:`~can.BusABC.send`
+                              0 and None will be converted to minimum value 1ms.
+                              float will be finally converted to integer.
```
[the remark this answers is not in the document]
jxltom: Makes sense, removed

can/interfaces/gs_usb.py
jxltom: Done

doc/interfaces/gs_usb.rst
jxltom: Docs is updated

can/interfaces/gs_usb.py
jxltom: Logging is added for future use

can/interfaces/gs_usb.py
jxltom: usb.core.USBError is catched

can/interfaces/gs_usb.py
jxltom: The docs is updated

doc/interfaces/gs_usb.rst
jxltom: Definitely the docs should be updated.

Sadly the Geschwister Schneider USB/CAN and candleLight USB CAN devices do not support message filtering internally, so message filtering can not be added in this interface.

The docs is updated and more info is provided

jxltom:
Hey @hardbyte thanks for your time for reviewing. I've already updated the PR according to the comments. Just let me know if there are more concerns. Thanks!

hardbyte (approved):
Thanks for making all those changes this looks good to merge.

PR #911, opened 2020-09-25
IXXAT: Add support for 666kbit/s bitrate

henzef:


henzef:
I am a bit unsure about the 666.666kbit/s rounding workaround. Is this an acceptable solution?

christiansandberg (approved):
I think it is an acceptable solution.

hardbyte (approved):

PR #913, opened 2020-09-29
Add support to ETAS Interfaces.

vincent-mailhol:
I would like to add a new module to python-can to support ETAS-BOA interface: https://www.etas.com/en/downloadcenter/18102.php?date=20160421

Thank you for your comments.


vincent-mailhol:
Just realized that I did the pull request against master instead of doing it against development… My bad, I will close and resend later, sorry for the disturbance…

PR #924, opened 2020-10-18
Slcan: fix fileno issue with windows

bmeisels:
Rework of https://github.com/hardbyte/python-can/pull/760 based on changes in https://github.com/hardbyte/python-can/pull/877 and the solution implemented in https://github.com/hardbyte/python-can/pull/902 .

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -10,6 +10,7 @@ Interface for slcan compatible interfaces (win32/linux).
 from typing import Any, Optional, Tuple
 from can import typechecking
 
+import io
 import time
 import logging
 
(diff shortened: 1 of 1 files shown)


hardbyte (approved):

can/interfaces/slcan.py
```diff
@@ -252,10 +253,10 @@ def shutdown(self) -> None:
         self.serialPortOrig.close()
 
     def fileno(self) -> int:
-        if hasattr(self.serialPortOrig, "fileno"):
+        try:
             return self.serialPortOrig.fileno()
-        # Return an invalid file descriptor on Windows
-        return -1
+        except io.UnsupportedOperation:
+            raise NotImplementedError("fileno is not implemented using current CAN bus on this platform")
```
felixdivo, line 259: Fix `black` formatter.

```suggestion
            raise NotImplementedError(
                "fileno is not implemented using current CAN bus on this platform"
            )
```

felixdivo (approved):

PR #931, opened 2020-11-13
updates canalystii

kdschlosser:
adds support for setting a different bitrate for each channel
adds support for supplying the channel number to receive from

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -67,11 +67,18 @@ class BusABC(metaclass=ABCMeta):
     def __str__(self) -> str:
         return self.channel_info
 
-    def recv(self, timeout: Optional[float] = None) -> Optional[Message]:
+    def recv(
+        self,
+        timeout: Optional[float] = None,
+        channel: Optional[int] = None
+    ) -> Optional[Message]:
         """Block waiting for a message from the Bus.
 
         :param timeout:
             seconds to wait for a message or None to wait indefinitely
+
+        :param channel:
+            can interface number. used on the CANalyst II adapter
 
         :return:
             None on timeout or a :class:`Message` object.
@@ -82,9 +89,11 @@ class BusABC(metaclass=ABCMeta):
         time_left = timeout
 
         while True:
-
             # try to get a message
-            msg, already_filtered = self._recv_internal(timeout=time_left)
+            try:
+                msg, already_filtered = self._recv_internal(timeout=time_left, channel=channel)
+            except TypeError:
+                msg, already_filtered = self._recv_internal(timeout=time_left)
 
             # return it, if it matches
             if msg and (already_filtered or self._matches_filters(msg)):
@@ -107,7 +116,9 @@ class BusABC(metaclass=ABCMeta):
                     return None
 
     def _recv_internal(
-        self, timeout: Optional[float]
+        self,
+        timeout: Optional[float],
+        channel: Optional[int] = None
     ) -> Tuple[Optional[Message], bool]:
         """
         Read a message from the bus and tell whether it was filtered.
@@ -135,6 +146,9 @@ class BusABC(metaclass=ABCMeta):
 
         :param float timeout: seconds to wait for a message,
                               see :meth:`~can.BusABC.send`
+
+        :param int channel: optional channel number to receive on.
+                            used for the CANalyst II adapter
 
         :return:
             1.  a message that was read or None on timeout
--- a/can/bus.py
+++ b/can/bus.py
@@ -69,16 +69,12 @@ class BusABC(metaclass=ABCMeta):
 
     def recv(
         self,
-        timeout: Optional[float] = None,
-        channel: Optional[int] = None
+        timeout: Optional[float] = None
     ) -> Optional[Message]:
         """Block waiting for a message from the Bus.
 
         :param timeout:
             seconds to wait for a message or None to wait indefinitely
-
-        :param channel:
-            can interface number. used on the CANalyst II adapter
 
         :return:
             None on timeout or a :class:`Message` object.
@@ -90,10 +86,8 @@ class BusABC(metaclass=ABCMeta):
 
         while True:
             # try to get a message
-            try:
-                msg, already_filtered = self._recv_internal(timeout=time_left, channel=channel)
-            except TypeError:
-                msg, already_filtered = self._recv_internal(timeout=time_left)
+
+            msg, already_filtered = self._recv_internal(timeout=time_left)
 
             # return it, if it matches
             if msg and (already_filtered or self._matches_filters(msg)):
@@ -117,8 +111,7 @@ class BusABC(metaclass=ABCMeta):
 
     def _recv_internal(
         self,
-        timeout: Optional[float],
-        channel: Optional[int] = None
+        timeout: Optional[float]
     ) -> Tuple[Optional[Message], bool]:
         """
         Read a message from the bus and tell whether it was filtered.
@@ -146,9 +139,6 @@ class BusABC(metaclass=ABCMeta):
 
         :param float timeout: seconds to wait for a message,
                               see :meth:`~can.BusABC.send`
-
-        :param int channel: optional channel number to receive on.
-                            used for the CANalyst II adapter
 
         :return:
             1.  a message that was read or None on timeout
--- a/can/interfaces/canalystii.py
+++ b/can/interfaces/canalystii.py
@@ -10,6 +10,21 @@ VCI_USBCAN2 = 4
 STATUS_OK = 0x01
 STATUS_ERR = 0x00
 
+ERR_CAN_OVERFLOW = 0x0001
+ERR_CAN_ERRALARM = 0x0002
+ERR_CAN_PASSIVE = 0x0004
+ERR_CAN_LOSE = 0x0008
+ERR_CAN_BUSERR = 0x0010
+
+ERR_DEVICEOPENED = 0x0100
+ERR_DEVICEOPEN = 0x0200
+ERR_DEVICENOTOPEN = 0x0400
+ERR_BUFFEROVERFLOW = 0x0800
+ERR_DEVICENOTEXIST = 0x1000
+ERR_LOADKERNELDLL = 0x2000
+ERR_CMDFAILED = 0x4000
+ERR_BUFFERCREATE = 0x8000
+
 TIMING_DICT = {
     5000: (0xBF, 0xFF),
     10000: (0x31, 0x1C),
@@ -79,6 +94,28 @@ class VCI_BOARD_INFO(Structure):
         ('str_Serial_Num', c_char * 20),
         ('str_hw_Type', c_char * 40),
         ('Reserved', c_ushort * 4)
+    ]
+
+
+class VCI_CAN_STATUS(Structure):
+    _fields_ = [
+        ('ErrInterrupt', c_ubyte),
+        ('regMode', c_ubyte),
+        ('regStatus', c_ubyte),
+        ('regALCapture', c_ubyte),
+        ('regECCapture', c_ubyte),
+        ('regEWLimit', c_ubyte),
+        ('regRECounter', c_ubyte),
+        ('regTECounter', c_ubyte),
+        ('Reserved', c_ulong)
+    ]
+
+
+class VCI_ERR_INFO(Structure):
+    _fields_ = [
+        ('ErrCode;', c_uint),
+        ('Passive_ErrData', c_byte * 3),
+        ('ArLost_ErrData', c_byte)
     ]
 
 
@@ -253,6 +290,75 @@ class CANalystIIBus(BusABC):
         )
 
         return bool(frames_sent)
+
+    def status(self) -> Dict:
+        """
+        CAN Status
+
+        returns a dictionary with the following keys
+
+            * `err_interrupt`
+            * `mode`
+            * `status`
+            * `al_capture`
+            * `ec_capture`
+            * `ew_limit`
+            * `rx_count`
+            * `tx_count`
+
+        :rtype: dict
+        """
+        status = VCI_CAN_STATUS()
+        try:
+            CANalystII.VCI_ReadCANStatus(
+                VCI_USBCAN2,
+                self.device,
+                self.channel,
+                byref(status)
+            )
+
+        except:
+            pass
+
+        return dict(
+            err_interrupt=status.ErrInterrupt,
+            mode=status.regMode,
+            status=status.regStatus,
+            al_capture=status.regALCapture,
+            ec_capture=status.regECCapture,
+            ew_limit=status.regEWLimit,
+            rx_count=status.regRECounter,
+            tx_count=status.regTECounter
+        )
+
+    def error_info(self) -> Dict:
+        """
+        CAN Error
+
+        returns a dictionary with the following keys
+
+        `error_code`
+        `passive_err_data`
+        `ar_lost_err_data`
+
+        :rtype: dict
+        """
+        error_info = VCI_ERR_INFO()
+        try:
+            CANalystII.VCI_ReadErrInfo(
+                VCI_USBCAN2,
+                self.device,
+                self.channel,
+                byref(error_info)
+            )
+        except:
+            pass
+
+        return dict(
+            error_code=error_info.ErrCode,
+            passive_err_data=error_info.Passive_ErrData,
+            ar_lost_err_data=error_info.ArLost_ErrData
+        )
 
     def available(self) -> int:
         """
--- a/can/interfaces/canalystii.py
+++ b/can/interfaces/canalystii.py
@@ -392,7 +392,7 @@ class CANalystIIBus(BusABC):
         """
         return {key: value for key, value in self._b_info.items()}
 
-    def send(self, msg: Message, timeout=None):
+    def send(self, msg: Message, timeout=None) -> None:
         """
         Send CAN Frame
 
@@ -402,8 +402,9 @@ class CANalystIIBus(BusABC):
         :param timeout: timeout is not used here
         :type timeout: None
 
-        :return: `True` if frame was sent else `False`
-        :rtype: bool
+        :rtype: None
+
+        :raises: `CanError` if unable to send the message
         """
 
         can_objs = (VCI_CAN_OBJ * 1)()
@@ -413,6 +414,7 @@ class CANalystIIBus(BusABC):
         can_objs[0].SendType = 1
         can_objs[0].RemoteFlag = int(msg.is_remote_frame)
         can_objs[0].DataLen = msg.dlc
+
         for j in range(len(msg.data)):
             can_objs[0].Data[j] = msg.data[j]
 
@@ -424,7 +426,14 @@ class CANalystIIBus(BusABC):
             1
         )
 
-        return bool(frames_sent)
+        if not frames_sent:
+            raise CanError(
+                'Unable to send message '
+                '(device: {0}, channel: {1})'.format(
+                    self.device,
+                    self.channel
+                )
+            )
 
     def status(self) -> Dict:
         """
@@ -534,10 +543,10 @@ class CANalystIIBus(BusABC):
 
         return message, False
 
-    def flush_tx_buffer(self):
+    def flush_tx_buffer(self) -> None:
         CANalystII.VCI_ClearBuffer(VCI_USBCAN2, self.device, self.channel)
 
-    def shutdown(self):
+    def shutdown(self) -> None:
         # stops receiving new frames
         CANalystII.VCI_ResetCAN(VCI_USBCAN2, self.device, self.channel)
 
--- a/can/interfaces/canalystii.py
+++ b/can/interfaces/canalystii.py
@@ -125,8 +125,9 @@ class CANalystIIBus(BusABC):
 
         self.channel = channel
         self.device = device
-        self.channel_info = "CANalyst-II: device {}, channels {}".format(
-            self.device, self.channel
+        self.channel_info = "CANalyst-II: device {}, channel {}".format(
+            self.device,
+            self.channel
         )
 
         if bitrate is not None:
@@ -145,10 +146,8 @@ class CANalystIIBus(BusABC):
                 pass
 
             b_info = VCI_BOARD_INFO()
-            CANalystII.VCI_ReadBoardInfo(VCI_USBCAN2, device, byref(b_info))
-
-            if CANalystII.VCI_OpenDevice(VCI_USBCAN2, self.device, 0) == STATUS_ERR:
-                logger.error("VCI_OpenDevice Error")
+            if CANalystII.VCI_ReadBoardInfo(VCI_USBCAN2, device, byref(b_info)) == STATUS_ERR:
+                logger.error("VCI_ReadBoardInfo Error")
 
             self._b_info = dict(
                 hardware_version=b_info.hw_Version,
@@ -160,6 +159,9 @@ class CANalystIIBus(BusABC):
                 serial_number=str(b_info.str_Serial_Num, encoding='ascii'),
                 hardware_type=str(b_info.str_hw_Type, encoding='ascii')
             )
+
+            if CANalystII.VCI_OpenDevice(VCI_USBCAN2, self.device, 0) == STATUS_ERR:
+                raise ValueError('Unable to open interface, the device number supplied may be incorrect')
 
             _in_use[device] = []
         else:
@@ -185,9 +187,8 @@ class CANalystIIBus(BusABC):
         )
 
         if status == STATUS_ERR:
-            logger.error("VCI_InitCAN Error ({0})".format(channel))
             self.shutdown()
-            return
+            raise ValueError("VCI_InitCAN Error, incorrect channel number?")
 
         status = CANalystII.VCI_StartCAN(
             VCI_USBCAN2,
@@ -196,7 +197,7 @@ class CANalystIIBus(BusABC):
         )
 
         if status == STATUS_ERR:
-            logger.error("VCI_StartCAN Error ({0})".format(channel))
+            logger.error("VCI_StartCAN Error, device:{0} channel:{1}".format(device, channel))
             self.shutdown()
             return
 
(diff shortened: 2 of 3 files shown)


kdschlosser:
Here is an example of how it works.

```python

import can.interfaces.canalystii
from can import Message
can = can.interfaces.canalystii.CANalystIIBus(channel=[0, 1], bitrate=[500000, 500000])

print(can.board_info())

messages = []
for i in range(1, 11):
    data = bytearray([0x01 * i, 0x02 * i, 0x03 * i, 0x04 * i, 0x05 * i, 0x01, 0x01, 0x01])
    message = Message(arbitration_id=i, is_extended_id=False, channel=i % 2, dlc=8, data=data, is_rx=False)
    messages += [message]

can.send(messages)

while True:
    msg1 = can.recv(timeout=1, channel=0)
    if msg1 is not None:
        d = ' '.join(hex(item)[2:].upper() for item in msg1.data)
        print('channel:', 0)
        print('id:', msg1.arbitration_id)
        print('data length:', msg1.dlc)
        print('data:', d)
        print()

    msg2 = can.recv(timeout=1, channel=1)
    if msg2 is not None:
        d = ' '.join(hex(item)[2:].upper() for item in msg2.data)
        print('channel:', 1)
        print('id:', msg2.arbitration_id)
        print('data length:', msg2.dlc)
        print('data:', d)
        print()

    if msg1 is None and msg2 is None:
        break
```

and the output is

```
{'hardware_version': 2304, 'firmware_version': 0, 'driver_version': 2304, 'library_version': 2308, 'irq': 0, 'num_can_interfaces': 2, 'serial_number': '', 'hardware_type': ''}
channel: 0
id: 1
data length: 8
data: 1 2 3 4 5 1 1 1

channel: 1
id: 2
data length: 8
data: 2 4 6 8 A 1 1 1

channel: 0
id: 3
data length: 8
data: 3 6 9 C F 1 1 1

channel: 1
id: 4
data length: 8
data: 4 8 C 10 14 1 1 1

channel: 0
id: 5
data length: 8
data: 5 A F 14 19 1 1 1

channel: 1
id: 6
data length: 8
data: 6 C 12 18 1E 1 1 1

channel: 0
id: 7
data length: 8
data: 7 E 15 1C 23 1 1 1

channel: 1
id: 8
data length: 8
data: 8 10 18 20 28 1 1 1

channel: 0
id: 9
data length: 8
data: 9 12 1B 24 2D 1 1 1

channel: 1
id: 10
data length: 8
data: A 14 1E 28 32 1 1 1
```

can/bus.py
```diff
@@ -67,12 +67,19 @@ def __init__(
     def __str__(self) -> str:
         return self.channel_info
 
-    def recv(self, timeout: Optional[float] = None) -> Optional[Message]:
+    def recv(
+        self,
+        timeout: Optional[float] = None,
+        channel: Optional[int] = None
+    ) -> Optional[Message]:
```
hardbyte, line 74: Most of the benefit of python-can is the abstraction between the network interfaces via `Bus` and higher level code. Changing the `Bus` interface shouldn't be part of this PR.

hardbyte (changes requested):

can/bus.py
kdschlosser: Then how would you go about changing the receive so that it is able to work on both channel 0 and channel 1 and not just channel 0?

kdschlosser: one other thing is the inability to specify the number of messages wanting to get received. with the canalyst-ii adapter there is an api function to obtain the number of messages in waiting and this can be used to create the array that is passed to the receive function of the array. the api will fill the array emptying the internal buffer on the device. The way it is currently done is inefficient and can lead to running out of available buffer.

Maybe it would be a better to add in \*\*kwargs as a parameter. this would allow any subclasses the flexibility to add additional parameters if they are needed.

hardbyte: My preference is to keep the `**kwargs` for just the Bus initializer. Several interfaces add methods to `Bus` though, that would be okay. I'd suggest adding an internal bus state such as `recv_channel` and a setter method to modify it. Another method would also work for peeking at the number of messages in the recv buffer.

Is there support for receiving from all `Bus` channels which might make a better default in `_recv_internal`?

kdschlosser: it can be done by calling the API's receive method once for each channel. But this to will break API with `_recv_internal` because it can no longer return a message it would have to return a list or a tuple of messages.

I can override the recv method because the handling of the timeout is also not ideal because of the loop it makes and the repeated called to receive. The API already has a concession for supplying a timeout. This may be the best way to go about it and provide a channel parameter.

I had added it to the Bus class because I would have thought that there has to be more then just this one interface that has more then a single channel.

Let me make a change and see if it is something you like better.

christiansandberg:
What's the difference from creating two instances, one for each channel? How would you make sure that both channels are received from continuously without risking missing messages while you wait for messages from the other channel?

kdschlosser:
@christiansandberg 

These devices do have a hardware buffer that can hold a few thousand frames. using multiple instances of the class, one for each channel has it's own set of complications because of how the API is set up.

I also added the method `available(channel)` which allows the user to check a channel to see if there are wny messages available.

```python
while True:
    if adapter.available(0):
        msg = adapter.recv(channel=0)
    if adapter.available(1):
        msg = adapter.recv(channel=1)
```

something like this could also be done if using a separate thread to process the received frame.

```python
while True:
    while adapter.available(0):
        msg = adapter.recv(channel=0)
        process_frame(msg)
    while adapter.available(1):
        msg = adapter.recv(channel=1)
        process_frame(msg)
```

christiansandberg:
The first code sample has a problem in that there is no place where the code will block which means your application will use up all CPU resources while doing pretty much nothing.

The other sample has another problem because if no messages are received on one channel, you will not see the messages from the other channel.

If you use python-can as intended, possibly with the Notifier class, your application will be both responsive and only use the CPU when needed.

christiansandberg:
Here's an example on how you could rewrite your code:

```python
import can.interfaces.canalystii
from can import Message, Notifier

def process_frame(msg):
    print(msg)

can0 = can.interfaces.canalystii.CANalystIIBus(channel=0, bitrate=500000)
can1 = can.interfaces.canalystii.CANalystIIBus(channel=1, bitrate=500000)
notifier = can.Notifier([can0, can1], [process_frame])

time.sleep(10)

notifier.stop()
can0.shutdown()
can1.shutdown()
```

kdschlosser:
@christiansandberg 

It can be done that way but then it becomes the libraries job to hold reference to the created instances so that functions like the ones listed below that are in the API do not get called more then a single time. If they do it would cause problems for all instances related to that adapter.

`CANalystII.VCI_OpenDevice(VCI_USBCAN2, self.device, 0)`
`CANalystII.VCI_UsbDeviceReset(VCI_USBCAN2, self.device, 0)`
`CANalystII.VCI_CloseDevice(VCI_USBCAN2, self.device)`

The library would have to track what is open and what is closed to be able to know if it should run those interface specific commands or not. It would require quite a bit of additional code in the library to be able to do. and the end result is you will still have the same issue as what you have stated in the previous post about missing a can frame.

If there is a problem with frames getting dropped then the user can create 2 threads to handle the incoming frames, one to collect them and another to process the received frames.

There is also the issue of `can.notifier.Notifier` creating a thread to handle each bus. I do not believe that the API is going to play nice with `CANalystII.VCI_Receive` getting called from 2 different threads at the same time. Because of the limited (almost non existent) documentation about the API we cannot make assumptions as to whether or not the API had been made thread safe.

The use of `can.notifier.Notifier` would not be able to be used and instead code like the following would ned to be used by the user.

<details>
  <summary>Code Example (click to expand)</summary>

```python
import can.interfaces.canalystii
import threading
from collections import deque


class ProcessThread(threading.Thread):

    def __init__(self):
        # allows thread to exit gracefully
        self._exit_event = threading.Event()
        # stop spinning wheel in the thread if 
        # there are no messages to be processed
        self._recv_event = threading.Event()
        # high speed data queue
        self._recv_queue = deque()

        threading.Thread.__init__(self)

    def run(self):
        while not self._exit_event.is_set():
            self._recv_event.wait()
            self._recv_event.clear()
            
            while len(self._recv_queue):
                message = self._recv_queue.popleft()
                print(message.timestamp)
                print('channel:', message.channel)
                print('id:', hex(message.arbitration_id)[2:].upper().zfill(8))
                print('extended frame:', message.is_extended)
                data = ' '.join(
                    hex(message.data[i])[2:].upper().zfill(2) 
                    for i in range(message.dlc)
                )
                print('data:', data)
    
    def stop(self):
        self._recv_queue.clear()
        self._exit_event.set()
        self._recv_event.set()
        
    def add(self, message):
        self._recv_queue.append(message)
        self._recv_event.set()
    

process_thread = ProcessThread()

can_bus = can.interfaces.canalystii.CANalystIIBus([0, 1], bitrate=[500000, 33330])

try:
    while True:
        msg = can_bus.recv(timeout=0.01, channel=0)
        if msg is not None:
            process_thread.add(msg)
            
        msg = can_bus.recv(timeout=0.01, channel=1)
        if msg is not None:
            process_thread.add(msg)
            
except KeyboardInterrupt:
    can_bus.shutdown()
    process_thread.stop()
```

</details>

kdschlosser:
ideally the recv should return a list of collected messages and the number of messages to receive should be user settable via a parameter for the recv method. Receiving one at a time is really inefficient, the library was written to return a message from the recv method and not a list of messages. So to keep with the libraries API the recv method only collects a single message so the return value is the same across the library.

when dealing with high speed CAN networks there is the possibility of having a receive buffer overrun by collecting only a single message at a time even using the code example in my last post.

christiansandberg:
Ok, I see your point. In my view combining multiple channels in one bus only makes sense if you can receive from them simulateously, like the Vector interface. In this case I would say it was a mistake trying to combine them in the first place. I think it would have been better to let the interface keep a reference count of devices in use so it can open and close them once.

About receiving multiple messages, you can certainly do that optimization in your interface. Read multiple messages to a buffer and when recv is called and there is still messages in the buffer you can return them one by one until the buffer is empty. In any case the messages must be processed one by one so the small overhead of an extra function call in Python is neglible.

kdschlosser:
ok cool. so long as we are on the same page.. It can be done your way, and if that is the way that it is wanted then I can do that. It is going to add additional code in order to open the interface is none exist and to close the interface when there are no more channels that are open. I personally like separating the can channels. Technically speaking a channel is a bus and it should only encapsulate a single channel and not the entire interface.

Maybe keeping track of the interfaces that are in use as well as the channels in use on those interfaces is the better approach. It keeps the API proper in the library.

It wouldn't be all that difficult to keep track of the devices and channels that are in use.

can/interfaces/canalystii.py
```diff
@@ -69,132 +57,355 @@ class VCI_CAN_OBJ(Structure):
...
+        Timing0: Optional[int] = None,
+        Timing1: Optional[int] = None,
         can_filters=None,
         **kwargs,
     ):
         """
+        CANalystII Interface
+
+        :param channel: channel number to use
+        :type channel: int
```
christiansandberg, line 140: I don't think you need to specify the type here since you added types to the interface.

can/interfaces/canalystii.py
```diff
@@ -69,132 +57,355 @@ class VCI_CAN_OBJ(Structure):
...
-        :param device: device number
-        :param bitrate: CAN network bandwidth (bits/s)
-        :param Timing0: customize the timing register if bitrate is not specified
-        :param Timing1:
         :param can_filters: filters for packet
         """
-        super().__init__(channel=channel, can_filters=can_filters, **kwargs)
 
-        if isinstance(channel, (list, tuple)):
-            self.channels = channel
```
christiansandberg, line 95: Since this is a breaking change (which I think is okay since next is a major release and it was a bug that you could only read from the first channel) it would probably be good to raise a helpful exception about multiple channels not being supported anymore.

can/interfaces/canalystii.py
```diff
@@ -69,132 +57,355 @@ class VCI_CAN_OBJ(Structure):
...
+            else:
+                raise ValueError("Bitrate is not supported ({0})".format(bitrate))
 
-        if Timing0 is None or Timing1 is None:
+        if None in (Timing0, Timing1):
             raise ValueError("Timing registers are not set")
 
+        if device not in _in_use:
+            try:
+                CANalystII.VCI_UsbDeviceReset(VCI_USBCAN2, self.device, 0)
```
christiansandberg, line 181: What does this do? Is it necessary? Maybe write a comment why this was added.

christiansandberg (commented):
I think this looks better!

can/interfaces/canalystii.py
kdschlosser: Does sphinx pick up on the data types when compiling the documentation if the type is not supplied in the docstring? I was not sure of the answer to that question so I simply included the type in the docstring thinking that it wouldn't hurt at all.

can/interfaces/canalystii.py
kdschlosser: There are more then one variation of the Canalyst-II devices. Some of them being clones. Because the dll and the so file are not being included with this library that means any one of the different variations can be used. I have found that not all of them have the UsbDeviceReset function. But I have found that the ones that do seem to play a little nicer when the device gets reset just before setting it up.

can/interfaces/canalystii.py
kdschlosser: fair enough

kdschlosser:
I have a question as to how the arbitration id's are stored in the Message class. When dealing with extended frames I have seen the arbitration ids stored in 2 ways. the first way is `id | 0x80000000` (32bit) the second way is `id & 0x7FFFFFFF` (29bit) essentially I am asking if I should store the id as 32 bit id or the 29 bit id.

christiansandberg:
The `arbitration_id` should only contain the 29-bit ID. The `is_extended_id` is used to flag that it is an extended ID.

can/interfaces/canalystii.py
```diff
@@ -69,132 +58,494 @@ class VCI_CAN_OBJ(Structure):
...
-        )
 
-        if msg.channel is not None:
-            channel = msg.channel
-        elif len(self.channels) == 1:
-            channel = self.channels[0]
+        can_objs = (VCI_CAN_OBJ * 1)()
+
+        arbitration_id = msg.arbitration_id
+        if arbitration_id & 0x80000000:
```
christiansandberg, line 409: Only `msg.is_extended_id` is used.

kdschlosser: my point to this is an unknowing user could place say an arbitration id of `0x9003E099` and not set the is_extended flag, with no checks in place to see if the arbitration id is a 32 bit id then it can cause problems for the user making it difficult to nail down the actual issue.

`arbitration_id = 0x9003E099 & 0x7FFFFFFF  # 0x1003E099`
`is_extended = bool(0x9003E099 & 0x80000000)`


a small modification of the Message class would eliminate any possibility of having an arbitration is set that is an extended id. and you can dd a property to get the full arbitration id including the set bit for extended.

```python

class Message:
    """
    The :class:`~can.Message` object is used to represent CAN messages for
    sending, receiving and other purposes like converting between different
    logging formats.

    Messages can use extended identifiers, be remote or error frames, contain
    data and may be associated to a channel.

    Messages are always compared by identity and never by value, because that
    may introduce unexpected behaviour. See also :meth:`~can.Message.equals`.

    :func:`~copy.copy`/:func:`~copy.deepcopy` is supported as well.

    Messages do not support "dynamic" attributes, meaning any others than the
    documented ones, since it uses :attr:`~object.__slots__`.
    """

    __slots__ = (
        "timestamp",
        "arbitration_id",
        "is_extended_id",
        "is_remote_frame",
        "is_error_frame",
        "channel",
        "dlc",
        "data",
        "is_fd",
        "is_rx",
        "bitrate_switch",
        "error_state_indicator",
        "__weakref__",  # support weak references to messages
    )

    def __init__(
        self,
        timestamp: float = 0.0,
        arbitration_id: int = 0,
        is_extended_id: bool = True,
        is_remote_frame: bool = False,
        is_error_frame: bool = False,
        channel: Optional[typechecking.Channel] = None,
        dlc: Optional[int] = None,
        data: Optional[typechecking.CanData] = None,
        is_fd: bool = False,
        is_rx: bool = True,
        bitrate_switch: bool = False,
        error_state_indicator: bool = False,
        check: bool = False,
    ):
        """
        To create a message object, simply provide any of the below attributes
        together with additional parameters as keyword arguments to the constructor.

        :param check: By default, the constructor of this class does not strictly check the input.
                      Thus, the caller must prevent the creation of invalid messages or
                      set this parameter to `True`, to raise an Error on invalid inputs.
                      Possible problems include the `dlc` field not matching the length of `data`
                      or creating a message with both `is_remote_frame` and `is_error_frame` set to `True`.

        :raises ValueError: iff `check` is set to `True` and one or more arguments were invalid
        """
        self.timestamp = timestamp

         if arbitration_id & 0x80000000:
             self.is_extended_id = True
             arbitration_id = arbitration_id  & 0x7FFFFFFF 
         else:
             self.is_extended_id = is_extended_id

        self.arbitration_id = arbitration_id
        self.is_remote_frame = is_remote_frame
        self.is_error_frame = is_error_frame
        self.channel = channel
        self.is_fd = is_fd
        self.is_rx = is_rx
        self.bitrate_switch = bitrate_switch
        self.error_state_indicator = error_state_indicator

        if data is None or is_remote_frame:
            self.data = bytearray()
        elif isinstance(data, bytearray):
            self.data = data
        else:
            try:
                self.data = bytearray(data)
            except TypeError:
                err = "Couldn't create message from {} ({})".format(data, type(data))
                raise TypeError(err)

        if dlc is None:
            self.dlc = len(self.data)
        else:
            self.dlc = dlc

        if check:
            self._check()

    @property
    def full_arbitration_id(self):
        if self.is_extended:
            return self.arbitration_id | 0x80000000
        return self.arbitration_id

```

christiansandberg: I believe that identifying an extended message using a specific bit in the CAN ID is implementation specific and not part of the standard. Or is there some part of the standard that I've missed?

kdschlosser: you must have missed it.


image


Standard Frame CAN 2.0A
image


Field name | Length (bits) | Purpose
-- | -- | --
Start-of-frame | 1 | Denotes the start of frame transmission
11 bit Identifier | 11 | A (unique) identifier which also represents the message priority
Remote transmission request (RTR)| 1 | Must be dominant (0) for data frames and recessive (1) for remote request frames 
Identifier extension bit (IDE) | 1 | Must be dominant (0) for base frame format with 11-bit identifiers
Reserved bit (r0) | 1 | Reserved bit.  Must be dominant (0), but accepted as either dominant or recessive.
Data length code (DLC) | 4 | Number of bytes of data (0–8 bytes)[a]
Data field| 0–64 (0-8 bytes) | Data to be transmitted (length in bytes dictated by DLC field)
CRC | 15 | Cyclic redundancy check
CRC delimiter | 1 | Must be recessive (1)
ACK slot | 1 | Transmitter sends recessive (1) and any receiver can assert a dominant (0)
ACK delimiter | 1 | Must be recessive (1)
End-of-frame (EOF) | 7 | Must be recessive (1)


Extended Frame CAN 2.0B
image

Field name | Length (bits) | Purpose
-- | -- | --
Start-of-frame | 1 | Denotes the start of frame transmission
11 bit Identifier | 11 | First part of the (unique) identifier which also represents the message priority
Substitute remote request (SRR) | 1 | Must be recessive (1)
Identifier extension bit (IDE) | 1 | Must be recessive (1) for extended frame format with 29-bit identifiers
18 bit Identifier | 18 | Second part of the (unique) identifier which also represents the message priority
Remote transmission request (RTR) | 1 | Must be dominant (0) for data frames and recessive (1) for remote request frames
Reserved bits (r1, r0) | 2 | Reserved bits which must be set dominant (0), but accepted as either dominant or recessive
Data length code (DLC) | 4 | Number of bytes of data (0–8 bytes)[a]
Data field | 0–64 (0-8 bytes) | Data to be transmitted (length dictated by DLC field)
CRC | 15 | Cyclic redundancy check
CRC delimiter | 1 | Must be recessive (1)
ACK slot | 1 | Transmitter sends recessive (1) and any receiver can assert a dominant (0)
ACK delimiter | 1 | Must be recessive (1)
End-of-frame (EOF) | 7 | Must be recessive (1)


we are interested in "Identifier extension bit (IDE)" which is the bit that gets toggled to on if the frame is an extended frame.
you also have the message priority that is embedded in the 11 bit Identifier.

Which I made a boo boo when I did the last commit for this PR. I added to many F''s and 0's to the checking for an extended frame.

christiansandberg: I just see that the IDE bit is completely separate from the identifier. Why bit 31? What do the other two bits map to? In case of an 11-bit ID, where are the rest of the bits?

Aren't you just referring to details in the socketcan or Vector implementations?

kdschlosser: It doesn't really matter i guess, because of how dbc files set the extended bit is different from what the can specification is and can arbitration ids located on the internet for dealing with vehicles being posted different ways and dbc files made for vehicles being made incorrectly there is no easy way to know what is correct.

christiansandberg: That's just an internal representation in dbc files (and other Vector specific formats and APIs) but their user interfaces always display the correct ID without the extra highest bits. So we shouldn't expose these non-standard details in our API either.

can/interfaces/canalystii.py
```diff
@@ -69,132 +58,488 @@ class VCI_CAN_OBJ(Structure):
...
+
+        can_objs = (VCI_CAN_OBJ * 1)()
+
+        can_objs[0].ID = msg.arbitration_id
+        can_objs[0].ExternFlag = int(msg.is_extended_id)
+        can_objs[0].SendType = 1
+        can_objs[0].RemoteFlag = int(msg.is_remote_frame)
+        can_objs[0].DataLen = msg.dlc
+        for j in range(msg.dlc):
+            can_objs[0].Data[j] = msg.data[j]
```
christiansandberg, line 414: This will not work for remote frames. Don't use dlc to determine length of data.

can/interfaces/canalystii.py
```diff
@@ -69,132 +58,488 @@ class VCI_CAN_OBJ(Structure):
...
+            1
         )
 
-        if msg.channel is not None:
-            channel = msg.channel
-        elif len(self.channels) == 1:
-            channel = self.channels[0]
-        else:
-            raise ValueError("msg.channel must be set when using multiple channels.")
+        return bool(frames_sent)
```
christiansandberg, line 424: Raise an exception if frame could not be sent so the behavior is same as other interfaces.

kdschlosser: you are looking at a stale commit with the raising of a CanError

can/interfaces/canalystii.py
kdschlosser: Ok this is fixed in the last commit.

can/interfaces/canalystii.py
christiansandberg: The latest version of the send method seem to return a boolean. Do you mean that you have changed to raising an exception? In that case I can't see that change for some reason.

kdschlosser: I thought you were wanting me to change this
```python
else:
    raise ValueError("msg.channel must be set when using multiple channels.")
```
to a can error.

I can change the other to raising an exception if wanted. I would like to make a single point before I do, you may change your mind on it. Exception catching is expensive compared to boolean evaluation.

I used this script
```python
import timeit

code1 = """\
test = True

def send():
    return test

while not send():
    pass

while not send():
    pass

"""

code2 = """\
test = True

def send():
    if not test:
        raise RuntimeError()

try:
    send()
except RuntimeError:
    pass

"""

print(timeit.timeit(code1, number=1000000))
print(timeit.timeit(code2, number=1000000))
```

and these are the results for running each one 1000000 times
boolean = 0.337939405
exception = 0.42874530899999996

running this test 1000000 times may seem excessive, it's not. take a vehicle for instance, say there are 20 OBDII PIDS that I want to collect data from every 5 milliseconds. that's 4000 messages a second, 240000 messages a minute and after only 5 minutes of the application running we are at 1200000. with a time difference of 90 milliseconds that ends up costing 1.296 seconds every hour the application runs. after 24 hours it's at 31 seconds.

using the exception catching way you end up with a double whammy because you need to evaluate the returned value in order to decide whether or not you want to raise the exception nd in the user code exception catching needs to be put into place, where as we are not evaluating anything in the send method.

I know it seems trivial but it really adds up over time

christiansandberg: Perhaps so but that would be a major change to the API which many applications rely upon. That needs to be discussed in a separate issue.

kdschlosser: I understand, thought it worth mentioning.

When I first got into programming I was writing firmware for micro controllers, and because of the extremely limited resources available on those things code had to be written with resources being the #1 thing to pay attention to. Plus I hate it when an application does not have it's code optimized (cough.. firefox, cough..chrome) and is a resource hog when it shouldn't be. One of the reasons why I do not like Java, gobbles up memory and doesn't give it back to the OS when it is no longer in use.

I do what I can to keep things running as efficiently as possible and sometimes that requires writing it over and over again and lots of testing to get it dialed in for the proper balance of memory consumption vs CPU use.

christiansandberg: Same here, I also come from an embedded background, but sometimes I need to remind myself that it is a high level language after all and it is not designed to be fast.

christiansandberg (approved):
Just have a look at the CI errors. You probably need to reformat with black. Thank you for your patience!

hardbyte:
😕

kdschlosser:
I have decided that I am going to use the code changes I have made for my own personal use. I wanted to offer the changes I have made so others could benefit from them, but it seem that there is a conflict in programming ideals. I am not one for wasting resources unnecessarily, it leads to more applications like Firefox and Chrome which consume resources without regard to other applications running on the same machine.

I have explained the impact caused by some of the code and why it is not a good way to go about it and how it could potentially lead to missed frames. If I make changes to the code I have written that makes it slower without providing some kind of a benefit  it goes against my ideals and it is not something that I am willing to do. 

I spend quite a bit of time testing the code I write for efficiency and changing it so it runs faster and uses less resources. It is a pride thing I guess, I hope you can understand that....

Issue #934, opened 2020-11-14
timestamp inaccuracy on Windows


danielhrisca:
The function ``time.time`` is not reliable for setting the time stamps on Windows. 

Even if the resolution is good enough, the value returned by ``time.time`` is updated according to the interval Windows timer (see https://stackoverflow.com/questions/1938048/high-precision-clock-in-python)

On my PC this defaults to 10ms steps which in my opinion is a very high inaccuracy.

christiansandberg:
Which context are you referring to? Message timestamps are derived by hardware when possible, but some interfaces simply don't provide this in which case we use time.time() as a fallback.

danielhrisca:
I was looking for example at the Vector implementation were and offset is computed based on ``time.time``

christiansandberg:
That's just to convert from relative to absolute timestamps. There will obviously be an absolute error due the fact that we don't read the hardware time and PC time simultaneously but relative accuracy between messages will still be very accurate.

danielhrisca:
I my use case I need to synchronize the CAN messages with some other device data, using the PC time as reference. How could I do this as accurately as possible?

christiansandberg:
I don't know a better way unfortunately. Maybe you could measure the error somehow and compensate with an offset. Feel free to improve it!

hardbyte:
The resolution of `time.time` using a recent Python on a recent windows is <1ms - see https://www.python.org/dev/peps/pep-0564/#windows but if you want to suggest an improvement feel free.

danielhrisca:
@hardbyte the resolution is good but the value is refreshed at best every ~0.5ms if you change the Windows timer interval. By default it is set at ~10ms. Running the test code from the linked SO thread will show this clearly

hardbyte:
Hmm I got ~1ms (not ~10ms) on a windows laptop without changing anything. I'm not convinced we can do much better from python-can?

I ran this:

```python
import time

# measure the smallest time delta by spinning until the time changes
def measure():
    t0 = time.time()
    t1 = t0
    while t1 == t0:
        t1 = time.time()
    return t1-t0

samples = [measure() for i in range(30)]

for s in samples:
    print(f'time delta: {s:.12f} seconds') 
```

I see `time delta: 0.000000238419 seconds` on Linux and `time delta: 0.001001238419 seconds` on Windows.

danielhrisca:
@hardbyte @christiansandberg 
I've send PR #936  to improve the accuracy for Vector and Kvaser buses

PR #936, opened 2020-11-15
Windows time accuracy

danielhrisca:
Improve time stamp accuracy of the CAN messages for Kvaser and Vector buses on the Windows platform. (see #934)

On Windows the value returned by `time.time` is refreshed at best every 0.5ms, and on some PCs the default refresh rate is every 10ms.

The idea is to get the value of `time.perf_counter` at the moment when `time.time` is refreshed and use this as a reference point.


can/util.py
```diff
@@ -326,6 +327,26 @@ def rename_kwargs(func_name, kwargs, aliases):
                 warnings.warn("{} is deprecated".format(alias), DeprecationWarning)
 
 
+def time_perfcounter_correlation():
+    """ on Windows platform this will help with the accuracy of using the time.time fucntion.
```
hardbyte, line 331:
```suggestion
    """Get the `perf_counter` value nearest to when time.time() is updated on Windows.
    
    On non Windows platforms the current time and perf_counter is directly returned.
    
    Note this value is based on when `time.time()` is observed to update from Python, it is not directly returned by the operating system.
```

can/util.py
```diff
@@ -326,6 +327,26 @@ def rename_kwargs(func_name, kwargs, aliases):
...
+
+    Returns
+    -------
+    t, performance_counter : (time.time, float)
+        time.time value and perf_counter value when the time.time is updated
+
+    """
```
hardbyte, line 338:
```suggestion
    :return: (t, performance_counter) time.time value and perf_counter value when the time.time is updated
    """
```

can/util.py
```diff
@@ -326,6 +327,26 @@ def rename_kwargs(func_name, kwargs, aliases):
...
+    if platform.system() == "Windows":
+        t0 = time()
+        while 1:
```
hardbyte, line 341:
```suggestion
        while True:
```

hardbyte (approved):
Thanks for the PR. I'd like a basic test of the added utility function - just check it returns results of the expected types and ensure it always returns within say 500ms. 

Might be worth looking at the docs for those interfaces and adding a note too.

Is there any harm in removing the platform check and including this correction for all operating systems? It can only help other platforms right?

christiansandberg:
Or query the resolution using `get_clock_info`. If it's too low do the compensation.

zariiii9003:
Don't you have the same `time()`  inaccuracies in the correlation step `t1, performance_counter = time(), perf_counter()`?

hardbyte (approved):
Good stuff.

🕐

christiansandberg (approved):

can/interfaces/vector/canlib.py
```diff
@@ -295,11 +300,22 @@ def __init__(
         # Calculate time offset for absolute timestamps
         offset = xlclass.XLuint64()
         try:
-            try:
-                xldriver.xlGetSyncTime(self.port_handle, offset)
-            except VectorError:
-                xldriver.xlGetChannelTime(self.port_handle, self.mask, offset)
-            self._time_offset = time.time() - offset.value * 1e-9
+            if time.get_clock_info("time").resolution > 1e-5:
```
rliebscher, line 303: There is a lot of code duplication. Why not wrapping it in an object like this in can_util:
```
class PreciseTime:
    def __init__(self):
            # use this if the resolution is higher than 10us
            if get_clock_info("time").resolution > 1e-5:
                t0 = time()
                while True:
                    t1, performance_counter = time(), perf_counter()
                    if t1 != t0:
                        break
                self.time = t1
                self.perfcounter = performance_counter
            else:
                self.time = time()
                self.perfcounter = None # not needed

    def now(self):
        if self.perf_counter is None:
            return time()
        return self.time + (perf_counter() - self.perfcounter)
```
Then you need no extra if to you use it, you could just write:
```
            precisetime = PreciseTime()
            try:
                xldriver.xlGetSyncTime(self.port_handle, offset)
            except VectorError:
                xldriver.xlGetChannelTime(self.port_handle, self.mask, offset)
            self._time_offset = precisetime.now() - offset.value * 1e-9
```

rliebscher: BTW: the constructor could have also a param required_resolution instead of hard coding 1e-5

danielhrisca:
What should I do next to get the ball rolling again for this PR?

hardbyte:
I'll merge it now, mergify usually takes over once there are 2 approvals and the CI has passed - not sure why it didn't this time.

PR #937, opened 2020-11-15
Update mergify config

hardbyte:


felixdivo:
@hardbyte This branch can be deleted, right?

Issue #46, opened 2016-01-07
J1939 logger script interface + channel args


hardbyte:
Originally reported by: **Brian Thorne (Bitbucket: [hardbyte](https://bitbucket.org/hardbyte), GitHub: [hardbyte](https://github.com/hardbyte))**

---

Add an additional argument (-i) to `J1939_logger.py`  to specify the type of interface 

Need to be able to specify the channel as `can0` and interface as `socketcan`. 

This functionality is implemented in the standard can_logger.py script so could basically copy the interface portions over.

---
- Bitbucket: https://bitbucket.org/hardbyte/python-can/issue/46

PR #940, opened 2020-11-19
Version bump CI tools

felixdivo:
The automated static analysis tools are updated to their most recent version.

EDIT: I basically updated the tools in `requirements-lint.txt` and some images in Travis CI. Then I and re-ran the formatter, which  causes much of the diff. I also removed the formatting job from Travis as it was already done by Github actions. I removed the normal testing from Travis, as it is much easier to specify in Github actions (and we now also test more platforms, like properly test on macOS!). We now also test on newer versions of Python as well (currently up to `3.10.0-alpha.4`) and on Pypy on all platforms (also new).


hardbyte (approved):
:medal_military:

felixdivo:
How useful is it to have both Github actions as well as AppVeyor an Travis CI? Wouldn't it be nice to at least remove the tests from Travis CI as they seem to be very slow and a duplicate. Also, do we need AppVeyor at all?

hardbyte:
> How useful is it to have both Github actions as well as AppVeyor an Travis CI? Wouldn't it be nice to at least remove the tests from Travis CI as they seem to be very slow and a duplicate. Also, do we need AppVeyor at all?
> 
> EDIT: Proposal:
> 
>     * retain all Github workflow tests, which test with CPython on Linux, macOS and Windows
> 
>     * Move CPython 3.10-dev & nightly from Travis to Github workflow
> 
>     * remove Travis CI testing except: pypy3 and socketcan tests
> 
>     * retain the other Travis jobs like the linter and publishing (for now at least, we can consider moving them in another PR)
> 
>     * remove AppVeyor testing as it is entirely covered by Github workflows (right?)

Travis-CI _might_ be a bit faster now I just [migrated](https://docs.travis-ci.com/user/migrate/open-source-repository-migration#frequently-asked-questions) to the travis-ci.com version.
But yes I think just relying on github workflow except where we need additional functionality is the way forward. 

Is there more you want to do in this PR?

felixdivo:
I basically updated the tools in `requirements-lint.txt` and now also in `.travis.yml`. I'd wait for the discussion in #951 before doing more, since the macOS build seems to fail for now but there is no need for debugging if we delete anyways.

felixdivo:
Seems like this uncovered [an actual problem](https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=1458088551#step:4:101): https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=1458088551#step:4:101. Pypy support for that lib is aparrently not available, see https://github.com/python/typed_ast/issues/111.

felixdivo:
I also opened #46

felixdivo:
The problem on Python version 3.10 with `python-lazy-object-proxy` should by fixed by `wheel` version 0.36.2.

felixdivo:
Seems like this time, we actually uncovered a bug in our repository: https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=1753568353#step:5:333

zariiii9003:
We could set python 3.10 to experimental and merge this. What do you think?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error

felixdivo:
@zariiii9003 Is this what you thought about?

zariiii9003:
Yes that's what i meant, thank you. I'm just wondering why the GitHub tests were not run on your last commit. 

Also: the example says `runs-on: ${{ matrix.os }}` while it is `runs-on: ubuntu-latest` in our build.yml. Is this correct?

zariiii9003:
black is incompatible with pypy3. So it fails on step `pip install -r requirements-lint.txt`
Should we separate tests and linters into different jobs? Or is there a way to skip linters only for pypy?

felixdivo:
Ok, so there seem to be more failures...

Somehow efc5f26889a7e593e6d70ed7e486e65e6af1e467 did not disable the `TestMessageSync` test, which is causing problems on GitHub Actions on macOS and Windows. Additionally, there seems to be an issue with the scripts not being available on WIndows.

zariiii9003:
On windows the mypy_extensions are not found during the script-test. I guess the tox environment might not be passed to the subprocess call...

felixdivo:
Thanks @zariiii9003. Somehow that variable was not set despite [this doc]() saying it would be.

The last error seems to be [`File "D:\\a\\python-can\\python-can\\can\\message.py", line 11, in <module>\r\n    from . import typechecking\r\n  File "D:\\a\\python-can\\python-can\\can\\typechecking.py", line 9, in <module>\r\n    import mypy_extensions\r\nModuleNotFoundError: No module named \'mypy_extensions\'`](https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=2331114631#step:5:361). But it also says [`mypy-extensions==0.4.3`](https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=2331114631#step:5:10) is installed right above. I don't really know how tox works though...

felixdivo:
Dependencies now install nicely on PyPy on Windows.

felixdivo:
It's unclear to my, why `test/back2back_test.py::TestThreadSafeBus::test_concurrent_writes` is [failing on macOS, PyPy3](https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=2335703643#step:5:1918). Threading should work in PyPy just like in CPython, where the tests succeeds.

zariiii9003:
Here's a new one (for me): https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=2336410553#step:5:332

And the pypy3 threading error also occurs on windows.

felixdivo:
> Here's a new one (for me): https://github.com/hardbyte/python-can/pull/940/checks?check_run_id=2336410553#step:5:332

Should be fixed. The runners have such an unreliable timing that any kind of deadline/timing assert seems to fail eventually.

felixdivo:
Asyncio error in python 3.10 is tracked in #1005

felixdivo:
@zariiii9003 Everything seems to be done now, right?

zariiii9003:
Depends, do we keep Appveyor?

felixdivo:
It would also touch a few (even more) lines, I'd just do it in a separate PR right after this one is merged. It's already somewhat too large I think.

can/ctypesutil.py
```diff
@@ -42,7 +42,7 @@ def map_symbol(self, func_name, restype=None, argtypes=(), errcheck=None):
                 "Could not map function '{}' from library {}".format(
                     func_name, self._name
                 )
-            )
+            ) from None
```
zariiii9003, line 45: This should be:

```python3
        except AttributeError as attribute_error:
            raise ImportError(
                "Could not map function '{}' from library {}".format(
                    func_name, self._name
                )
            ) from attribute_error
```
Otherwise Traceback is lost, right?

can/interface.py
```diff
@@ -39,7 +41,7 @@ def _get_class_for_interface(interface):
             "Cannot import module {} for CAN interface '{}': {}".format(
                 module_name, interface, e
             )
-        )
+        ) from None
```
zariiii9003, line 44: `raise from e`

can/interface.py
```diff
@@ -29,7 +29,9 @@ def _get_class_for_interface(interface):
     try:
         module_name, class_name = BACKENDS[interface]
     except KeyError:
-        raise NotImplementedError("CAN interface '{}' not supported".format(interface))
+        raise NotImplementedError(
+            "CAN interface '{}' not supported".format(interface)
+        ) from None
```
zariiii9003, line 34: raise from KeyError reference

can/interface.py
```diff
@@ -49,7 +51,7 @@ def _get_class_for_interface(interface):
             "Cannot import class {} from module {} for CAN interface '{}': {}".format(
                 class_name, module_name, interface, e
             )
-        )
+        ) from None
```
zariiii9003, line 54: `raise from e`

can/interfaces/pcan/pcan.py
```diff
@@ -620,16 +620,19 @@ def _detect_available_configs():
...
+                    continue
+                has_fd = False
+                error, value = library_handle.GetValue(i["id"], PCAN_CHANNEL_FEATURES)
+                if error == PCAN_ERROR_OK:
+                    has_fd = bool(value & FEATURE_FD_CAPABLE)
+                channels.append(
+                    {"interface": "pcan", "channel": i["name"], "supports_fd": has_fd}
+                )
+            except AttributeError:  # Ignore if this fails for some interfaces
+                pass
```
zariiii9003, line 635: Do we know the reason? Is some warning necessary here?

test/config.py
```diff
@@ -16,20 +16,27 @@ def env(name: str) -> bool:
...
+    or IS_GITHUB_ACTIONS
+    or env("CI")
+    or env("CONTINUOUS_INTEGRATION")
+)
 
-if IS_APPVEYOR and IS_TRAVIS:
+if IS_APPVEYOR and IS_TRAVIS and IS_GITHUB_ACTIONS:
     raise EnvironmentError(
-        "IS_APPVEYOR and IS_TRAVIS cannot be both True at the same time"
+        "only one of IS_APPVEYOR and IS_TRAVIS and GITHUB_ACTIONS max be True at the same time"
```
zariiii9003, line 39: typo in "may"

can/ctypesutil.py
felixdivo: Well, this was intentional, since the message is already quite verbose. If you want to commit your proposal, go ahead. ;)

can/interface.py
felixdivo: Same as above. This was changed to please pylint (I think it was that one).

test/config.py
felixdivo:
```suggestion
        "only one of IS_APPVEYOR and IS_TRAVIS and GITHUB_ACTIONS may be True at the same time"
```

can/ctypesutil.py
zariiii9003: If it's intentional then i trust you there :)

can/interfaces/pcan/pcan.py
felixdivo: No, I do not really know what is failing here.

zariiii9003 (approved):

can/interfaces/pcan/pcan.py
felixdivo: Let's just open an issue on this.

felixdivo: Tracked in #1007

felixdivo:
@hardbyte Do you wish to do/have someone do another review or is it OK to be merged as is?

hardbyte:
Good work hunting down all the loose ends! Happy to merge

felixdivo:
@zariiii9003 Nice work, thanks for joining. I'll assemble a much simpler PR that will remove AppVeyor in a minute.

PR #944, opened 2020-11-21
removes use of the bloated pywin32 library

kdschlosser:
Most of the code I hard coded in what pywin32 does. There is a portion of the code that used wmi to enumerate usb devices and would return a piece of the instance id. I elected to use an already made library for this (pysetupdi) instead of coding it myself, which I can do if needed. This library is not listed on the python package index and is installed from the authors GitHub repository using setuptools. Collecting the instance ids using pysetupdi is going to be leaps and bounds faster then using COM objects to access the Windows WMI database. It is also going to be less prone to problems.

I have not tested the code because I do not have the devices that used pywin32. If there are people that do have these devices and are willing to report back if the code changes work or not that would be really helpful.

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -18,14 +18,12 @@ import abc
 import logging
 import threading
 import time
-
-# try to import win32event for event-based cyclic send task(needs pywin32 package)
-try:
-    import win32event
-
-    HAS_EVENTS = True
-except ImportError:
-    HAS_EVENTS = False
+from .events import (
+    CreateWaitableTimer,
+    CancelWaitableTimer,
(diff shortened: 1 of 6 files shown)


zariiii9003:
Can you find an alternative to that dependency link? That might cause issues when installing in a corporate environment.

kdschlosser:
yessum 

I am working on hard coding it into the library. I put some thought into it after I did the PR and I didn't care for the dependency link either.

pywin32 is big time overkill for what is needing to be done with this library.

kdschlosser:
well those are odd errors. why is ctypes.wintypes available on a Posix system?

PR #945, opened 2020-11-21
removes setuptools from install requires

kdschlosser:
It would be rather redundant to have setuptools in install requires when setuptools is being imported at the top of the setup.py file.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,8 +78,6 @@ setup(
     # see https://www.python.org/dev/peps/pep-0345/#version-specifiers
     python_requires=">=3.6",
     install_requires=[
-        # Setuptools provides pkg_resources which python-can makes use of.
-        "setuptools",
         "wrapt~=1.10",
         'windows-curses;platform_system=="Windows"',
         "filelock",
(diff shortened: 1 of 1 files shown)


felixdivo:
Hm, but maybe it is parsed by external tools? I doesn't really harm to have it in there, right?

kdschlosser:
the thing is it would never get to the point of even running the setup function if setup tools is not installed because the setup function is apart of setuptools and needs to be imported.

I am not sure if pip processes a requirements.txt file before running the setup program. If it does this would be the way that setuptools could get installed before running setup.py

this is what the requirements.txt file would have in it

```
setuptools
wrapt~=1.10
filelock
mypy_extensions<0.5.0,>=0.4.0

windows-curses; sys_platform=='win32'
pywin32; sys_platform=='win32'
```

I'll have to dink about with it and see.

hardbyte:
I'm inclined to leave it simply as documentation - because even once we move to a pyproject file we would still want to depend on `setuptools` or migrate from using `pkg_resources`.

kdschlosser:
whatever your guys want to do, It was something I noticed really has no function and causes the installation process to take one extra step when it doesn't have to.

felixdivo:
> to leave it simply as documentation

Sounds good. Comment it out and add "# it must already be installed at this point"?

hardbyte (approved):

PR #947, opened 2020-11-22
Pcan extended message

kail:
# Problem
It appears like the latest Mac PCUSB driver is using 32 bit MSG ID values instead of 64 bit. This breaks extended message IDs which show up as standard-size at the receiver.

For example:
```
message = can.Message(arbitration_id=0xF123, is_extended_id=True,
                      data=[0x11, 0x22, 0x33])
```
will actually end up with `arbitration_id=0x123`

# Solution
This change adds a (backwards compatible) flag `mac_use_standard_message` which uses the standard message type (as in not the mac-specific one) for communicating with the driver.

The arguments all use default values, and this will not break upgrades for any existing users.

This change was validated on `Mac OSX 10.14.6` with `libPCUSB v0.9`

### DIFF
--- a/can/interfaces/pcan/basic.py
+++ b/can/interfaces/pcan/basic.py
@@ -716,7 +716,7 @@ class PCANBasic:
             logger.error("Exception on PCANBasic.GetStatus")
             raise
 
-    def Read(self, Channel):
+    def Read(self, Channel, mac_use_standard_message=False):
 
         """
           Reads a CAN message from the receive queue of a PCAN Channel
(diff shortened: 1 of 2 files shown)


kail:
@hardbyte what's would be the best way to get this merged? Am I missing any steps that are part of contributing?

hardbyte:
> @hardbyte what's would be the best way to get this merged? Am I missing any steps that are part of contributing?

Sorry I didn't see your message. Ideally someone who has a pcan device would have taken a look.

I think this issue was fixed by https://github.com/hardbyte/python-can/pull/976

PR #952, opened 2020-11-26
changes absolute imports to explicit relative imports 

kdschlosser:
this is done so if a user wants to package the library into their application they will be able to without having to modify sys.path.
It is also done to prevent name conflicts with any other package that may be named can.

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -7,12 +7,12 @@ The main entry point to these classes should be through
 
 from typing import Optional, Sequence, Tuple, Union, Callable, TYPE_CHECKING
 
-from can import typechecking
+from . import typechecking
 
 if TYPE_CHECKING:
-    from can.bus import BusABC
+    from .bus import BusABC
 
-from can.message import Message
+from .message import Message
 
 import abc
 import logging
--- a/can/bus.py
+++ b/can/bus.py
@@ -4,17 +4,18 @@ Contains the ABC bus implementation and its documentation.
 
 from typing import cast, Any, Iterator, List, Optional, Sequence, Tuple, Union
 
-import can.typechecking
+from . import typechecking
 
 from abc import ABCMeta, abstractmethod
-import can
+
 import logging
 import threading
 from time import time
 from enum import Enum, auto
 
-from can.broadcastmanager import ThreadBasedCyclicSendTask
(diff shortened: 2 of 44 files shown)


hardbyte:
I'm not sure what you mean by local imports, we do absolute imports. PEP 8 explicitly recommends absolute imports.

I'm inclined towards rejecting this proposal, but you can try convince me (or other core devs) why we should accept it?

We put a lot of effort ensuring that python-can is installable as a dependency like most other Python packages, and of course there are tools for various platforms that will package Python applications along with their dependencies.

kdschlosser:
Problem 1
The way it is done now makes it so that python-can cannot have it's root folder name changed. There are lots of can utilities out there for Python, I am sure that there are other libraries that also share the same import name. So user would not be able to solve the problem of a library having the same name as another library by simply changing the name of the folder. This is because python-can is going to attempt to import  from possibly another library named can if the folder name gets changed.

Problem 2
OK so if I want to use this package in a program I am writing, I want the project to be "self contained" and not have to install any dependencies. something along the lines of running it off a flash drive. If I attempt to import python-can it will fail. This is because because python-can relies on being placed in the root of a path that is included in sys.paths. I know this is a simple thing to fix by adding the folder that contains python-can to sys.path. The issue is an IDE is not going to work the way it should and is not going to be able to provide code completion and intellisense.

kdschlosser:
also PEP8 states that explicit imports are OK to use. it's the implicit imports that are the problem

so with respect to python can you cannot do this inside any file.

```python
import bit_timing
```

You can do this
```python
from . import bit_timing
```

and this as well
```python
from can import bit_timing
```

The same argument I made above has also come up regarding using absolute imports in a library. If the folder name for the library changes the library is going to fail to load. And that is the reason why explicit relative imports are OK

zariiii9003:
I don't see the benefit. Also: Relative imports from the same level look okay. But i don't like things like `from ... import Message, CanError, BusABC`

felixdivo:
I also don't see the benefit. Changing the root module name would also probably mess with more than the imports.

hardbyte:
Thanks for the opinions @zariiii9003 and @felixdivo 

Closing as this well intentioned contribution doesn't clearly improve the project.

PR #960, opened 2020-12-18
Add support for changing the loopback flag

jjguti:
Small PR to allow the local loopback to be disabled/enabled on a socketcan interface.

### DIFF
--- a/test/test_socketcan_loopback.py
+++ b/test/test_socketcan_loopback.py
@@ -45,6 +45,8 @@ class LocalLoopbackSocketCan(unittest.TestCase):
             loopback_send_bus.send(msg)
             recv_msg = self._recv_bus.recv(self.TIMEOUT)
             self.assertIsNotNone(recv_msg)
+            recv_msg_lb = loopback_send_bus.recv(self.TIMEOUT)
+            self.assertIsNone(recv_msg_lb)
         finally:
             loopback_send_bus.shutdown()
 
@@ -62,6 +64,8 @@ class LocalLoopbackSocketCan(unittest.TestCase):
             noloopback_send_bus.send(msg)
             recv_msg = self._recv_bus.recv(self.TIMEOUT)
             self.assertIsNone(recv_msg)
+            recv_msg_nlb = noloopback_send_bus.recv(self.TIMEOUT)
+            self.assertIsNone(recv_msg_nlb)
         finally:
             noloopback_send_bus.shutdown()
 
--- a/test/test_socketcan_loopback.py
+++ b/test/test_socketcan_loopback.py
@@ -1,0 +1,65 @@
+"""
+This module tests sending messages on socketcan with and without local_loopback flag
+
+for a good explanation of why this might be needed:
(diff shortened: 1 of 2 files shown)


felixdivo:
@jjguti This looks like a neat addition. But I think we should add a test to make sure it correctly loops back and only does so if the flag is set. Could You add that? Then, this can be merged.

jjguti:
@felixdivo Hi, I have now added a test for the local_loopback flag, please tell me if it's good enough for a merge or if something else should be added.

felixdivo:
Thanks for your quick reply! And sorry for the long pause, there is recently quite a backlog of issues/PRs.

I think that the test is already nearly complete, but we should add a few lines, see my comments below.

test/test_socketcan_loopback.py
```diff
@@ -0,0 +1,70 @@
...
+        on the local device"""
+        loopback_send_bus = can.Bus(
+            interface=self.INTERFACE_1,
+            channel=self.CHANNEL_1,
+            bitrate=self.BITRATE,
+            local_loopback=True,
+        )
+        try:
+            msg = can.Message(arbitration_id=0x123, is_extended_id=False)
+            loopback_send_bus.send(msg)
```
felixdivo, line 45: Test that `loopback_send_bus.recv()` returns a message.

test/test_socketcan_loopback.py
```diff
@@ -0,0 +1,70 @@
...
+        noloopback_send_bus = can.Bus(
+            interface=self.INTERFACE_1,
+            channel=self.CHANNEL_1,
+            bitrate=self.BITRATE,
+            local_loopback=False,
+        )
+        try:
+            msg = can.Message(arbitration_id=0x123, is_extended_id=False)
+            noloopback_send_bus.send(msg)
+            recv_msg = self._recv_bus.recv(self.TIMEOUT)
```
felixdivo, line 63: Test that `Test that `loopback_send_bus.recv()` returns `None`.

test/test_socketcan_loopback.py
jjguti: loopback_send_bus.recv() does NOT return a message. The local_loopback flag (as far as I understand it), means the *other* local sockets open on the same machine will receive the message, but not your own socket. I can instead add an assert that it does not recv() in any case.

jjguti: A bit more info can be read here:

https://www.kernel.org/doc/html/latest/networking/can.html#socketcan-local-loopback1

felixdivo: Whoops. OK, that also sounds useful (even more useful). Then yes, do as you said. Maybe briefly document it in `SocketcanBus`?

jjguti: Added. This was very confusing for me also when I started looking at it :)

hardbyte (approved):
👍🏼

felixdivo:
Nice @jjguti!

felixdivo (approved):

Issue #971, opened 2021-01-20
error using multiple Vector devices


danielhrisca:
The setup contains 2 Vector devices and a CAN cable that connects the ECU and the 2 boxes (using the first piggy on each device):
* Vector Vn1630A
* Vector VH6501

``CANoe CAN1`` is assigned to the first piggy of the VH6501 device. 

If I assign ``MyAPP CAN1`` to the same piggy then I can record the traffic.
If I assign ``MyAPP CAN1`` to the first piggy of VN1630A then I get the following error:

```
Details: Traceback (most recent call last):
  File "MyAPP\xil\devices\can_device.py", line 276, in prepare
  File "lib\site-packages\can\util.py", line 303, in wrapper
  File "lib\site-packages\can\interfaces\vector\canlib.py", line 183, in __init__
  File "lib\site-packages\can\interfaces\vector\canlib.py", line 686, in get_application_config
  File "lib\site-packages\can\interfaces\vector\xldriver.py", line 35, in check_status
can.interfaces.vector.exceptions.VectorError: xlGetApplConfig failed (XL_ERROR)
```

Any idea about what is going on here?

zariiii9003:
Hello Daniel,

can you find the channel in `vector.canlib.get_channel_configs()`?

The function call 
```python
xldriver.xlGetApplConfig(
    app_name.encode(),
    app_channel,
    hw_type,
    hw_index,
    hw_channel,
    xldefine.XL_BusTypes.XL_BUS_TYPE_CAN,
)
```
is pretty straightforward, i have no clue how it could cause an error

PR #977, opened 2021-02-09
Fix vector get application config

danielhrisca:


felixdivo:
Hey @danielhrisca what problem does this PR try to solve? I don't quite get it. Was the value just not wrapped into a `c_uint`? Was it tested on real hardware? If yes, IMHO this can be merged.

zariiii9003:
@felixdivo i assume this solves #971

danielhrisca:
I think it was related to using the udsoncan package. I will check tomorrow

zariiii9003:
The [ctypes documentation](https://docs.python.org/3.10/library/ctypes.html#calling-functions) says 

> Python integers are passed as the platforms default C int type, their value is masked to fit into the C type.

So while using `app_channel = ctypes.c_uint(app_channel)` is technically correct it should not make a difference unless app_channel is greater than 2147483647 (2^7 - 1). 
@danielhrisca did you see any difference with your version?

zariiii9003 (approved):

felixdivo:
@zariiii9003 Shall we merge this an close #971?

zariiii9003:
I would say we can merge this, but not close the issue you mentioned. I'm not sure if this fixes anything, but i'm very confident that it does not break anything.

Let's wait and see if someone else has similar problems once we release version 4.

PR #980, opened 2021-02-18
Add Neousys WDT_DIO CAN interface

illuusio:
Bindings to Neousys DLL/SO so  one can use CAN interface


can/interfaces/neousys_wdt.py
```diff
@@ -0,0 +1,269 @@
+""" Neousys WDT_DIO CAN driver """
+
+#
+# This kind of interface can be found for example on Neousys POC-551VTC
+# One needs to have correct drivers and DLL (Share object for Linux) from Neousys
+#
+# https://www.neousys-tech.com/en/support-service/resources/category/299-poc-551vtc-driver
+#
+# Beware this is only tested on Linux kernel higher thant 5.3. This should be drop in
```
hardbyte, line 9:
```suggestion
# Beware this is only tested on Linux kernel higher than v5.3. This should be drop in
```

can/interfaces/neousys_wdt.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _recv_internal(self, timeout):
+        msg = None
+
+        # If there is message waiting in array
+        # pass it as new message
+        if len(self.recv_msg_array) > 0:
+            self.lock.acquire()
+            msg = self.recv_msg_array.pop(0)
+            self.lock.release()
```
hardbyte, line 219:
```suggestion
            with self.lock:
                msg = self.recv_msg_array.pop(0)
```

can/interfaces/neousys_wdt.py
```diff
@@ -0,0 +1,269 @@
...
+            self.channel = channel
+
+            self.device = device
+
+            self.channel_info = "Neousys WDT_DIO Can: device {}, channel {}".format(
+                self.device, self.channel
+            )
+
+            self.lock = threading.Lock()
+            self.recv_msg_array = []
```
hardbyte, line 154: Is there any reason why you are using a Lock + List rather than a [Queue](https://docs.python.org/3/library/queue.html#module-queue)?

can/interfaces/neousys_wdt.py
```diff
@@ -0,0 +1,269 @@
+""" Neousys WDT_DIO CAN driver """
```
hardbyte, line 1: Let's rename the interface as neousys and put all the code into a package at `can/interfaces/neousys`.

hardbyte (commented):
Look pretty good. A few minor comments inline.

Please add an interface documentation page before we merge

can/interfaces/neousys_wdt.py
illuusio: So I make subdir under interfaces or just rename file and class..

hardbyte: Sorry for the delay replying to you. Yes please make a subdir under `can/interfaces` and feel free to create as many modules within that.

illuusio:
@hardbyte I reorgazied to Queue and it's in own subdir. I have to test this more to be sure it works as expected. Need also make sure that there is no need to catch Queue.Empty and Full (they should be tested but documentation is little bit shy about this)

hardbyte:
Thanks @illuusio - please let me know when you'd like another review.

illuusio:
@hardbyte now it should be up to another review as it has unit test also (no marvelous but should catch most of the stuff). Strange I reformatted test_neosys.py with black but it's still failing I have to correct that.

hardbyte (approved):

can/interfaces/neousys/neousys.py
```diff
@@ -0,0 +1,272 @@
...
+# Beware this is only tested on Linux kernel higher than v5.3. This should be drop in
+# with Windows but you have to replace with correct named DLL
+#
+
+# pylint: disable=R0903
+# pylint: disable=R0902
+# pylint: disable=C0413
+# pylint: disable=E0202
+# pylint: disable=W0611
+# pylint: disable=R1725
```
felixdivo, line 18: This is probably easier to read:

```suggestion
# pylint: disable=too-few-public-methods
# pylint: disable=too-many-instance-attributes
# pylint: disable=wrong-import-position
# pylint: disable=method-hidden
# pylint: disable=unused-import
```

The last one, `R1725`/`super-with-arguments` is not required (see below).

can/interfaces/neousys/neousys.py
```diff
@@ -0,0 +1,272 @@
...
+class NeousysBus(BusABC):
+    """ Neousys CAN bus Class"""
+
+    def __init__(self, channel, device=0, bitrate=500000, **kwargs):
+        """
+        :param channel: channel number
+        :param device: device number
+        :param bitrate: bit rate. Renamed to bitrate in next release.
+        """
+        super(NeousysBus, self).__init__(channel, **kwargs)
```
felixdivo, line 145: In Python 3, just use `super()`:

```suggestion
        super().__init__(channel, **kwargs)
```

can/interfaces/neousys/neousys.py
felixdivo: Why are the last two required? That seems odd ...

felixdivo (approved):

PR #985, opened 2021-03-09
Disable command PCAN_ALLOW_ERROR_FRAMES on MAC OSx

sdorre:
As this command is not yet supported by PCANUsb lib from MACCAN
See: https://www.uv-software.de/files/downloads/MacCAN/PCANUSB/Library/OS_X_Library_for_PCANUSB_v0.10.readme

Signed-off-by: Stephane Dorre <stephane.dorre@gmail.com>

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -281,13 +281,19 @@ class PcanBus(BusABC):
         if result != PCAN_ERROR_OK:
             raise PcanError(self._get_formatted_error(result))
 
-        if platform.system() != "Darwin":
-            result = self.m_objPCANBasic.SetValue(
-                self.m_PcanHandle, PCAN_ALLOW_ERROR_FRAMES, PCAN_PARAMETER_ON
-            )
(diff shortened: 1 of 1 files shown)


felixdivo:
What about trying on every platform and only *allowing it to fail* on macOS (with a small `log.debug("ignoring error because ...")`? that way, it will automatically work once the newer versions of the driver support it.

sdorre:
> What about trying on every platform and only _allowing it to fail_ on macOS (with a small `log.debug("ignoring error because ...")`? that way, it will automatically work once the newer versions of the driver support it.

I can try. I would still keep the TODO to remove it once it is supported.

felixdivo:
Sounds good. There's also a merge conflict that you might want to resolve first (it's easy from the Web GUI here).

felixdivo:
I don't know what you have done in your force-push but the conflict remains.

sdorre:
> I don't know what you have done in your force-push but the conflict remains.

I am sorry i forget to fetch upstream. Here we go then

felixdivo (approved):

PR #987, opened 2021-03-12
add set/get pcan device number

gordon-epc:
Add get_pcan_device_number and set_pcan_device_number methods, which gives the ability to get/set PCAN device number/ID like the PEAK-System PCAN-View application.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -297,6 +297,41 @@ class PcanBus(BusABC):
         status = self.m_objPCANBasic.Reset(self.m_PcanHandle)
         return status == PCAN_ERROR_OK
 
+    def get_device_number(self):
+        """
+        Return the PCAN device number.
+
+        :rtype: int
+        :return: PCAN device number
+        """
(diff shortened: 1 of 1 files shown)


gordon-epc:
I tested the new methods by connecting a PEAK-System PCAN-USB interface (IPEH-002021) to a Windows PC, executing the PCAN-View application, calling the set method and verifying that PCAN-View updated accordingly, and calling get and checking that it returned the expected value. I also tested the set method with a variety of good and bad parameter values and the method behaved appropriately.

hardbyte:
Thanks for the contribution, could you please take a look at the failing tests - https://travis-ci.com/github/hardbyte/python-can/jobs/490305673

KloolK:
Confirmed to also work with PCAN-USB FD (IPEH-004022) 👍

gordon-epc:
It looks like the failing test also failed the same way on the [current head of develop](https://github.com/hardbyte/python-can/runs/2076218313).

hardbyte:
Ahh that's not great. Thanks for following up.

PR #989, opened 2021-03-12
USB2CAN Windows error 8 fix

wiboticalex:
This pull request fixes occasional "error 8" status errors that I was experiencing when using a USB2CAN by ignoring them. This seemed to be the correct thing to do since RCV_EMPTY is already ignored, and the CANAL API pdf states that this error means "Nothing available to read. FIFO is empty".

I tested with a 64bit usb2can.dll version 1.0.9.0 from the "WinUSB CANAL DLL v1.0.0" link on the 8devices page. MD5 hash: e8757f04e3333b7f3dedf22b7f3d4d24. Hardware version 2.0, firmware version 2.3.

I also added the error codes specified in the CANAL API pdf to an enum so that more informative error messages can be printed for other error codes.

### DIFF
--- a/can/interfaces/usb2can/usb2canInterface.py
+++ b/can/interfaces/usb2can/usb2canInterface.py
@@ -141,7 +141,7 @@ class Usb2canBus(BusABC):
 
         if status == CanalError.SUCCESS:
             rx = message_convert_rx(messagerx)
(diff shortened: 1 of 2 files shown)


felixdivo (approved):
Looks good. :)

hardbyte (approved):

PR #991, opened 2021-03-22
Fix find_library call in PCAN layer, where PCAN library is loaded on Mac

nworbttam:
This PR fixes an issue in the PCAN Interface layer, where the `find_library` function wasn't being referenced properly.


sdorre:
It is a duplicate to #984

hardbyte:
Thanks, merged the fix via #984

Issue #993, opened 2021-03-23
pcan: neutral/german error messages cause logger exceptions


XelaRellum:
Any return value other than `PCAN_ERROR_OK` for _PCAN_ SDK functions will be translated to the corresponding error text.

The parameter to the SDK function `GetErrorText` specifiying the language is `0` (neutral language). This seems to mean the language of the system, because on my German Windows 10 system this results in a Germany message.

There seems to be a bug in the decoding of the error message to UTF-8.

It seems to be easier for me to use language code `0x9` (English) instead of fixing the decoding bug since the exceptions use English texts as well without using to localize them.

felixdivo:
Fixed by #994.

PR #994, opened 2021-03-23
pcan: force english error messages (#993)

XelaRellum:
On non-english Windows systems the "neutral" language causes
an UTF-8 encoding error while trying to log an error/warning.

Force using English error message which will not cause these.


hardbyte (approved):
Seems like a good compromise. 

If you do feel like looking into why you got an encoding error with the German error text please open another issue/PR.

Issue #992, opened 2021-03-23
send periodic a list of messages without wrapping


zorce:
Is it possible to send a list of messages periodically without wrapping when reaching the end of the list?
using the duration based on len(send_list) * cycle_time will overshoot and retransmit.

Creating a list of messages to send cyclic (including 20 or so muxed frames)

```
send_list = list()
for bmsg in bmsg_list:
    send_list.append(can.Message(arbitration_id=txid, is_extended_id=False, data=bmsg._data))

duration = len(send_list) * inhibit
task = c.bus.send_periodic(send_list, inhibit, duration=duration)
```

hartkopp:
So you would like so send this sequence as a one-shot event?
The **count** element in struct bcm_msg_head can limit the number of transmissions:
https://www.kernel.org/doc/html/latest/networking/can.html#broadcast-manager-protocol-sockets-sock-dgram
So if it is set to "20 or so" you should get a one-shot sequence transmission ...

zorce:
Yes that is correct, would like to send it as a one-shot event. Creating a list of muxed messages for "CAN-in-Automation" that needs to be sent on demand. 
If that count would be accessable as a parameter to send_periodic that would solve it for me just perfectly.

hartkopp:
https://python-can.readthedocs.io/en/master/bcm.html#can.broadcastmanager.MultiRateCyclicSendTaskABC is potentially what you need here. With subsequent_period being set to zero ...
But from the documentation I was not able to see whether you can provide more than one CAN frame via this API (for nframes being 20 or so)

zorce:
Have a fix for this, but have no idea how to push it :P
it is based on tag 4.0.0-dev.0 as I also need the send periodic with a list of multiple messages.

PR #995, opened 2021-03-24
Add end count to send periodic in BCM

zorce:
Add possibility to send periodic messages with a stop count as a criteria to facilitate one of sequences.
Discussed in issue https://github.com/hardbyte/python-can/issues/992

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -104,6 +104,7 @@ class LimitedDurationCyclicSendTaskABC(CyclicSendTaskABC):
         messages: Union[Sequence[Message], Message],
         period: float,
         duration: Optional[float],
+        count: Optional[int],
     ):
         """Message send task with a defined duration and period.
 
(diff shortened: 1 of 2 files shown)


can/broadcastmanager.py
```diff
@@ -104,6 +104,7 @@ def __init__(
         messages: Union[Sequence[Message], Message],
         period: float,
         duration: Optional[float],
+        count: Optional[int],
```
hardbyte, line 107:
```suggestion
        count: Optional[int] = None,
```

can/broadcastmanager.py
```diff
@@ -222,12 +228,13 @@ def __init__(
...
                          on desired behaviour of `ThreadBasedCyclicSendTask`.
         """
-        super().__init__(messages, period, duration)
+        super().__init__(messages, period, duration, count)
         self.bus = bus
         self.send_lock = lock
         self.stopped = True
         self.thread = None
         self.end_time = time.perf_counter() + duration if duration else None
+        self.end_count = count if count else None
```
hardbyte, line 237:
```suggestion
        self.end_count = count
```

hardbyte (commented):
It would be good to connect the socketcan CyclicSendTask too if you're up for it?

A new test in `simplecyclic_test.py`  would be appreciated too

zorce:
Sure will modify your suggestions and try to add a test aswell.
And also see if I could add it for socketcan interface.

felixdivo:
@zorce You can also just click on "Commit Suggestion" right from the GUI to apply the changes by hardbyte.

felixdivo:
@zorce Some conflicts have popped up by now.

PR #1012, opened 2021-04-15
Add more typying

felixdivo:
Attempts to add type hints for all previously untyped parts of the files in `can/` (but not in subdirectories). To this end, `can.Bus.send_periodic` had to be changed a bit and `can.ctypesutil` partly rewritten. However, the result should be simpler to comprehend.

All changes are internal ones in that no code outside of the touched ones should need adaption.

Since only IXXAT currently uses `ctypesutil`, that interface should perhaps be tested with these changes. However, I do not have the appropriate hardware at hand.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -172,7 +172,7 @@ class BusABC(metaclass=ABCMeta):
 
     def send_periodic(
         self,
-        msgs: Union[Message, Union[List[Message], Tuple[Message, ...]]],
+        msgs: Union[Message, Sequence[Message]],
         period: float,
         duration: Optional[float] = None,
         store_task: bool = True,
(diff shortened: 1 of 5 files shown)


can/bus.py
```diff
@@ -172,7 +172,7 @@ def send(self, msg: Message, timeout: Optional[float] = None):
 
     def send_periodic(
         self,
-        msgs: Union[Sequence[Message], Message],
+        msgs: Union[Message, Union[List[Message], Tuple[Message, ...]]],
```
hardbyte, line 175: Any reason for this change? Just curious how it was incorrect, or why this is better?

hardbyte (approved):

can/bus.py
felixdivo: I changed it since that is more in line with what is actually checked and what the error tells the user. But now that I look at it, we should probably do `isinstance(msgs, S)` instead of `isinstance(msgs, (list, tuple))` and keep the old typing.

felixdivo:
I'm somewhat confused by *mergify*. Shall I just merge it manually?

felixdivo:
@Mergifyio refresh

hardbyte:
> I'm somewhat confused by _mergify_. Shall I just merge it manually?

Yeah it appears to be taking the mergify config from this branch rather than develop.

PR #1016, opened 2021-04-18
Fix a typo in serial docs

felixdivo:
I'm not sure whether this is only a typo, but I highly suspect so. Am I right? If not, the interface code would have to be adjusted.

The current docs on this are [here](https://python-can.readthedocs.io/en/develop/interfaces/serial.html#internals).


zariiii9003 (approved):
I don't know anything about the serial bus, but i know there's no S in hex.

PR #1020, opened 2021-04-20
Modify Ixxat driver config reading

### LINKED ISSUE #702: Ixxat driver ignores UniqueHardwareID and [r|t]xFifoSize from configuration
The Ixxat canlib ignores UniqueHardwareID, rxFifoSize and txFifoxSize from the configuration file. 
The reason is that the driver requests the configuration with CamelCase but it has to use lower case.

cranium-xx:
Fix for getting the configuration settings for ixxat.

The names of the configuration entries are not case sensitive, but the previous code tried to get them with CamelCase configuration names, which did not work.
Ixxat configuration entries are now read with lowercase names.

Closes #702

### DIFF
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -437,9 +437,9 @@ class IXXATBus(BusABC):
         log.info("Got configuration of: %s", kwargs)
         # Configuration options
         bitrate = kwargs.get("bitrate", 500000)
-        UniqueHardwareId = kwargs.get("uniquehardwareid", None)
-        rxFifoSize = kwargs.get("rxfifosize", 16)
-        txFifoSize = kwargs.get("txfifosize", 16)
+        UniqueHardwareId = kwargs.get("UniqueHardwareId", None)
+        rxFifoSize = kwargs.get("rxFifoSize", 16)
+        txFifoSize = kwargs.get("txFifoSize", 16)
         self._receive_own_messages = kwargs.get("receive_own_messages", False)
         # Usually comes as a string from the config file
         channel = int(channel)
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -437,9 +437,9 @@ class IXXATBus(BusABC):
         log.info("Got configuration of: %s", kwargs)
         # Configuration options
         bitrate = kwargs.get("bitrate", 500000)
-        UniqueHardwareId = kwargs.get("UniqueHardwareId", None)
-        rxFifoSize = kwargs.get("rxFifoSize", 16)
-        txFifoSize = kwargs.get("txFifoSize", 16)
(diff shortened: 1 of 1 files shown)


hardbyte:
Something is strange with this diff - it is essentially un-reviewable. Could you perhaps try again?

cranium-xx:
> 
> 
> Something is strange with this diff - it is essentially un-reviewable. Could you perhaps try again?

Yes, I see. I don't know, what happened there, because I only changed 3 lines. Maybe this comes through the upload via the GitHub webinterface. Now my Sourcetree works and I try to upload again.

felixdivo:
It might have been an issue with line endings. But thanks for fixing! :)

However, I think the problem to be fixed is actually none, now that I have a another look at it. Which case (`CamelCase` vs. `lowercase`) we use is just convention, right? And in the [current docs on `develop`](https://python-can.readthedocs.io/en/develop/interfaces/ixxat.html#configuration-file), it clearly states that they way it is currently read is correct. So `UniqueHardwareId` is what is written in both the docs and the code. And I personally think that the current way of spelling it is a bit easier to read, though that might be subjective.

Why are you trying to change that, or do I miss something here?

cranium-xx:
In my can.ini I used this:
UniqueHardwareId = HW490493

The line command `kwargs.get("UniqueHardwareId", None)` returned `None` in this case while `kwargs.get("uniquehardwareid", None)` correctly returns `HW490493`

So the configuration file entries are ignored when read with CamelCase names. They seem to be lowercased at some point.

felixdivo:
I can't find the place where the lowering would happen...

cranium-xx:
I just found the explanation here:
https://stackoverflow.com/questions/19359556/configparser-reads-capital-keys-and-make-them-lower-case
I tried it and set `config.optionxform = str` in `can/util.py` and it worked.  Now I get the configitem names in CamelCase.
I think that would be the preferrable fix or am I wrong?

felixdivo:
Okay, so the parameter reading currently on develop never could've worked when using `.ini` files ...

I guess being case insensitive is the most user friendly since it is highly unlikely that two settings will conflict only due to their case. Sorry for the long discussion. But that would be a major change.

Your curent PR would not work when users set their config via keyword arguments form within python code. Yes, changing `optionxform` seems like a good fix for now!

cranium-xx:
Ok, so how will we proceed? Shall I do the `optionxform` fix instead?
If yes, as a new PR or a change in this PR?

felixdivo:
I think it would be best to open a new PR and close this one. I could imagine this breaking some user code. But form my point of view this is a wanted bugfix.

The new PR should probably add to the config docs that config file options are case sensitive. The `my_parser.optionxform = lambda option: option` setting would probably be best since it does not modify the keys at all (although they are probably always strings).

Thank you for your contribution and swift communication!

PR #1025, opened 2021-04-22
Add specific exceptions

### LINKED ISSUE #356: Unify exception handling
There is currently no portable way of knowing what has gone wrong in case of an error. Perhaps there should be some more specific subclasses of `can.CanError` for some generally usable exceptions.

If this sounds like a good idea we should start a list of exceptions that could be useful.

* Transmit buffer full
* Unknown channel
* Hardware not connected
* Connection/handle is down
* Bus off
* Drivers could not be loaded

felixdivo:
This PR:
- adds specific base exceptions (they were discussed in #562 and before that in #356)
- documents these
- uses them in
  - all of the `can/*.py` modules
  - all of the `can/io/*.py` modules (it is not used there anyways, and that is correct)
  - the IXXAT interface (since this was already done in the original PR)
  - (This list excludes most of the interfaces. I propose doing them in separate PRs to keep the diffs and conflicts small. See [comment below](https://github.com/hardbyte/python-can/pull/1025#issuecomment-824973650).)

Closes #356. This PR is based on the work of @marcel-kanter done in #562.


felixdivo:
Seems like this can be merged soon. Not too much would need to be done.

felixdivo:
After this is merged, I'd open new issues/PRs for:

- adapting all the other interfaces
- reviewing / extending the tests

These would need to be closed before 4.0.0 is release. I can offer to do part of it.

can/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+
+class CanError(Exception):
+    """Base class for all CAN related exceptions."""
+
+
+class CanBackEndError(CanError):
+    """Indicates an error related to the backend (e.g. driver/OS/library).
+
+    Example scenarios:
+      - The interface does not exist
```
felixdivo, line 27: I need some feedback on this: What error should be raised if the interface does not exist (on this platform)? Also a `CanBackEndError`? That's how it is done in the proposal you are reading.

One could also consider `ImportError`, not `NotImplementedError` or a specific sub-class Exceptions that may inherit from any combination of them.

can/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+         +-- CanBackendError
+         +-- CanInitializationError
+         +-- CanOperationError
+         +-- CanTimeoutError
+
+Keep in mind that some functions and methods may raise different exceptions.
+For example, validating typical arguments and parameters might result in a
+:class:`ValueError`. This should be documented for the function at hand.
+"""
+
```
felixdivo, line 17: TODO for me: [Add return codes where applicable](https://github.com/hardbyte/python-can/pull/562#discussion_r283294876).

can/exceptions.py
hardbyte: How about a specific exception e.g., `CanInterfaceNotImplemented` inheriting from `CanError` and `NotImplementedError`?

can/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+
+class CanError(Exception):
+    """Base class for all CAN related exceptions."""
+
+
+class CanBackEndError(CanError):
```
hardbyte, line 23: Perhaps the name `CanInterfaceError` lines up better with the code?

Consider adding an optional argument with the interface name.

can/exceptions.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+
+
+class CanOperationError(CanError):
+    """Indicates an error while in operation.
+
+    Example scenarios:
+      - A call to a library function results in an unexpected return value
+      - An invalid message was received
+      - Attempted to send an invalid message
```
hardbyte, line 49: I assume that if python-can detects the error we raise a `ValueError`, but if the driver rejects it we would raise this?

can/exceptions.py
felixdivo: Sounds good!

can/exceptions.py
felixdivo: Yes. I guess it is sometimes hard to distinguish why sending a message caused an error, so I propose making this clearer in the docs.

can/exceptions.py
felixdivo: The concrete interface being used should probably always be clear from both the context in which it was raised (most projects probably only use very few interfaces) and the stack trace. Would you still say it would be useful?

felixdivo: Yes, I think that name would be a better fit, but also not perfect. Maybe, we don't really need it at all? If we have a separate error for the interface not being found, this error might become unnecessary? Because the new `CanInterfaceNotImplemented` could also encompass drivers not being present. As a user, this seems like a reasonable error: The functionality the user seeks to use is not implemented on the current system. Also, I think that the interface not being usable since it never exists and it not being available due to missing drivers is something that should be clear from the message. Having separate classes might make stuff more complicated than it needs to be.

hardbyte: 💯  Yes agree completely that if we can reduce the number of exceptions a user needs to be aware of that is a win.

You're right, no need to add complexity with the interface name as this (or rather the `CanInterfaceNotImplemented` exception will always be coming from a particular interface (or explicitly mentioned in the exception message).

felixdivo:
@Mergifyio update

felixdivo:
@hardbyte Can this be merged then?

can/exceptions.py
```diff
@@ -0,0 +1,97 @@
...
+:class:`ValueError`. This should always be documented for the function at hand.
+"""
+
+from typing import Optional
+
+
+class CanError(Exception):
+    """Base class for all CAN related exceptions.
+
+    If specified, the error code is automatically prepended to the message:
```
hardbyte, line 24:
```suggestion
    If specified, the error code is automatically appended to the message:

```

hardbyte (approved):

can/exceptions.py
felixdivo: Added it to #1045.

PR #1029, opened 2021-04-23
Fix size/layout of struct needed for time stamping

### LINKED ISSUE #1021: Socketcan timestamps are missing sub-second precision
Hi 

Im running python-can 3.3.4 with python 3.8.5. When I run the code below to listen to messages on can1...

```
bus = can.Bus(channel='can1', bustype='socketcan', bitrate='500000')
reader = can.BufferedReader()
notifier = can.Notifier(bus, [reader])
msg = reader.get_message(timeout=100)
```

... the message revived looks like this:

msg

There is no "is_rx" property as stated in the documentation, how can this be achieved? 

BR Stefan

felixdivo:
This caused the sub-second resolution to be cut off.

Fixes #1021. Also adds a regression test to catch this in the future and with other interfaces too.


felixdivo:
@Mergifyio update

felixdivo:
This is done.

tysonite (approved):
Those changes look quite good. It would be really nice to get them merged.

felixdivo:
Indeed, this is an important fix. @hardbyte Could you have quick look? It isn't a large diff.

hardbyte (approved):

PR #1030, opened 2021-04-23
interface for changing bitrate

### LINKED ISSUE #549: How to change SocketCan bitrate at runtime?
Dear,

I would like to implement an auto-baud feature in my application. The idea is to probe the bus for incoming messages until we find the right bitrate from a list.

Is it possible to do that using socketcan interface? If so, how could I do that? Do I have to create new bus instances everytime I change the bitrate or can I reuse the same instance?

Thank you!

cperkulator:
Quick implementation of changing the bitrate at runtime. I just want to get a feel if this is the interface we would like or if there is something better.

I just implemented the Vector interface for now but I can add socketcan eventually (would close #549 ).

This might be a start for implementing autobaud as well.

TODO 

- need to add error handling

### DIFF
--- a/can/interfaces/vector/canlib.py
+++ b/can/interfaces/vector/canlib.py
@@ -515,6 +515,12 @@ class VectorBus(BusABC):
     def send(self, msg: Message, timeout: typing.Optional[float] = None):
         self._send_sequence([msg])
 
+    def change_bitrate(self, bitrate: int) -> None:
+        xldriver.xlCanSetChannelBitrate(
+            self.port_handle, self.mask, bitrate
+        )
+        LOG.info("SetChannelBitrate: baudr.=%u", bitrate)
+
     def _send_sequence(self, msgs: typing.Sequence[Message]) -> int:
         """Send messages and return number of successful transmissions."""
         if self.fd:
--- a/can/interfaces/vector/canlib.py
+++ b/can/interfaces/vector/canlib.py
@@ -516,9 +516,7 @@ class VectorBus(BusABC):
         self._send_sequence([msg])
 
     def change_bitrate(self, bitrate: int) -> None:
-        xldriver.xlCanSetChannelBitrate(
-            self.port_handle, self.mask, bitrate
-        )
+        xldriver.xlCanSetChannelBitrate(self.port_handle, self.mask, bitrate)
         LOG.info("SetChannelBitrate: baudr.=%u", bitrate)
 
     def _send_sequence(self, msgs: typing.Sequence[Message]) -> int:
--- a/can/bus.py
+++ b/can/bus.py
@@ -314,6 +314,19 @@ class BusABC(metaclass=ABCMeta):
             if msg is not None:
                 yield msg
 
+    @abstractmethod
+    def change_bitrate(self, bitrate: int) -> None:
+        """Changes the bitrate of the underlying bus.
+
+        Override this method to enable runtime bitrate changes.
+
+        :param int bitrate: the new bitrate to set.
+
+        :raises can.CanError:
+            if the message could not be sent
+        """
+        raise NotImplementedError("Device not capable of runtime bitrate change")
+
     @property
     def filters(self) -> Optional[can.typechecking.CanFilters]:
         """


felixdivo:
Wouldn't it be nice to implement a more generic `change_parameters()` method that takes any parameters (just like on bus init)? The method can then complain whether that specific setting is supported to be changed or not.

can/interfaces/vector/canlib.py
```diff
@@ -515,6 +516,15 @@ def handle_canfd_event(self, event: xlclass.XLcanRxEvent) -> None:
     def send(self, msg: Message, timeout: typing.Optional[float] = None):
         self._send_sequence([msg])
 
+    @property
+    def bitrate(self) -> int:
+        return self.bitrate
```
felixdivo, line 521: Doesn't this recuse infinitely? Usually, I see this being implemented with a private attribute `self._bitrate`.

felixdivo:
Your proposal for implementing it as a property looks really nice! However, as a change to the core class of the library this should get more feedback.

(Side note: This could also be used in other places since it modularization the configuration of the bus. Sometimes, that might make things more readable.)

can/interfaces/vector/canlib.py
cperkulator: oops yeah I'll get that fixed. Just tossed this together quick to get some feedback.

zariiii9003:
Some questions:
- How would you deal with CAN FD? 
- Regarding the Vector interface: what happens if you do not have Init Access?

felixdivo:
> * How would you deal with CAN FD?

I guess we would not have to implement it right away. Besides, wouldn't the bitrate be handled automatically?

> * Regarding the Vector interface: what happens if you do not have Init Access?

I'd propose to simply throw one of the new exceptions, like a [`CanOperationError`](https://github.com/hardbyte/python-can/blob/develop/can/exceptions.py#L79).

cperkulator:
> > * How would you deal with CAN FD?
> 
> I guess we would not have to implement it right away. Besides, wouldn't the bitrate be handled automatically?
> 
> > * Regarding the Vector interface: what happens if you do not have Init Access?
> 
> I'd propose to simply throw one of the new exceptions, like a [`CanOperationError`](https://github.com/hardbyte/python-can/blob/develop/can/exceptions.py#L79).

Check out the latest commit. I think it should handle these issues.  I still need to add unit tests but testing from home this all works fine.  I'll need to test on my work setup too next week.

can/interfaces/socketcan/constants.py
```diff
@@ -43,11 +43,6 @@
 MSK_ARBID = 0x1FFFFFFF
 MSK_FLAGS = 0xE0000000
 
-PF_CAN = 29
```
cperkulator, line 46: python  has these integrated into the socket module. so these are a little redundant but I can leave them in if supporting before 3.3 is still required.

can/interfaces/socketcan/socketcan.py
```diff
@@ -21,6 +21,29 @@
 log_tx = log.getChild("tx")
 log_rx = log.getChild("rx")
 
+can_config_lib = ctypes.util.find_library("socketcan")
```
cperkulator, line 24: Simpler to use this library rather than parsing output from `ip` but adds a dependency.  Thoughts?

can/interfaces/vector/canlib.py
```diff
@@ -233,6 +233,8 @@ def __init__(
         )
 
         if permission_mask.value == self.mask:
+            self._init_access = True
```
cperkulator, line 236: I'm not a huge fan of this flag but I would rather not check for init access everytime.

can/interfaces/socketcan/constants.py
felixdivo: The version currently developed on the `develop` branch (4.0.0-dev) only supports Python 3.6+. So it's safe to drop them.

hartkopp:
> @hartkopp and @lumagi : What do you think about the socketcan part of this PR? Is that a viable solution to implement bit timing configuration for socketcan?

The bitrate setting of 'real' CAN interfaces can only be performed as network-admin (aka root) on Linux. That's why the `ip` tool is mostly used with `sudo ip` ...

And the setting affects all users on the system (not only python-can users), as the interface has to be shut down to change the bitrates (and has to be set to "up" afterwards). Shutting down the interfaces leads to -ENETDOWN return values for all user space applications working on that interface. Additionally the python-can application needs to be run as root to manipulate the interface status and the bitrate.

So reading the interface properties via `ip` tool or the netlink socket is fine - but setting new bitrates from python-can is probably not as feasible as suggested by the patch.

hartkopp:
> It also seems like `libsocketcan` is not maintained anymore. The last commit is from 2020 and it does not support CAN FD. So it's probably not a good foundation to build upon.

@marckleinebudde : Is there any ongoing activity planned for `libsocketcan` ?

lumagi:
I agree with @hartkopp. The executing process must be root or at least have the `NET_ADMIN` capability to change the interface. Neither one of those things is something you want to do to your Python interpreter on a regular basis.

marckleinebudde:
There is no immediate activity planned for libsocketcan, but patches are welcome. However I'd rather have a proper (dbus?) interface for `systemd-networkd` to configure CAN (and networking in general).

Issue #613, opened 2019-06-03
Enable remaining Pylint rules (?)


felixdivo:
Currently, the Pylint rules disable the following checks (see [`/.pylintrc-wip`](https://github.com/hardbyte/python-can/blob/a7a9c096134b6fb0b2308c870bb379c6e5d8f4f6/.pylintrc-wip#L63)):

```
disable=invalid-name,
        missing-docstring,
        empty-docstring,
        line-too-long,
        too-many-lines,
        bad-whitespace,
        bad-continuation,
        wrong-import-order,
        ungrouped-imports,
        wrong-import-position,
        too-many-function-args,
        locally-disabled,
        too-many-instance-attributes,
        too-few-public-methods,
        too-many-public-methods,
        too-many-branches,
        too-many-arguments,
        too-many-locals,
        too-many-statements,
        too-many-nested-blocks,
        no-else-raise,
        bad-indentation,
        wildcard-import,
        fixme,
        broad-except,
        redefined-builtin,
        no-else-return,
        redefined-argument-from-local,
        abstract-class-instantiated,
        multiple-statements,
        cyclic-import,
        useless-else-on-loop,
        duplicate-code,
        inconsistent-return-statements,
        arguments-differ,
        unused-wildcard-import,
        logging-fstring-interpolation,
        logging-format-interpolation,
        unused-argument,
        abstract-method,
        attribute-defined-outside-init,
        protected-access,
        keyword-arg-before-vararg
```

Do we aim at eventually removing all of them? The default config does disable some rules as well (see [`/.pylintrc`](https://github.com/hardbyte/python-can/blob/a7a9c096134b6fb0b2308c870bb379c6e5d8f4f6/.pylintrc#L63))

#601 removed all previously disabled linter rules that did require only small changes.

felixdivo:
Maybe we could do a quick poll:

👍: Yay, remove all exclusions from the general exclusion list above!

👎: Nay, keep some exclusions in there! (Which ones?)

PR #1032, opened 2021-04-24
Simplify linting

felixdivo:
Removes the confusing `.pylint-wip` file to make the entire repo use the same config. Also improve the code here and there to remove some more previously disabled rules.

Related to #613.

This PR should be squashed!


felixdivo:
No more changes will be pushed here (except a review demands some).

hardbyte (approved):

PR #1033, opened 2021-04-24
Add more type annotations

### LINKED ISSUE #1027: Add typing to can/interface.py
This is needed to make the type checker actually check it and would help the documentation generation.

felixdivo:
Started out to fix #1027 but noticed that a bit more was missing.


can/notifier.py
```diff
@@ -125,15 +125,15 @@ def _rx_thread(self, bus: BusABC):
...
         if msg is not None:
             self._on_message_received(msg)
 
-    def _on_message_received(self, msg: Message):
+    def _on_message_received(self, msg: Message) -> None:
         for callback in self.listeners:
-            res = callback(msg)
-            if self._loop is not None and asyncio.iscoroutine(res):
+            res = cast(Optional[Awaitable[Any]], callback(msg))
+            if res is not None and self._loop is not None and asyncio.iscoroutine(res):
```
hardbyte, line 136: Is this always the case? The result of calling the `callback` (for listeners that just calls `on_message_received`) is an Awaitable?

hardbyte (approved):
Just spotted one thing that I'm not sure about

can/notifier.py
felixdivo: Nope, good catch. As the next line states, it can also be `None`.

```suggestion
            res = cast(Union[None, Optional[Awaitable[Any]]], callback(msg))
            if res is not None and self._loop is not None and asyncio.iscoroutine(res):
```

felixdivo:
@Mergifyio update

PR #1034, opened 2021-04-24
Catch time before 1970 in ASCReader

cperkulator:
`timestamp` method doesn't like time before 1970 apparently ans so an `OSError` was thrown.


tox.ini
```diff
@@ -11,7 +11,7 @@ deps =
     pyserial~=3.0
 
 commands =
-    pytest
+    pytest {posargs}
```
hardbyte, line 14: What's this?

cperkulator: Allows additional args to be passed through from tox. Needed to just run a single test.

felixdivo (approved):
Thanks!

Issue #1039, opened 2021-04-30
Notifier fails silently when using Listeners


sdorre:
It may be a mistake on my side. But I wanted to double check.

**Describe the bug**
When using the Notifier with the default `_rx_thread` processing thread, if an exception occurs, it tries to call `on_error()` from all the Listeners. 
As the classes all inherits from `class Listener(metaclass=ABCMeta):`
This class has already defined `on_error` But does not do anything.
```
    def on_error(self, exc: Exception) -> None:
        """This method is called to handle any exception in the receive thread.

        :param exc: The exception causing the thread to stop
        """
```

So in the Thread the error handling never says anything. And Error is not displayed anywhere. 
```
        except Exception as exc:  # pylint: disable=broad-except
            self.exception = exc
            if self._loop is not None:
                self._loop.call_soon_threadsafe(self._on_error, exc)
                raise
            elif not self._on_error(exc):
                raise
```

This is an example class:
```
class CustomListener(Listener):

    def on_message_received(self, msg: Message):
        raise Exception("What should we do now?")

    def stop(self):
        self._log_file.close()
```

So if that is not a bug, what should the `def on_error(self, exc: Exception):` look like?

felixdivo:
I guess that such a Listener would do the job of crashing the thread:

```python
class ErrorCrashListener(Listener):

    def on_error(self, exception: Exception):
        raise exception
```

The API is indeed a bit strange here. If you do not override it, error handling defaults to doing nothing (the method as implemented in the base `Listener`). Not even the exception is logged. But also, the tread continues to listen for messages. I changed that in #1040.

felixdivo:
@sdorre Does that answer your question or did it create more? ;)

sdorre:
No that is right. I should simply raise it by myself. Thanks for the hint!

PR #1040, opened 2021-05-02
Log suppressed error in notifier

felixdivo:
Currently, the default implementation is very silent since `can.Listener` already overrides `on_error()` and causes the exception to be suppressed.

Came up due to #1039.

### DIFF
--- a/can/notifier.py
+++ b/can/notifier.py
@@ -121,9 +121,13 @@ class Notifier:
             self.exception = exc
             if self._loop is not None:
                 self._loop.call_soon_threadsafe(self._on_error, exc)
+                # Raise anyways
                 raise
             elif not self._on_error(exc):
+                # If it was not handled, raise the exception here
                 raise
+            else:
+                logger.info("suppressed exception: %s", exc)
 
     def _on_message_available(self, bus: BusABC) -> None:
         msg = bus.recv(0)
@@ -138,6 +142,10 @@ class Notifier:
                 self._loop.create_task(res)
 
     def _on_error(self, exc: Exception) -> bool:
+        """Calls ``on_error()`` for all listeners if they implement it.
+
+        :returns: ``True`` if and only if at least one handler was called
+        """
         listeners_with_on_error = [
             listener for listener in self.listeners if hasattr(listener, "on_error")
         ]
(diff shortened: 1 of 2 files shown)


can/notifier.py
```diff
@@ -138,6 +142,10 @@ def _on_message_received(self, msg: Message) -> None:
                 self._loop.create_task(res)
 
     def _on_error(self, exc: Exception) -> bool:
+        """Calls ``on_error()`` for all listeners if they implement it.
+
+        :returns: ``True`` if and only if at least one handler was called
+        """
         listeners_with_on_error = [
             listener for listener in self.listeners if hasattr(listener, "on_error")
```
zariiii9003, line 150: Isn't this the actual root cause? We check whether `on_error` exists, but the base class `Listener` already defined it. So every listenere has that method whether it handles the exception or not.

I'd suggest to raise NotImplementedError in the base class and call the method for every listener here.

zariiii9003 (changes requested):

can/notifier.py
felixdivo: Yeah, I guess. Sounds reasonable. I'll adjust this PR real quick, then you can give some feedback.

felixdivo:
@Mergifyio rebase

felixdivo:
@sdorre The default listener has now changed, and explicitly signals that it does not handle errors at all.

can/notifier.py
zariiii9003: Looks good to me. Thank you. But i really hope CI works again soon :D

zariiii9003 (approved):

felixdivo:
@hardbyte What do you think? Is this change fine to be merged?

can/notifier.py
```diff
@@ -138,14 +143,22 @@ def _on_message_received(self, msg: Message) -> None:
...
 
     def _on_error(self, exc: Exception) -> bool:
-        listeners_with_on_error = [
-            listener for listener in self.listeners if hasattr(listener, "on_error")
-        ]
+        """Calls ``on_error()`` for all listeners if they implement it.
 
-        for listener in listeners_with_on_error:
-            listener.on_error(exc)
+        :returns: ``True`` if and only if at least one handler was called
```
hardbyte, line 148:
```suggestion
        :returns: ``True`` if at least one error handler was called.
```

I think it is easy to misread the "if and only if at least one ..." as "if and only if one ...".

hardbyte (approved):
Looks good

PR #1042, opened 2021-05-06
Add CAN FD support to canutils reader and writer

simontegelid:
This PR adds CAN FD support to the canutils reader and writer. I also sneaked in a small formatting improvement to use bytes.hex instead of the str/list/join-thing.

### DIFF
--- a/test/data/example_data.py
+++ b/test/data/example_data.py
@@ -112,8 +112,6 @@ TEST_MESSAGES_CAN_FD = sort_messages(
     [
         Message(is_fd=True, data=range(64)),
         Message(is_fd=True, data=range(8)),
-        Message(is_fd=True, bitrate_switch=True, is_remote_frame=True),
-        Message(is_fd=True, error_state_indicator=True, is_remote_frame=True),
         Message(is_fd=True, data=range(8), bitrate_switch=True),
         Message(is_fd=True, data=range(8), error_state_indicator=True),
     ]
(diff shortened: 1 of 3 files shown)


test/data/example_data.py
```diff
@@ -112,8 +112,6 @@ def sort_messages(messages):
     [
         Message(is_fd=True, data=range(64)),
         Message(is_fd=True, data=range(8)),
-        Message(is_fd=True, bitrate_switch=True, is_remote_frame=True),
```
felixdivo, line 115: Why did you remove these?

felixdivo:
Nice addition!

test/data/example_data.py
simontegelid: I explained it in the commit message: "Remove RTR tests for FD frames since they doesn't exist
and thus cannot be represented in the canutils log format.
See "CAN with Flexible Data-Rate, version 1.0, Bosch,
April 17th, 2012, pp.11": "The REMOTE TRANSMISSION REQUEST
(RTR) bit only exists in CAN format frames. [...] There are
no REMOTE FRAMES in CAN FD format." "

felixdivo: Ah, nice catch! Sorry, I did not recognize the additional lines in the message. Looks fine.

felixdivo:
Could you run `black python-can/can/io/canutils.py` once to fix the formatting complaint [here](https://github.com/hardbyte/python-can/pull/1042/checks?check_run_id=2534489978#step:5:73)? Else, this looks fine to be merged.

felixdivo (approved):

simontegelid:
I think it should be fine now!

hardbyte (approved):
Thanks!

PR #1052, opened 2021-05-17
Fix VectorError bug

zariiii9003:
Fixes #1047, fixes #1050

PR #1054, opened 2021-05-19
vector interface: do not raise error for empty queue status

jazi007:


zariiii9003:
This could possibly break client code that relies on these errors being raised. This is also unnecessary once #1052 is merged.
I will close this for now. You can reopen if you disagree.

PR #1056, opened 2021-05-22
add tests for PCAN

DawidRosinski:
New file `test_pcan.py` added and `tox.ini` updated to allow parameterized testing.

PR #1061, opened 2021-06-01
Fix CI timeout

zariiii9003:
Constrain dependency versions so the pip resolver can finish in adequate time. Otherwise GHA runs into timeout [as seen here](https://github.com/hardbyte/python-can/pull/1056/checks?check_run_id=2708990832#step:5:11)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -5,8 +5,8 @@ deps =
     pytest~=5.3
     pytest-timeout~=1.3
     pytest-cov~=2.8
-    coverage<5
-    codecov~=2.0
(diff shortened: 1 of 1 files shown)


felixdivo (approved):
Good catch. Also fixes Travis CI. However, it found a small pylint issue: https://travis-ci.com/github/hardbyte/python-can/jobs/510144309#L589

felixdivo:
Merging as it at least gets things to work a bit better. #1068 Should also be merged soon to get all CI green again.

PR #1062, opened 2021-06-02
Preserve capitalization when reading config files

### LINKED ISSUE #702: Ixxat driver ignores UniqueHardwareID and [r|t]xFifoSize from configuration

felixdivo:
This change was discussed in #1020, where @cranium-xx worked on a fix for #702. Previously, all parameters were converted to lower case, and some parameters with upper case naming could therefore not be set. This is fixed by this PR.

Fix #702.

### DIFF
--- a/can/util.py
+++ b/can/util.py
@@ -52,6 +52,10 @@ def load_file_config(
         name of the section to read configuration from.
     """
     config = ConfigParser()
+
+    # make sure to not transform the entries such that capitalization is preserved
+    config.optionxform = lambda entry: entry  # type: ignore
+
     if path is None:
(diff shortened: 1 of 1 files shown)


felixdivo:
@hardbyte Do you known what might cause the GitHub Actions to hang? It sat there waiting and doing nothing für 7+ hours until i canceled and restated it. But as before, it just states `Starting workflow run...`.

zariiii9003 (approved):

felixdivo:
@Mergifyio rebase

PR #1068, opened 2021-06-05
Improve IO module typing

felixdivo:
This also fixes the mypy issue found in #1061.

### DIFF
--- a/can/io/generic.py
+++ b/can/io/generic.py
@@ -4,7 +4,6 @@ Contains a generic class for file IO.
 
 from abc import ABCMeta
 from typing import (
-    Any,
     Optional,
     cast,
     Iterable,
@@ -15,6 +14,7 @@ from typing import (
     ContextManager,
 )
 from typing_extensions import Literal
+from types import TracebackType
 
 import can
 import can.typechecking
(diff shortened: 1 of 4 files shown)


felixdivo:
@Mergifyio rebase

can/io/generic.py
```diff
@@ -3,19 +3,20 @@
 """
 
 from abc import ABCMeta
-from typing import Any, Optional, cast, Union, TextIO, BinaryIO, Type
+from typing import Any, Optional, cast, Iterable, Union, TextIO, BinaryIO, Type, ContextManager
+from typing_extensions import Literal
```
zariiii9003, line 7: We do not have the `typing_extensions` dependency in setup.py. Do we need it?

felixdivo: Yeah, also realized that. It somehow was present in my venv though so I thought I'd give it a try.

I'd propose to add it, yes. It's the backport of newer typing additions to older Python versions (the `Literal` was [added in 3.8](https://docs.python.org/3/library/typing.html#typing.Literal)). Shall I add it to this PR?

zariiii9003: Personally i don't mind it.

felixdivo:
@zariiii9003 It should work now.

zariiii9003:
Mypy still does not seem happy: https://travis-ci.com/github/hardbyte/python-can/jobs/511375046#L586

felixdivo:
@zariiii9003 It seems like pylint does not recognize the type annotations that it really is an `Iterable` and reports a false positive. Everything work now. Do you approve this PR?

zariiii9003 (approved):

PR #1072, opened 2021-06-07
Add logconvert script

simontegelid:
Make use of the log readers and writers to just convert between different log file formats.

### DIFF
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -10,10 +10,9 @@ from can import LogReader, Logger, SizedRotatingLogger
 
 
 class ArgumentParser(argparse.ArgumentParser):
-
     def error(self, message):
         self.print_help(sys.stderr)
-        self.exit(errno.EINVAL, '%s: error: %s\n' % (self.prog, message))
+        self.exit(errno.EINVAL, "%s: error: %s\n" % (self.prog, message))
 
 
 def main():
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -60,8 +60,9 @@ def main():
         for m in reader:  # pylint: disable=not-an-iterable
             logger(m)
     except KeyboardInterrupt:
-        pass
+        sys.exit(1)
     finally:
+        reader.stop()
         logger.stop()
 
 
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -15,15 +15,6 @@ def main():
     )
 
     parser.add_argument(
-        "-o",
-        "--output",
-        dest="output",
-        help="Output filename, type dependent on suffix see can.LogReader.",
-        default=None,
-        required=True,
-    )
-
-    parser.add_argument(
         "-s",
         "--file_size",
         dest="file_size",
@@ -33,27 +24,29 @@ def main():
     )
 
     parser.add_argument(
-        "infile",
-        metavar="input-file",
+        "input",
+        metavar="INFILE",
         type=str,
-        help="Log file to convert from. For supported types see can.LogReader.",
+        help="Input filename. Type dependent on suffix see can.LogReader.",
     )
 
-    # print help message when no arguments were given
-    if len(sys.argv) < 2:
-        parser.print_help(sys.stderr)
-        raise SystemExit(errno.EINVAL)
+    parser.add_argument(
+        "output",
+        metavar="OUTFILE",
+        type=str,
+        help="Output filename. Type dependent on suffix see can.LogReader.",
+    )
 
-    results = parser.parse_args()
+    args = parser.parse_args()
 
-    reader = LogReader(results.infile)
+    reader = LogReader(args.input)
 
-    if results.file_size:
+    if args.file_size:
         logger = SizedRotatingLogger(
-            base_filename=results.output, max_bytes=results.file_size
+            base_filename=args.output, max_bytes=args.file_size
         )
     else:
-        logger = Logger(filename=results.output)
+        logger = Logger(filename=args.output)
 
     try:
         for m in reader:  # pylint: disable=not-an-iterable
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -41,7 +41,7 @@ def main():
         "output",
         metavar="OUTFILE",
         type=str,
-        help="Output filename. Type dependent on suffix see can.LogReader.",
+        help="Output filename. The type is dependent on the suffix, see can.Logger.",
     )
 
     args = parser.parse_args()
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -9,8 +9,15 @@ import errno
 from can import LogReader, Logger, SizedRotatingLogger
 
 
+class ArgumentParser(argparse.ArgumentParser):
+
+    def error(self, message):
+        self.print_help(sys.stderr)
+        self.exit(errno.EINVAL, '%s: error: %s\n' % (self.prog, message))
+
+
 def main():
-    parser = argparse.ArgumentParser(
+    parser = ArgumentParser(
         description="Convert a log file from one format to another.",
     )
 
--- a/can/logconvert.py
+++ b/can/logconvert.py
@@ -19,7 +19,7 @@ def main():
         "-o",
         "--output",
         dest="output",
-        help="""Output filename, type dependent on suffix see can.LogReader.""",
+        help="Output filename, type dependent on suffix see can.LogReader.",
         default=None,
         required=True,
     )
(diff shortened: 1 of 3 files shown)


simontegelid:
I consider it done. Would be great with a few more tests though, I just tested it to the same extent as the other scripts.

felixdivo:
Hey @simontegelid! This looks like a very nice addition to the library. Just a little question: Could we make the script accept the output as a second argument without specifying `-o`? For example: `logconvert.py input.file output.file`? I think that would be simpler.

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+    parser = argparse.ArgumentParser(
+        "python -m can.logconvert",
+        description="Convert a log file from one format to another.",
+    )
+
+    parser.add_argument(
+        "-o",
+        "--output",
+        dest="output",
+        help="""Output filename, type dependent on suffix see can.LogReader.""",
```
felixdivo, line 22:
```suggestion
        help="Output filename, type dependent on suffix see can.LogReader.",
```

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+        default=None,
+        required=True,
+    )
+
+    parser.add_argument(
+        "-s",
+        "--file_size",
+        dest="file_size",
+        type=int,
+        help="""Maximum file size in bytes. Rotate log file when size threshold is reached.""",
```
felixdivo, line 32:
```suggestion
        help="Maximum file size in bytes. Rotate log file when size threshold is reached.",
```

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+import sys
+import argparse
+import errno
+
+from can import LogReader, Logger, SizedRotatingLogger
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        "python -m can.logconvert",
```
felixdivo, line 14: I don't think that we need this, do we?

```suggestion
```

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    # print help message when no arguments were given
+    if len(sys.argv) < 2:
+        parser.print_help(sys.stderr)
+        raise SystemExit(errno.EINVAL)
+
+    results = parser.parse_args()
+
+    reader = LogReader(results.infile)
```
felixdivo, line 50: We should probably close the file descriptors afterwards. `with LogReader(results.infile) as reader:` and similarly for the logger should be enough.

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+            base_filename=results.output, max_bytes=results.file_size
+        )
+    else:
+        logger = Logger(filename=results.output)
+
+    try:
+        for m in reader:  # pylint: disable=not-an-iterable
+            logger(m)
+    except KeyboardInterrupt:
+        pass
```
felixdivo, line 63:
```suggestion
        print("Interrupted! Quitting.")
```

felixdivo (changes requested):

can/logconvert.py
simontegelid: I agree that just "pass" isn't the best idea here. I think it would be better with sys.exit(1) though, instead of writing something to stdout and then exit with 0. What do you think?

can/logconvert.py
simontegelid: Yes, will fix this.

can/logconvert.py
simontegelid: No, not really. But I added it to keep consistency with the other scripts. I think it should be here until all the scripts are cleaned up.

simontegelid:
> Hey @simontegelid! This looks like a very nice addition to the library. Just a little question: Could we make the script accept the output as a second argument without specifying `-o`? For example: `logconvert.py input.file output.file`? I think that would be simpler.

I think the -o flag should be there to avoid mixing positional arguments of various types to avoid issues with future updates (eg. adding support for multiple input files or something)

can/logconvert.py
felixdivo: 👍

I'll remove that line in a separate PR real quick so you can just apply this suggestion.

felixdivo: See #1099

can/logconvert.py
felixdivo: Sounds even better!

felixdivo:
> I think the -o flag should be there to avoid mixing positional arguments of various types to avoid issues with future updates

Well, one could use flags like `-o additional.file`  then too (if such features ever gets added). I'd argue that the the 1-to-1 conversion will probably stay the most used variant and should thus be the most simple to use.

simontegelid:
> > I think the -o flag should be there to avoid mixing positional arguments of various types to avoid issues with future updates
> 
> Well, one could use flags like `-o additional.file` then too (if such features ever gets added). I'd argue that the the 1-to-1 conversion will probably stay the most used variant and should thus be the most simple to use.

Fair enough. Fixed now!

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+        type=int,
+        help="Maximum file size in bytes. Rotate log file when size threshold is reached.",
+        default=None,
+    )
+
+    parser.add_argument(
+        "input",
+        metavar="INFILE",
+        type=str,
+        help="Input filename. Type dependent on suffix see can.LogReader.",
```
felixdivo, line 37:
```suggestion
        help="Input filename. Type dependent on suffix, see can.LogReader.",
```

can/logconvert.py
```diff
@@ -0,0 +1,69 @@
...
+    )
+
+    parser.add_argument(
+        "output",
+        metavar="OUTFILE",
+        type=str,
+        help="Output filename. Type dependent on suffix see can.LogReader.",
```
felixdivo, line 44:
```suggestion
        help="Output filename. The type is dependent on the suffix, see can.Logger.",
```

can/logconvert.py
felixdivo: Seems like this is the only thing left to be done before merging.

simontegelid: I solved this with a .stop() in the finally clause to avoid using contextmanagers or so due to the SizedRotatingLogger vs Logger instances. Will do the same thing as using `with`. Is that good enough?

can/logconvert.py
```diff
@@ -0,0 +1,68 @@
...
+        logger = Logger(filename=args.output)
+
+    try:
+        for m in reader:  # pylint: disable=not-an-iterable
+            logger(m)
+    except KeyboardInterrupt:
+        sys.exit(1)
+    finally:
+        reader.stop()
+        logger.stop()
```
felixdivo, line 64: I think this should do, it's a couple of lines shorter and maybe a little bit easier to understand:

```suggestion
    with LogReader(args.input) as reader:

        if args.file_size:
            logger = SizedRotatingLogger(
                base_filename=args.output, max_bytes=args.file_size
            )
        else:
            logger = Logger(filename=args.output)

        with logger:
            try:
                for m in reader:  # pylint: disable=not-an-iterable
                    logger(m)
            except KeyboardInterrupt:
                sys.exit(1)
```

can/logconvert.py
felixdivo: I commented below. I think you can just apply it as is and we can merge it. Functionally, they should do pretty much the same.

can/logconvert.py
simontegelid: Just tested it, but `SizedRotatingLogger` doesn't implement `__enter__` so `python -m can.logconvert -s 1000 log.log log.asc` doesn't work.

felixdivo: Whoops! That's a bug, I suppose. `SizedRotatingLogger` should inherit from `Logger`, which would then work since `BaseIOHandler` implements the context manager protocol.

simontegelid:
So, I rebased this whole thing, therefore the force-push. Now it works as intended, using `with logger`, etc.

felixdivo (approved):

felixdivo:
@hardbyte Do you approve?

hardbyte:
@felixdivo Indeed I do, sorry to keep you waiting for my review on so many PRs. Where you are confident of a change I'm happy for you to decide to merge without my review. 

@simontegelid thanks for the feature!

felixdivo:
@hardbyte Thanks for having a look! Your offer sounds good, I think if we have two or more reviews, it's already quite a good assurance that a change is valid and wanted. 😃

PR #1077, opened 2021-06-08
Supress 'SystemError: DeviceNotFound' in cantact detect_available_configs()

felixdivo:
This only occurs if the `cantact` library is installed but no device is present.


hardbyte (approved):

felixdivo:
@Mergifyio refresh

Issue #1046, opened 2021-05-11
Specific Exceptions: Adapting the interfaces


felixdivo:
Follow up to #1025. All interfaces need to be changed:

- [x] ics_neovi -> @felixdivo #1073; fix in #1080
- [x] ixxat -> @felixdivo already done in the original #1025
- [x] kvaser -> @felixdivo #1074
- [x] neousys -> @felixdivo #1076
- [x] pcan -> @felixdivo #1152
- [x] seeedstudio -> @felixdivo #1090
- [x] serial -> @felixdivo #1088
- [x] socketcan -> @felixdivo #1087
- [x] systec -> @felixdivo #1158
- [x] udp_multicast -> @felixdivo #1089
- [x] usb2can -> @felixdivo #1091
- [x] vector -> @felixdivo already started in the original #1025; fixed in #1052; completed in #1083
- [x] canalystii -> @projectgus #1127
- [x] cantact -> @felixdivo #1153
- [x] gs_usb -> @felixdivo #1168
- [x] iscan -> @felixdivo #1092
- [x] nixnet -> @felixdivo #1154
- [x] nican -> @felixdivo #1100
- [x] robotell -> @felixdivo #1155
- [x] slcan -> @felixdivo #1156 and #1166
- [x] virtual -> @felixdivo #1079

zariiii9003:
@felixdivo can you summarize the necessary changes?

felixdivo:
Sure: #1025 added subclass exceptions to `can.CanError`, which should now be used by the interfaces. One should go trough the interfaces one by one and replace any use of `can.CanError` by some more specific exception type (where applicable).

Does that help? :) If you want to help, just add your name or some PR link to the list above. I'll tackle this the coming week.

felixdivo:
Reviewing all interfaces is quite a bit of work. Help if of course much appreciated.

felixdivo:
THis is complete once all PRs which are linked in the issue text are merged.

PR #1091, opened 2021-06-13
Specific Exceptions: Adapting usb2can interface (+ simplify + partially type)

felixdivo:
Part of #1046.

### DIFF
--- a/can/interfaces/usb2can/usb2canabstractionlayer.py
+++ b/can/interfaces/usb2can/usb2canabstractionlayer.py
@@ -182,7 +182,9 @@ class Usb2CanAbstractionLayer:
 
     def blocking_receive(self, handle, msg, timeout) -> CanalError:
         with error_check("Blocking Receive Failed"):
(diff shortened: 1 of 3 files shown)


can/interfaces/usb2can/usb2canabstractionlayer.py
```diff
@@ -102,6 +102,14 @@ class CanalMsg(Structure):
     ]
 
 
+@contextmanager
+def error_check(error_message: str) -> None:
+    try:
+        yield
+    except Exception as error:
+        raise can.CanOperationError(error_message) from error
```
hardbyte, line 110: That's neat

hardbyte (approved):

can/interfaces/usb2can/usb2canabstractionlayer.py
felixdivo: Thank's Brian for the chunk of reviews!

PR #1069, opened 2021-06-05
Test on 3.10.0-beta.2

zariiii9003:
Pytest update was necessary due to https://github.com/pytest-dev/pytest/pull/8540

<details><summary>Trace</summary>
<p>

```
GLOB sdist-make: /home/runner/work/python-can/python-can/setup.py
gh create: /home/runner/work/python-can/python-can/.tox/gh
gh installdeps: pytest~=5.3, pytest-timeout~=1.3, pytest-cov~=2.8, coverage==4.5.*, codecov==2.1.10, hypothesis~=4.56, pyserial~=3.0
gh inst: /home/runner/work/python-can/python-can/.tox/.tmp/package/1/python-can-4.0.0.dev2.zip
gh installed: attrs==21.2.0,certifi==2021.5.30,chardet==4.0.0,codecov==2.1.10,coverage==4.5.4,hypothesis==4.57.1,idna==2.10,more-itertools==8.8.0,msgpack==1.0.2,mypy-extensions==0.4.3,packaging==20.9,pluggy==0.13.1,py==1.10.0,pyparsing==2.4.7,pyserial==3.5,pytest==5.4.3,pytest-cov==2.10.1,pytest-timeout==1.4.2,python-can @ file:///home/runner/work/python-can/python-can/.tox/.tmp/package/1/python-can-4.0.0.dev2.zip,requests==2.25.1,sortedcontainers==2.4.0,urllib3==1.26.5,wcwidth==0.2.5,wrapt==1.12.1
gh run-test-pre: PYTHONHASHSEED='4019826288'
gh run-test: commands[0] | pytest
Traceback (most recent call last):
  File "/home/runner/work/python-can/python-can/.tox/gh/bin/pytest", line 8, in <module>
    sys.exit(main())
  File "/home/runner/work/python-can/python-can/.tox/gh/lib/python3.10/site-packages/_pytest/config/__init__.py", line 105, in main
    config = _prepareconfig(args, plugins)
  File "/home/runner/work/python-can/python-can/.tox/gh/lib/python3.10/site-packages/_pytest/config/__init__.py", line 257, in _prepareconfig
    return pluginmanager.hook.pytest_cmdline_parse(
  File "/home/runner/work/python-can/python-can/.tox/gh/lib/python3.10/site-packages/pluggy/hooks.py", line 286, in __call__
    return self._hookexec(self, self.get_hookimpls(), kwargs)
[...]

PR #1094, opened 2021-06-14
Add new language classifiers to setup.py

felixdivo:
Shouldn't be too controversial. We strive to support Python 3.10 and we are already compatible with `3.10.0-beta.2` (see #1069).


hardbyte (approved):
👍🏼

Issue #972, opened 2021-01-25
Calculate Bus-Load 


invertium:
A feature to calculate the bus load in percent over a given timeframe would be very useful.

PR #1105, opened 2021-07-09
WIP: Add method to calculate load on a bus

wbarnha:
PR created to address Issue #972. WIP since I'm not done testing this.

I also preemptively opened a PR because I'm having some difficulty determining what the bitrate of a particular CAN interface is configured to be. If anyone could point me in the right direction, I'd be grateful. I saw that some custom interfaces have bitrates hard-coded in their respective constants.py files.


hartkopp:
You should take a look into `canbusload` from [SocketCAN CAN utils:](https://github.com/linux-can/can-utils/)
https://github.com/linux-can/can-utils/blob/master/canbusload.c

At least when you are working on Linux (Debian/Ubuntu) `apt install can-utils` should work for you as it installs `canbusload` too.

wbarnha:
> You should take a look into `canbusload` from [SocketCAN CAN utils:](https://github.com/linux-can/can-utils/)
> https://github.com/linux-can/can-utils/blob/master/canbusload.c
> 
> At least when you are working on Linux (Debian/Ubuntu) `apt install can-utils` should work for you as it installs `canbusload` too.

Thank you for the suggestion. It seems the method I implemented is somewhat different since it's intended to be used as a method of an object such as ThreadSafeBus. I have tested my code but have not hardcoded test cases.

wbarnha:
Going to give this some more attention- I'd like to add additional functionality so that a thread is spawned to track CAN traffic in the background.

wbarnha:
No longer need this feature.

JayTeeBat:
Hi @wbarnha,
I was interested in this feature, do you mind explaining why it is not needed anymore?

Thx

wbarnha:
Long story short, this was written when I was not aware of all tools in `can-utils`. I believe there's an application that tracks bitrate for you, but unfortunately I cannot remember its name.

Issue #935, opened 2020-11-14
Slow library import time 


kjngineering:
I have been attempting to use python-can on an embedded buildroot system, which essentially will be a CAN based display.
Hardware is relatively low level, single ARM Cortex-A7 1.2GHz, 64mb ram, SPI can interface, SD card based storage/boot media.

I had done some function tests with python-can originally and it had worked as intended, but now I am getting to optimisations and I am stumped. Now I have introduced python-can into my runtime, library import -- "import can" is taking a huge block of time which is interfering with the user experience.

As library imports are blocking (as far as I know) its causing a blank screen for the duration of the load time until the display library can make a call and refresh the display.

My typical load times for the can library are in the 2800ms > 3000ms range, which is forever compared to the rest of the process, some benchmarks for reference:
- Power on to boot screen ~ 1200ms
- Power on to user environment ~1800ms
- Script runtime to user display ~500ms (no import-can)
- Script runtime to user display ~3500ms (with import-can)
- Configure can interface (socket can) start receiving packets once library is loaded (10ms)

I'm using the bare minimum of the library:
- SocketCan interface
- Basic Filters
- Single Listener

No logger, no asnycio,  no other interfaces or  parts of the API.

Is this load time normal? 
or is there an issue which is slowing down the use of the library? 
what avenues exist to speed up this process? (selective loading?)

some data from the target hardware:

```
# python -c "import can; print(can.__version__)"
3.3.4
```

```
python -X importtime -c 'import can'
import time: self [us] | cumulative | imported package
import time:      8001 |       8001 | _frozen_importlib_external
import time:      2443 |       2443 |   time
import time:      4229 |       6671 | zipimport
import time:      1298 |       1298 |     _codecs
import time:      8647 |       9945 |   codecs
import time:      5928 |       5928 |   encodings.aliases
import time:      9471 |      25343 | encodings
import time:      4001 |       4001 | encodings.utf_8
import time:      1948 |       1948 | _signal
import time:      5758 |       5758 | encodings.latin_1
import time:      1175 |       1175 |     _abc
import time:      5806 |       6980 |   abc
import time:      5761 |      12741 | io
import time:      1500 |       1500 |       _stat
import time:      5071 |       6571 |     stat
import time:     21029 |      21029 |     _collections_abc
import time:      3479 |       3479 |       genericpath
import time:      4619 |       8097 |     posixpath
import time:     11335 |      47031 |   os
import time:      4056 |       4056 |   _sitebuiltins
[...]

PR #1109, opened 2021-07-21
Faster import

wbarnha:
I created this PR to address issue #935, since I was having similar issues with imports. Originally, `python-can` took four seconds to import. After removing the call to run `iter_entry_points`, the import time was down to three seconds. Removing the import call for `pkg_resources` reduced the import time by two seconds down to only needing one second for importing `python-can`.

I manually ran `iter_entry_points` in a similar setup to see if the dictionary for `BACKENDS` would change, but it seems nothing would change at all. Since everything is manually defined, it seems there's no need for `iter_entry points`?

Am I missing something here? I tested these changes with `socketcan` and I didn't have any issues.


wbarnha:
Whoops, I messed up creating a new branch.

PR #1110, opened 2021-07-21
Speed up imports by removing pkg_resources

wbarnha:
I created this PR to address issue #935, since I was having similar issues with imports. This PR is built off my previous PR  #1105 since I think this change warrants more discussion.

Originally, python-can took four seconds to import. After removing the call to run `iter_entry_points`, the import time was down to three seconds. Removing the import call for `pkg_resources` reduced the import time by two seconds down to only needing one second for importing python-can.

I manually ran `iter_entry_points` in a similar setup as when the import process is run, to see if the dictionary for `BACKENDS` would change, but it seems nothing would change at all. Since everything is manually defined, it seems there's no need for `iter_entry points`?

Am I missing something here? I tested these changes with `socketcan` and I didn't have any issues.

### DIFF
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -33,8 +33,8 @@ try:
     if 'can.interface' in entry:
         BACKENDS.update({interface.name: tuple(interface.value.split(':')) for interface in entry['can.interface']})
 except ImportError:
-    from pkg_resources import iter_entry_points as entry_points
-    entry = entry_points("can.interface")
+    from pkg_resources import iter_entry_points
+    entry = iter_entry_points("can.interface")
     BACKENDS.update({interface.name: (interface.module_name, interface.attrs[0]) for interface in entry})
 
 VALID_INTERFACES = frozenset(list(BACKENDS.keys()))
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -2,10 +2,6 @@
 Interfaces contain low level implementations that interact with CAN hardware.
 """
 
-import warnings
-from pkg_resources import iter_entry_points
-
-
 # interface_name => (module, classname)
 BACKENDS = {
     "kvaser": ("can.interfaces.kvaser", "KvaserBus"),
@@ -30,12 +26,5 @@ BACKENDS = {
     "nixnet": ("can.interfaces.nixnet", "NiXNETcanBus"),
     "neousys": ("can.interfaces.neousys", "NeousysBus"),
 }
-
-BACKENDS.update(
-    {
-        interface.name: (interface.module_name, interface.attrs[0])
-        for interface in iter_entry_points("can.interface")
-    }
-)
 
 VALID_INTERFACES = frozenset(list(BACKENDS.keys()))
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -30,8 +30,9 @@ BACKENDS = {
 try:
     from importlib.metadata import entry_points
     entry = entry_points()
-    BACKENDS.update({interface.name: tuple(interface.value.split(':')) for interface in entry['can.interface']})
-except:
+    if 'can.interface' in entry:
+        BACKENDS.update({interface.name: tuple(interface.value.split(':')) for interface in entry['can.interface']})
+except ImportError:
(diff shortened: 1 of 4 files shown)


christiansandberg:
This is used for loading external interfaces as plugins (see https://python-can.readthedocs.io/en/develop/interfaces.html) so it's not possible to just remove it. You could however use [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#entry-points) instead if possible, and fallback to setuptools on older Python versions.

wbarnha:
> This is used for loading external interfaces as plugins (see https://python-can.readthedocs.io/en/develop/interfaces.html) so it's not possible to just remove it. You could however use [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#entry-points) instead if possible, and fallback to setuptools on older Python versions.

Okay, this was the response I was hoping to hear. I'll try the suggested changes and commit some additional changes

can/interfaces/__init__.py
```diff
@@ -27,4 +32,11 @@
     "neousys": ("can.interfaces.neousys", "NeousysBus"),
 }
 
+BACKENDS.update(
+    {
+        interface.name: (interface.module_name, interface.attrs[0])
```
christiansandberg, line 37: I think importlib uses slightly different attributes. You could test it by installing python-can-remote (on both Python 3.9 and older):

```
pip install python-can-remote
```

wbarnha: That helped very much, thank you. It turns out I was misreading the documentation for importlib. I have gone ahead and added a fix that properly utilizes importlib now.

wbarnha:
I think the last solution to this problem is making some architectural changes to how bus objects are initialized. A suggestion was made to initialize plugins on-demand, in order to reduce import time for other users.

christiansandberg:
Just revert bus.py since it is related to another PR. Then I think it looks good.

wbarnha:
Made the mistake of rebasing and had to revert my changes for canbus load calculations, but things should be fine.

I made some additional changes to allow older versions than Python 3.8 to only load user-installed interfaces on-demand. I really would like these features backported as well in order to improve import time for older versions of Python. But this brings up an issue where I can no longer have `can.VALID_INTERFACES` including the user-installed interfaces at runtime. I am willing to revert the changes so that the only change included is importlib.

christiansandberg:
It's probably a limited number of users that rely on VALID_INTERFACES but since it is a breaking change I'm hesitant to include that. If one wants to take advantage of all optimizations they should update Python and other packages to the latest version. Python 3.9 will soon be the most commonly used version as it gets included in various distributions.

wbarnha:
I think that's a reasonable issue. I'll go ahead and revert the last commit.

christiansandberg (approved):

pierreluctg:
Typically the imports would stay at the top of the module and use a feature flag (like `use_importlib_metadata` for example) to drive the logic bellow.

can/interfaces/__init__.py
```diff
@@ -31,11 +27,14 @@
...
-    }
-)
+try:
+    from importlib.metadata import entry_points
+    entry = entry_points()
+    if 'can.interface' in entry:
+        BACKENDS.update({interface.name: tuple(interface.value.split(':')) for interface in entry['can.interface']})
+except ImportError:
+    from pkg_resources import iter_entry_points as entry_points
+    entry = entry_points("can.interface")
```
pierreluctg, line 37: What drives the need to rename the import here?

wbarnha: It was originally leftover code to make things more concise, but this renamed import can be changed if you like.

hardbyte (approved):
Nice, I wasn't aware of importlib's `entry_points` api.

can/interfaces/__init__.py
pierreluctg: Typically the imports would stay at the top of the module and use a feature flag (like `use_importlib_metadata` for example) to drive this logic here.

Up to you if you want to keep renaming the import or not. Not renaming makes the code easier to read IMO.

wbarnha: I'm indifferent, but since you believe the code is easier to read this way, I'll go ahead and make the change for you.

pierreluctg (approved):

PR #1114, opened 2021-07-28
canalystii: Fix transmitting onto a busy bus

projectgus:
This is to fix the issue that the CANalystIIBus interface works fine in a quiet setup (like a loopback test) but randomly fails to send messages if the bus is busy.

The "SendType" field has at least two possible values, and value `1` causes the firmware to drop the message if there is bus contention when sending it.

I can make this into a configuration parameter if you prefer, just let me know. I was going to do this but then I couldn't really think of a use case for unreliable transmission. (With `SendType 1`, there isn't anything in the protocol to tell the software that the message was dropped, it's just silently dropped!)

Also made a small addition in the docs to note the binary library dependency.

Thanks for maintaining this library, it's been a pleasure to work with! (Well, after I worked out why I was randomly losing messages! :laughing:)

### DIFF
--- a/can/interfaces/canalystii.py
+++ b/can/interfaces/canalystii.py
@@ -32,6 +32,9 @@ class VCI_CAN_OBJ(Structure):
     ]
 
 
+SENDTYPE_ALLOW_RETRY = 0  # Retry send if there is bus contention
+SENDTYPE_ONE_SHOT = 1  # Drop the message if transmission fails first time
+
 VCI_USBCAN2 = 4
 
 STATUS_OK = 0x01
@@ -145,7 +148,7 @@ class CANalystIIBus(BusABC):
             msg.arbitration_id,
             0,
             0,
-            1,
+            SENDTYPE_ALLOW_RETRY,
             msg.is_remote_frame,
             extern_flag,
             msg.dlc,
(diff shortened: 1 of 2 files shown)


projectgus:
Also, I have a question: initially when I was debugging this I thought the root cause was a USB race condition in the firmware buffers so I spent some time reverse engineering the USB protocol. In fact I've recreated the entire driver without the binary library, using [libusb1](https://pypi.org/project/libusb1/). This version has (I think) the same functionality, probably slightly higher performance, and allows customizing the number of messages buffered by software.

... then I realised about the SendType parameter, of course.

If I polished the libusb based driver up a bit, would you be interested in replacing the current binary library driver?

hardbyte:
> Also, I have a question: initially when I was debugging this I thought the root cause was a USB race condition in the firmware buffers so I spent some time reverse engineering the USB protocol. In fact I've recreated the entire driver without the binary library, using [libusb1](https://pypi.org/project/libusb1/). This version has (I think) the same functionality, probably slightly higher performance, and allows customizing the number of messages buffered by software.
> 
> ... then I realised about the SendType parameter, of course.
> 
> If I polished the libusb based driver up a bit, would you be interested in replacing the current binary library driver?

Yes that sounds great! If you package it up for Pypi we can set it as an optional dependency and have another backend that doesn't require manual driver installation.

hardbyte (approved):

projectgus:
> Yes that sounds great! If you package it up for Pypi we can set it as an optional dependency and have another backend that doesn't require manual driver installation.

@hardbyte No problems, will try to sort this out in the next few weeks and submit a new PR for it. To check I'm understanding correctly: you'd prefer me to follow a model like the gs_usb support - where I maintain a "canalystii" package in pypi, rather than putting that part of the driver into python-can?

BTW, please let me know if you need me to do anything to pass the CI check.

PR #1116, opened 2021-08-13
fix(ctypesutil): correctly apply wrapped lib function as CLibrary member

ausserlesh:
While using the IXXAT driver we observed that the errcheck function `can.interfaces.ixxat.canlib.__check_status` is never called after library function calls, preventing the detection of timeouts etc. by the caller. The root cause is that the original library function instead of the wrapped function is called.


hardbyte (approved):
Thanks!

PR #1125, opened 2021-09-09
Avoid flooding the logger with ~500 errors when they are the same

pierreluctg:
Only log/warn each unique error once with repeat count, when needed.
Adding dot to separate end the short description sentence before the long description.


hardbyte (approved):
handy feature

felixdivo (approved):

PR #1127, opened 2021-09-12
canalystii: Replace binary library with python driver

projectgus:
Replaces the binary ControlCAN.dll/controlcan.so library with a Python implementation of the driver: https://pypi.org/project/canalystii/

## Improvements

* When bus is configured with multiple channels, messages can be received from all of them (previous implementation of `recv()` would always receive from the first-configured channel).
* Fixes the issue with `send()` on a busy bus that was also noted in #1114
* Possible to configure the software RX message queue size.
* Send timeout now relates to the message being sent on the bus, not the USB layer timeout (within constraints of what the USB protocol allows us to detect).
* Adds MacOS support (currently untested, I've verified that Linux and Windows work but haven't tested MacOS. Of course as there is currently no MacOS support, it's not going to be broken either!)
* Adds support for non-x86 Linux such as ARM (also untested, but assuming it will work same as on x86).

## Performance

Replacing native code with Python code creates some performance impact. On my Linux machine (i7-6500U, approx 5 years old) current 'develop' branch uses 27% CPU to run can.logger at maximum message rate (~7800 messages/sec at 1Mbps bitrate). On this branch, it increases to 38% CPU. Logging two channels simultaneously at maximum rate is not possible on 'develop', but on this branch CPU usage is 44%.

My method to measure this is pretty unscientific: `time timeout 20 python -m can.logger -i canalystii -c 0 -b 1000000 -f spammer.txt` and then use a script to check no messages were dropped in the log file. Sender is [this Arduino Sketch](https://github.com/projectgus/python-canalystii/blob/main/tests/CAN_Spammer.ino).


projectgus:
@hardbyte I added some tests for the canalystii interface (mocking at the driver interface layer), when you have a minute could you please trigger another workflow run?

(Alternatively, if you're amenable to it, you could merge my approved smaller PR #1114 and I'll rebase this branch onto it - then I won't be a first-time contributor and workflows should trigger automatically. :slightly_smiling_face:.)

hardbyte:
Have squashed your other PR. Appreciated it if if you wouldn't mind rebasing this one

projectgus:
@hardbyte Thanks! Have rebased and squashed the commits in this branch, too.

PR #1128, opened 2021-09-14
add socketcan parameter to ignore CAN error frames

nbusser:
I noticed that, by default, the error frames are always filtered by socketcan busses.
Thus, recv method on the bus will always capture these error frames, autonomously of user-defined filters.
I added an optional parameter in the \_\_init\_\_ method of SocketcanBus, allowing user to ignore all received error frames.


felixdivo (approved):
Looks useful!

nbusser:
I bump the PR. It is open for two months now.

pierreluctg (approved):

hardbyte (approved):

PR #1140, opened 2021-10-17
socketcand Backend for python-can

domologic:
We are using a Raspberry PI with MCP2515-based CAN-HAT, running a a [socketcand](https://github.com/linux-can/socketcand) as "IP-to-CAN-Gateway damon". For this, we have implemented a new backend.
socketcand uses an TCP-based ASCII protocol to represent the CAN frames. The most relevant advantage against an UDP-based protocol is that the order of messages is guaranteed.

It would be nice if you could merge this into your repository. Please let me know if we have something to add before you can merge the code.

### DIFF
--- a/can/interfaces/socketcand/__init__.py
+++ b/can/interfaces/socketcand/__init__.py
@@ -1,0 +1,9 @@
+"""
+Interface to socketcand
+see https://github.com/linux-can/socketcand
+
+Copyright (C) 2021  DOMOLOGIC GmbH
+http://www.domologic.de
+"""
+
+from .socketcand import SocketCanDaemonBus
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -6,21 +6,6 @@ Authors: Marvin Seiler, Gerrit Telkamp
 
 Copyright (C) 2021  DOMOLOGIC GmbH
 http://www.domologic.de
-
-
-This program is free software; you can redistribute it and/or
-modify it under the terms of the GNU Lesser General Public
-License as published by the Free Software Foundation; either
-version 3 of the License, or (at your option) any later version.
-
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
-Lesser General Public License for more details.
-
-You should have received a copy of the GNU Lesser General Public License
-along with this program; if not, write to the Free Software Foundation,
-Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """
 import can
 import socket
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -18,8 +18,6 @@ import traceback
 from collections import deque
 
 log = logging.getLogger(__name__)
-log.setLevel(logging.INFO)
-
 
 def convert_ascii_message_to_can_message(ascii_message: str) -> can.Message:
     if not ascii_message.startswith("< frame ") or not ascii_message.endswith(" >"):
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -32,6 +32,7 @@ from collections import deque
 
 log = logging.getLogger(__name__)
 
+
 def convert_ascii_message_to_can_message(ascii_msg: str) -> can.Message:
     if not ascii_msg.startswith("< frame ") or not ascii_msg.endswith(" >"):
         log.warning(f"Could not parse ascii message: {ascii_msg}")
@@ -86,7 +87,7 @@ class SocketCanDaemonBus(can.BusABC):
         self.__receive_buffer = ""  # i know string is not the most efficient here
         connect_to_server(self.__socket, self.__host, self.__port)
         self._expect_msg("< hi >")
-        
+
         log.info(
             f"SocketCanDaemonBus: connected with address {self.__socket.getsockname()}"
         )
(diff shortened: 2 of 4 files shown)


felixdivo:
ping @domologic

domologic:
Felix, thanks for your ping and sorry that I've lost this scope.
You are right - it does not make sense to set the logging level here. I've changed this in our branch.

domologic:
Felix, please check again. Some doc is also available, you may merge now.

can/interfaces/socketcand/socketcand.py
```diff
@@ -43,8 +43,8 @@ def convert_can_message_to_ascii_message(can_message: can.Message) -> str:
     data = can_message.data
     length = can_message.dlc
     bytes_string = " ".join("{:x}".format(x) for x in data[0:length])
-    ascii_message = f"< send {can_id:X} {length:X} {bytes_string} >"
-    return ascii_message
+    ascii_msg = f"< send {can_id:X} {length:X} {bytes_string} >"
+    return ascii_msg
```
felixdivo, line 47:
```suggestion
    return f"< send {can_id:X} {length:X} {bytes_string} >"
```

can/interfaces/socketcand/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
+Interface to socketcand
+see https://github.com/linux-can/socketcand
+
+Copyright (C) 2021  DOMOLOGIC GmbH
+http://www.domologic.de
```
felixdivo, line 6: A HTTPS link would be nice, can it be that your website does not support that? 🙃

doc/interfaces/socketcand.rst
```diff
@@ -0,0 +1,113 @@
...
+    RestartSec=100ms
+    EOT
+
+Enable ``systemd-networkd`` on reboot and start it immediately (if it was not already startet):
+
+.. code-block:: bash
+
+    sudo systemctl enable systemd-networkd
+    sudo systemctl start systemd-networkd
+
```
felixdivo, line 75: This is in principle a nice tutorial, but it does not really explain the interface in particular. (Do you agree?) We could also just reference the socketcan page here, since any set up socketcan interface would work, right?

can/interfaces/socketcand/__init__.py
domologic: Yes, HTTPS is currently not supported - the website is quite old, it is only a kind of virtual name card . We will support it HTTPS by beginning of next year. Is the http link a problem for you?

doc/interfaces/socketcand.rst
domologic: Yes, the tutorial provides much more information than really necessary. The idea was to give a full setup tutorial, what is currently hard to find, and what is also not provided by sockencand. But I can remove this section if you want.

can/interfaces/socketcand/__init__.py
felixdivo: We can leave it there, I just wanted to give you a "heads up" on this.

doc/interfaces/socketcand.rst
```diff
@@ -0,0 +1,113 @@
...
+    Timestamp: 1637791111.609763    ID: 0000031d    X Rx                DLC:  8    16 27 d8 3d fe d8 31 24
+    Timestamp: 1637791111.634630    ID: 00000587    X Rx                DLC:  8    4e 06 85 23 6f 81 2b 65
+
+The following section will show how to get the stuff installed on a Raspberry Pi with a MCP2515-based
+CAN interface, e.g. available from Waveshare: https://www.waveshare.com/rs485-can-hat.htm .
+
+Install CAN Interface for a MCP2515 based interface on a Raspberry Pi
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Add the following lines to ``/boot/config.txt``. Please take care on the frequencsy of the crystal on your MCP2515 board:
```
felixdivo, line 46:
```suggestion
Add the following lines to ``/boot/config.txt``. Please take care on the frequency of the crystal on your MCP2515 board:
```

doc/interfaces/socketcand.rst
felixdivo: It is a useful and well-written tutorial indeed! So first of all, thank you for contributing it.

It's fine if we add a few sentences on it working with any socketcan connection, not just the one you mentioned. It should be clear that it is only an example.

domologic: Ok, I've split it off a bit as a separate section "Socketcand Quickstart", as it is done in the socketcan interface doc.

doc/interfaces/socketcand.rst
domologic: Thanks, the typo is fixed now

doc/interfaces/socketcand.rst
```diff
@@ -0,0 +1,116 @@
...
+    Timestamp: 1637791111.409415    ID: 000005f7    X Rx                DLC:  8    86 de e6 0f 42 55 5d 39
+    Timestamp: 1637791111.434377    ID: 00000665    X Rx                DLC:  8    97 96 51 0f 23 25 fc 28
+    Timestamp: 1637791111.609763    ID: 0000031d    X Rx                DLC:  8    16 27 d8 3d fe d8 31 24
+    Timestamp: 1637791111.634630    ID: 00000587    X Rx                DLC:  8    4e 06 85 23 6f 81 2b 65
+
+Socketcand Quickstart
+---------------------
+
+The following section will show how to get the stuff installed on a Raspberry Pi with a MCP2515-based
+CAN interface, e.g. available from Waveshare: https://www.waveshare.com/rs485-can-hat.htm .
```
felixdivo, line 44:
```suggestion
CAN interface, e.g. available from `Waveshare <https://www.waveshare.com/rs485-can-hat.htm>`__.
However, it will also work with any other socketcan device.
```

domologic:
Felix, please check again

felixdivo:
Why did you add the license in bed0ade? I don't think that is needed and it can be removed. It's already in the project root.

domologic:
Ah ok. Its removed now.

felixdivo (approved):
Looks good!

One could add tests (see [comment](https://github.com/hardbyte/python-can/pull/1140#issuecomment-951813898)) (especially appealing here, as this driver/backend does not need hardware), but it's not required.

domologic:
Yes, I would also like to see tests here. This would need a kind of "virtual remote station", implementing the same protocol. I hope we can contribute this in future.

felixdivo:
> Yes, I would also like to see tests here.

So shall we wait with merging until then?

domologic:
No, please erge. We'll contribute tests later, but I can not say when.

felixdivo:
Great, @domologic! Looking forward to more contributions like this :)

PR #1147, opened 2021-10-26
Improve IO class hierarchy + Allow to use SizedRotatingLogger as a context manager

felixdivo:
This PR

- Allows to use all `BaseRotatingLogger` instances to be used as context managers
- Adds a test for `SizedRotatingLogger` as a context manager and cleans up some other tests
- Therefore, it allows to finish #1072
- Makes `BaseRotatingLogger::_get_new_writer` private since it should only be used internally (else, one could have forgotten to close the old file)
- Small documentation improvements
- Reworks the type hierachy: Now most IO readers correctly subclass `FileIOMessageWriter` (previously, it was unused). Now `BaseRotatingLogger::_get_new_writer` looks for that class in order to determine whether the file is supported or not
- Adds some missing types here and there (`can.typechecking.AcceptedIOType` is not an `Optional` any more; its usages in `can.util` were changed accordingly)
- Bumps the CI tool versions

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -7,7 +7,6 @@ Example .asc files:
 """
 
 from typing import cast, Any, Generator, IO, List, Optional, Union, Dict
-from can import typechecking
 
 from datetime import datetime
 import time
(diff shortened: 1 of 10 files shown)


hardbyte (approved):

felixdivo:
Thanks guys!

PR #1148, opened 2021-10-26
Tiny: Fix the way CI status conditions are specified for Mergify

felixdivo:
Apparently [our setup does not work](https://github.com/hardbyte/python-can/runs/4006357753).
Instead, we should do it [like in this repo](https://github.com/Mergifyio/mergify-engine/blob/main/.mergify.yml#L14-L19).

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,7 +1,8 @@
 queue_rules:
   - name: default
     conditions:
-      - "status-success=test,format"  # "GitHub Actions works slightly differently [...], only the job name is used."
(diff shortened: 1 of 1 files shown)


felixdivo:
@hardbyte Do you need anything for the review? If this PR is merged, most/all of my other PRs should go green.

hardbyte (approved):

hardbyte:
Agh my bad, missed this one. Sorry that you had a ping me @felixdivo

felixdivo:
No problem 😃

felixdivo:
Seems like that didn't fully work, but also the other PRs do not really depend on this.

PR #1152, opened 2021-10-27
Specific Exceptions: Adapting pcan interface

felixdivo:
Part of #1046.

Also partly cleans up the docstrings in `basic.py` and uses some subtests in `test_pcan.py` (the parameter `name` was else unused in most parameterized tests).


can/interfaces/pcan/pcan.py
```diff
@@ -2,15 +2,18 @@
 Enable basic CAN over a PCAN USB device.
 """
 
+from .basic import *  # Do this first since the wildcard imports a lot
```
zariiii9003, line 5: Why did you put this above the builtins?

zariiii9003 (approved):

can/interfaces/pcan/pcan.py
felixdivo: Ah, good point. I think it was due to a linter complaining. I'm not sure though ... it seemed clever at the time. ^^ Should I change it?

zariiii9003: Unless there's a good reason i would keep the usual order of imports: builtins, third party libraries and then the rest. But it is not a big deal. I'd rather remove the wildcard import altogether, but that's outside the scope of your PR. :smile:

felixdivo: Ah, since the wildcard cause the issues in the first place, I actually removed it. To somewhat reduce the amount of imports, I moved three simple dicts/lists to the `basic.py`. Reviewing the commit in isolation is probably the easiest, the changes are really simple.

can/interfaces/pcan/pcan.py
```diff
@@ -651,11 +597,12 @@ def _detect_available_configs():
     def status_string(self) -> Optional[str]:
         """
         Query the PCAN bus status.
-        :return: The status in string.
+
+        :return: The status description, if any was found.
         """
-        if self.status() in PCAN_DICT_STATUS:
+        try:
```
felixdivo, line 603: This change prevents a hypothetical error if `self.status()` changes between lines.

felixdivo:
@Mergifyio rebase

can/interfaces/pcan/pcan.py
```diff
@@ -5,12 +5,54 @@
...
-from can.util import len2dlc, dlc2len
-from .basic import *
+
+from ...message import Message
+from ...bus import BusABC, BusState
+from ...util import len2dlc, dlc2len
+from ...exceptions import CanError, CanOperationError, CanInitializationError
+
+
+from .basic import (
```
zariiii9003, line 18: We could also just use a namespace e.g. `basic.PCAN_BITRATES`.

felixdivo: Sure, also possible. I found it to be redundant since everything is prefixed with `PCAN_` anyways and it makes the class code more readable. The imports are simple to skim anyhow.

felixdivo:
What a [weird failure](https://github.com/hardbyte/python-can/runs/4209517987?check_suite_focus=true#step:5:449):

```
_____________ TestBaseRotatingLogger.test_rotate_without_rotator ______________

self = <test.test_rotating_loggers.TestBaseRotatingLogger object at 0x000001b81f183590>
tmp_path = WindowsPath('C:/Users/runneradmin/AppData/Local/Temp/pytest-of-unknown/pytest-0/test_rotate_without_rotator0')

    def test_rotate_without_rotator(self, tmp_path):
        logger_instance = self._get_instance(tmp_path)
    
        source = str(tmp_path / "source.txt")
        dest = str(tmp_path / "dest.txt")
    
        assert os.path.exists(source) is False
        assert os.path.exists(dest) is False
    
        logger_instance._get_new_writer(source)
        logger_instance.stop()
    
        assert os.path.exists(source) is True
        assert os.path.exists(dest) is False
    
>       logger_instance.rotate(source, dest)

test\test_rotating_loggers.py:99: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <test.test_rotating_loggers.TestBaseRotatingLogger._get_instance.<locals>.SubClass object at 0x000001b81f183a28>
source = 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_rotate_without_rotator0\\source.txt'
dest = 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_rotate_without_rotator0\\dest.txt'

    def rotate(self, source: StringPathLike, dest: StringPathLike) -> None:
        """When rotating, rotate the current log.
    
        The default implementation calls the :attr:`rotator` attribute of the
        handler, if it's callable, passing the source and dest arguments to
        it. If the attribute isn't callable (the default is `None`), the source
        is simply renamed to the destination.
    
        :param source:
            The source filename. This is normally the base
            filename, e.g. `"test.log"`
        :param dest:
            The destination filename. This is normally
            what the source is rotated to, e.g. `"test_#001.log"`.
        """
        if not callable(self.rotator):
            if os.path.exists(source):
>               os.rename(source, dest)
E               PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_rotate_without_rotator0\\source.txt' -> 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_rotate_without_rotator0\\dest.txt'

can\io\logger.py:177: PermissionError

```

felixdivo:
> What a [weird failure](https://github.com/hardbyte/python-can/runs/4209517987?check_suite_focus=true#step:5:449):

Will be fixed by a6256e8 as soon as #1166 is merged

felixdivo:
@Mergifyio rebase

felixdivo:
@hardbyte Do you think you'll have time to review the open RPs for the [4.0.0 Release](https://github.com/hardbyte/python-can/milestone/8) in the near future? There are also a few PRs that are approved by others and are waiting to be merged. It would be nice to include them into the next release (but the main blocker are the PRs around *Specific Exceptions*).

PR #1154, opened 2021-10-27
Specific Exceptions: Adapting nixnet interface

felixdivo:
Part of #1046.

Since the *NI-XNET* interface was not released in a stable release yet (see #968), we can freely change the exception handling without deprecation notices.

### DIFF
--- a/can/interfaces/nixnet.py
+++ b/can/interfaces/nixnet.py
@@ -13,7 +13,9 @@ import sys
 import time
 import struct
 
-from can import CanError, BusABC, Message
+from can import BusABC, Message
+from ..exceptions import CanInitializationError, CanOperationError
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


felixdivo:
@Mergifyio rebase

felixdivo:
I'd appreciate a review of this change

felixdivo:
I'd appreciate a review of this change

can/interfaces/nixnet.py
```diff
@@ -120,16 +118,18 @@ def __init__(
             self.__session_send.start()
             self.__session_receive.start()
 
-        except errors.XnetError as err:
-            raise NiXNETError(function="__init__", error_message=err.args[0]) from None
+        except errors.XnetError as error:
+            raise CanInitializationError(
+                f"{error.args[0]} ({error.error_type})", error.error_code
+            ) from None
```
hardbyte, line 124: `raise from None` is to explicitly remove the earlier exception from the traceback right?

hardbyte (approved):

can/interfaces/nixnet.py
felixdivo: Yes. 😄 But all the information should now be in the new exception. That's also how it was before.

felixdivo:
Thanks Brian!

PR #1159, opened 2021-10-30
Add changed byte highlighting to viewer.py

MattWoodhead:
I have been using python-can for message sniffing on a Tractor control bus, and a feature I had seen in some other message viewers was highlighting bytes when their value changes.
I have added this functionality (toggleable - off by default) to viewer.py, and found it useful for what I was trying to achieve so thought I would share it here.

viewer

viewer_changed_bytes_highlighting

I hope sending a PR like this out of the blue is OK, if not please reject :)


can/viewer.py
```diff
@@ -353,6 +390,7 @@ def parse_args(args):
         "\n        | ESQ/q   | Exit the viewer         |"
         "\n        | c       | Clear the stored frames |"
         "\n        | s       | Sort the stored frames  |"
+        "\n        | h       | Toggle byte highlights  |"
```
felixdivo, line 393: Maybe better state that byte *changes* are visualized?

```suggestion
        "\n        | h       | Toggle highlight byte changes |"
```

can/viewer.py
```diff
@@ -460,6 +498,14 @@ def parse_args(args):
         default=2,
     )
 
+    # suppressed as the user doesn't need to change it
+    optional.add_argument(
```
felixdivo, line 502: Why is this change required?

felixdivo (changes requested):
Cool feature! Thanks for the contribution!

can/viewer.py
MattWoodhead: Since the introduction of #1151, I have received the following AttributeError when trying to run the viewer

```
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/pi/python-can/can/viewer.py", line 582, in <module>
    main()
  File "/home/pi/python-can/can/viewer.py", line 573, in main
    bus = _create_bus(parsed_args, **additional_config, app_name="python-can viewer")
  File "/home/pi/python-can/can/logger.py", line 105, in _create_bus
    if parsed_args.app_name:
AttributeError: 'Namespace' object has no attribute 'app_name'
```
As _create_bus from logger.py now requires the app_name argument, I thought adding this as an optional argument with a default value in Viewer.py was the simplest way to sort this without hard coding a value into the call to _create_bus on line 578. Please suggest an alternative if you dont want to do it this way as I am not deeply familiar with the library internals.

felixdivo: Hm I see. Shouldn't it also work if we used `logger._create_base_argument_parser` altogether instead of all the `optional.add_argument(...)` calls for bitrate, channel, etc.? Then, `app_name` would have already be added.

MattWoodhead: Yep, that works great. Thanks for the hint :)
I will push this in a commit.

felixdivo (approved):
Looks good to me!

PR #1163, opened 2021-11-09
Silently ignore system variables found in blf file

sander-visser:
These are observed in logs created by CANoe 10.0 SP7


sander-visser:
Maybe

PR #1167, opened 2021-11-11
Fix formatter and bump ci tools

felixdivo:
Firstly, the `black` formatter [failed recently](https://github.com/hardbyte/python-can/runs/4180378756?check_suite_focus=true#step:5:14). Bumping the version solves this problem (see [here](https://github.com/python/typed_ast/issues/169#issuecomment-964360487)).

Secondly, bumping the other CI tools made some more changes nessesary. I suppose none of them are really controversial or interesting. As the linters are somehow not executed any more, not all linter issues are fixed by this (there are quite a few that have accumulated).

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -12,7 +12,7 @@ __version__ = "4.0.0-dev.2"
 
 log = logging.getLogger("can")
 
-rc: Dict[str, Any] = dict()
+rc: Dict[str, Any] = {}
 
 from .listener import Listener, BufferedReader, RedirectReader, AsyncBufferedReader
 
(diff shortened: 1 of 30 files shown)


felixdivo:
I don't get which change caused the PCAN tests to fail ...

felixdivo:
For now, I reverted the changes to `can/interfaces/pcan/pcan.py` in facba9308d6bd775dbb7d757d710a346d7af9061. It would've caused merge conflicts with #1152 anyways.

hardbyte (approved):

felixdivo:
Thanks!

PR #1170, opened 2021-11-15
Allowing for extra config arguments in can.logger

pierreluctg:
I do not believe that interface/backend specific optional configuration should be added to `can.logger` itself.

It is actually very useful to be able to use those interface optional configuration and this change allows to use them in `can.logger`.

This change should also be backward compatible with #1142.

The other option/enhancement, is to define interface specific sub-parsers.

### DIFF
--- a/can/logger.py
+++ b/can/logger.py
@@ -61,7 +61,7 @@ def _create_base_argument_parser(parser: argparse.ArgumentParser) -> None:
         "extra_args",
         nargs=argparse.REMAINDER,
         help="""\
-        The remainding arguments will be use for the interface initialisation.
+        The remainding arguments will be used for the interface initialisation.
         For example, `-i vector -c 1 --app-name=MyCanApp` is the equivalent to 
         opening the bus with `Bus('vector', channel=1, app_name='MyCanApp')`
         """,
--- a/can/logger.py
+++ b/can/logger.py
@@ -57,10 +57,7 @@ def _create_base_argument_parser(parser: argparse.ArgumentParser) -> None:
         help="Bitrate to use for the data phase in case of CAN-FD.",
     )
 
-    parser.add_argument(
-        "--app_name",
-        help="""App name can be necessary in the initializer. For example with Vector.""",
-    )
+    parser.add_argument("extra_args", nargs=argparse.REMAINDER)
 
 
 def _append_filter_argument(
@@ -102,8 +99,6 @@ def _create_bus(parsed_args: Any, **kwargs: Any) -> can.Bus:
         config["fd"] = True
     if parsed_args.data_bitrate:
         config["data_bitrate"] = parsed_args.data_bitrate
-    if parsed_args.app_name:
-        config["app_name"] = parsed_args.app_name
 
     return Bus(parsed_args.channel, **config)  # type: ignore
 
@@ -127,6 +122,19 @@ def _parse_filters(parsed_args: Any) -> CanFilters:
             can_filters.append({"can_id": can_id, "can_mask": can_mask})
 
     return can_filters
+
+
+def _parse_additonal_config(unknown_args):
+    additional_config = dict(
+        (
+            (
+                arg.split("=", 1)[0].lstrip("--").replace("-", "_"),
+                arg.split("=", 1)[1],
+            )
+            for arg in unknown_args
+        )
+    )
+    return additional_config
 
 
 def main() -> None:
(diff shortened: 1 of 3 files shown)


felixdivo:
Yes @pierreluctg, you are right! Sadly, `test/test_scripts.py` does not test this, which would've been neat.

We should, however, also verify that it does not collide with #1159. It makes `viewer` also use `_create_base_argument_parser(parser)`.

felixdivo (approved):

pierreluctg:
> Yes @pierreluctg, you are right! Sadly, `test/test_scripts.py` does not test this, which would've been neat.
> 
> We should, however, also verify that it does not collide with #1159. It makes `viewer` also use `_create_base_argument_parser(parser)`.

I moved the remainder arg to `main` to avoid issue with other users of `_create_base_argument_parser`

felixdivo:
> I moved the remainder arg to `main`

Oh, but then `viewer.py` will break when specifying something like `--app_name`, right? I'd rather suggest to leave it in `_create_base_argument_parser()` and just move the creation of `extra_config` into `_create_bus()`, which is used by `viewer.py` too.

felixdivo:
#1159 is now merged.

pierreluctg:
I have rebased the code with #1159 and adapted `viewer` to use the additional arguments.

can/logger.py
```diff
@@ -129,6 +124,19 @@ def _parse_filters(parsed_args: Any) -> CanFilters:
     return can_filters
 
 
+def _parse_additonal_config(unknown_args):
+    additional_config = dict(
+        (
```
felixdivo, line 129: I think we can leave out these parens.

felixdivo (approved):
Nice!

felixdivo (approved):

zariiii9003:
There should be a description of this behavior when calling the help message

<details>
<summary>Console output</summary>


```text
$ python -m can.logger -h
usage: logger.py [-h] [-c CHANNEL]
                 [-i {gs_usb,seeedstudio,cantact,robotell,kvaser,udp_multicast,neousys,iscan,usb2can,serial,ixxat,systec,nixnet,vector,canalystii,virtual,nican,socketcan,neovi,pcan,slcan}
]
                 [-b BITRATE] [--fd] [--data_bitrate DATA_BITRATE] [-f LOG_FILE] [-s FILE_SIZE] [-v]
                 [--filter {<can_id>:<can_mask>,<can_id>~<can_mask>} [{<can_id>:<can_mask>,<can_id>~<can_mask>} ...]] [--active | --passive]
                 ...

Log CAN traffic, printing messages to stdout or to a given file.

positional arguments:
  extra_args

optional arguments:
  -h, --help            show this help message and exit
  -c CHANNEL, --channel CHANNEL
                        Most backend interfaces require some sort of channel. For example with the serial interface the channel might be a rfcomm device: "/dev/rfcomm0" With the
                        socketcan interfaces valid channel examples include: "can0", "vcan0"
  -i {gs_usb,seeedstudio,cantact,robotell,kvaser,udp_multicast,neousys,iscan,usb2can,serial,ixxat,systec,nixnet,vector,canalystii,virtual,nican,socketcan,neovi,pcan,slcan}, --interface {g
s_usb,seeedstudio,cantact,robotell,kvaser,udp_multicast,neousys,iscan,usb2can,serial,ixxat,systec,nixnet,vector,canalystii,virtual,nican,socketcan,neovi,pcan,slcan}
                        Specify the backend CAN interface to use. If left blank, fall back to reading from configuration files.
  -b BITRATE, --bitrate BITRATE
                        Bitrate to use for the CAN bus.
  --fd                  Activate CAN-FD support
  --data_bitrate DATA_BITRATE
                        Bitrate to use for the data phase in case of CAN-FD.
  -f LOG_FILE, --file_name LOG_FILE
                        Path and base log filename, for supported types see can.Logger.
  -s FILE_SIZE, --file_size FILE_SIZE
                        Maximum file size in bytes. Rotate log file when size threshold is reached.
  -v                    How much information do you want to see at the command line? You can add several of these e.g., -vv is DEBUG
  --filter {<can_id>:<can_mask>,<can_id>~<can_mask>} [{<can_id>:<can_mask>,<can_id>~<can_mask>} ...]
                        R|Space separated CAN filters for the given CAN interface: <can_id>:<can_mask> (matches when <received_can_id> & mask == can_id & mask) <can_id>~<can_mask>
                        (matches when <received_can_id> & mask != can_id & mask) Fx to show only frames with ID 0x100 to 0x103 and 0x200 to 0x20F: python -m can.viewer -f 100:7FC
                        200:7F0 Note that the ID and mask are always interpreted as hex values
  --active              Start the bus as active, this is applied by default.
  --passive             Start the bus as passive.
```
</details>

pierreluctg:
@zariiii9003 the help message have been added.

can/logger.py
```diff
@@ -57,7 +57,15 @@ def _create_base_argument_parser(parser: argparse.ArgumentParser) -> None:
         help="Bitrate to use for the data phase in case of CAN-FD.",
     )
 
-    parser.add_argument("extra_args", nargs=argparse.REMAINDER)
+    parser.add_argument(
+        "extra_args",
+        nargs=argparse.REMAINDER,
+        help="""\
+        The remainding arguments will be use for the interface initialisation.
```
zariiii9003, line 64:
```suggestion
        The remainding arguments will be used for the interface initialisation.
```

zariiii9003 (approved):
Looks good to me.

PR #1174, opened 2021-11-22
Calling stop_all_periodic_tasks() in BusABC

pierreluctg:
Currently, only socketcan is calling stop_all_periodic_tasks() on shutdown and respecting the fact that periodic tasks should be stopped on Bus instance shutdown (according to BusABC.send_periodic docstring).

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -411,6 +411,7 @@ class BusABC(metaclass=ABCMeta):
         Called to carry out any interface specific cleanup required
         in shutting down a bus.
         """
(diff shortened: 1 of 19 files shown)


felixdivo (changes requested):
Looks like an important change, but I suggest that now with this change, socketcan closes it's tasks via `super().shutown()`.

We should also do a quick review of the other buses to make sure they also call through `super()` in `BusABC::shutdown`.

felixdivo (approved):

felixdivo (approved):

Issue #1179, opened 2021-12-02
Test of ixxat interface got over-simplified and now fails


marcel-kanter:
The HardwareTestCase got over-simplified in commit 48b2acf and now fails. This test case must be executed with at least one adapter connected to the test system.

The test method should be reverted back.

marcel-kanter:
@hardbyte or @felixdivo assign this to me
fix is in work already and will be included in the fix of #1164 because there some tests will be added

PR #1141, opened 2021-10-18
Ixxat bus state and hardware errors detection

cowo78:
Implemented IXXATBus.state()  property.
Hardware errors are detected in _recv_internal() and raise an exception.

PR #1180, opened 2021-12-03
Adapter enumeration

marcel-kanter:
First part with the fix of #1179

The basic principle of the adapter enumeration is described in the commit message and here:

In case you have connected multiple IXXAT adapters, you have to select them by using their unique hardware id.
To get a list of all connected IXXAT adapters you can use the function ``list_adapters()`` as demonstrated below:

    >>> from can.interfaces.ixxat import IXXATBus
    >>> for hwid in IXXATBus.list_adapters():
    ...     print("Found IXXAT adapter with hardware id '%s'." % hwid)
    Found IXXAT adapter with hardware id 'HW441489'.
    Found IXXAT adapter with hardware id 'HW107422'.

The method list_adapters shall return an iterable with adapter identifiers or an empty list. If an driver must be installed, then an exception shall be thrown.

This principle should be rolled out to all other interfaces, because most of the time you deal with one type of interface but more likely with multiple adapters. To unify the interface for creating the busses, the adapter keyword argument should be used. As this does not imply that a serial number exists. For PCI cards the adapter identifier maybe the place in the PCI bus, or for network bridges this might be an arbitraty name which stands for an IP-address and port, etc.

The next step should be an method that returns the capabilities for an adapter.
Both methods then can be used to build a dict of all configurations of an interface.
The methods in each interface then can be used to build an dict of all configurations of all interfaces (recursively).

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -450,6 +450,13 @@ class BusABC(metaclass=ABCMeta):
     def fileno(self) -> int:
         raise NotImplementedError("fileno is not implemented using current CAN bus")
 
+    def list_adapters(self) -> List[Any]:
+        """ Lists all adapters for this interface. The adapter identifier can be used to open a specific adapter.
+
+        MAY NOT BE IMPLEMENTED BY ALL INTERFACES
+        """
+        raise NotImplementedError()
+
 
 class _SelfRemovingCyclicTask(CyclicSendTaskABC, ABC):
     """Removes itself from a bus.
--- a/can/interfaces/ixxat/__init__.py
+++ b/can/interfaces/ixxat/__init__.py
@@ -1,7 +1,8 @@
 """
-Ctypes wrapper module for IXXAT Virtual CAN Interface V3 on win32 systems
+Ctypes wrapper module for IXXAT Virtual CAN Interface on win32 systems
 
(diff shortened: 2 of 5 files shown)


felixdivo:
Just as a not, this will probably conflict with the much older #1119, which I'd recommend merging first. (And there's also #1141)

pierreluctg:
@marcel-kanter , how is this different from `detect_available_configs` https://github.com/hardbyte/python-can/blob/develop/can/interface.py#L128?

marcel-kanter:
> @marcel-kanter , how is this different from `detect_available_configs` https://github.com/hardbyte/python-can/blob/develop/can/interface.py#L128?

Basically this provides the steps in between and is an attempt to unify the arguments for the bus instantiation.
The list of init arguments for defining the adapter:

neovi :  serial
ixxat : unique_hardware_id
kvaser : (channel ?)
neousys : device
pcan : (channel ?)
seedstudio: channel
serial: channel
socketcan: (channel ?)
systec: device_number
udp_multicase: channel + port
usb2can: channel or serial (serial is declared legacy)
vector: app_name ?

(The ones with the question mark are a guess.)

In the end detect_available_configs can needs to enumerate the adapters and needs to find the configurations for each adapter. You can do this in one big step or provide the intermediate steps.

However _detect_available_configs should be without the leading underscore.
_detect_available_configs is not implemented for ixxat interface.

PR #1181, opened 2021-12-03
Raise ValueError if message data if over max frame length

### LINKED ISSUE #1177: ics will crash when FD message's payload size is over 64 bytes
**Describe the bug**
If we send a FD message which is larger than 64 bytes ICS will crash instead of raising an exception. 
**To Reproduce**
Steps to reproduce the behavior (please add longer code examples below):
```python
ign_data = [1]
ign_data.extend([0]*64)

frame = can.Message(
    arbitration_id=0x3B3,
    data=ign_data,
    is_fd=self.is_fd,
    is_extended_id=False
)
bus.send_periodic(frame, period,  timeout)
```

**Expected behavior**

Throw an exception for a FD message with data over limit.

**Additional context**

OS-version:
windows 10
Python version:
3.8.10
python-can version:
3.3.4
python-can interface/s:
ics_neovi

  
```python
def func():
    return 'hello, world!'
```
</details>

pierreluctg:
Fixes #1177

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -417,6 +417,18 @@ class NeoViBus(BusABC):
         """
         if not ics.validate_hobject(self.dev):
             raise CanOperationError("bus not open")
+
+        # Check for valid DLC to avoid passing extra long data to the driver
+        if msg.is_fd:
+            if msg.dlc > 64:
+                raise ValueError(
+                    f"DLC was {msg.dlc} but it should be <= 64 for CAN FD frames"
+                )
+        elif msg.dlc > 8:
+            raise ValueError(
+                f"DLC was {msg.dlc} but it should be <= 8 for normal CAN frames"
+            )
+
         message = ics.SpyMessage()
 
         flag0 = 0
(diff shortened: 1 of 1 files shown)


felixdivo:
Thanks @pierreluctg. Would you agree that raising a `ValueError` would be more appropriate? I think it is better to be explicit to the user here, since there probably is no reasonable use of the send method with such invalid data. It would also be (in my opinion, at least) closer to the docstring.

hardbyte (approved):

felixdivo (changes requested):
See above.

pierreluctg:
> Thanks @pierreluctg. Would you agree that raising a `ValueError` would be more appropriate? I think it is better to be explicit to the user here, since there probably is no reasonable use of the send method with such invalid data. It would also be (in my opinion, at least) closer to the docstring.

I agree with raising `ValueError`, however, I do not think this should be interface specific. This should be done by the `Message` class. Unless there is a valid use case for a `Message` with data exceeding 64 bytes?

felixdivo:
Hm, I think we did not verify the messages objects due to (untested?) performance concerns (?).

There is this parameter to the message class: https://github.com/hardbyte/python-can/blob/develop/can/message.py#L65

pierreluctg:
@felixdivo is it fair to ask the `Message` object creator to validate the data (using the `check`) if needed and simply truncate the data here to avoid sending invalid data to the driver layer?

felixdivo (approved):

PR #1188, opened 2021-12-13
Implement check for minimum version of pcan library

pkess:
Hello there,

while using this library for a long time now i experienced some issues in the past that the initialization was not working. I checked and found out that there was a parameter set during initialization:
```python
        result = self.m_objPCANBasic.SetValue(
            self.m_PcanHandle, PCAN_ALLOW_ERROR_FRAMES, PCAN_PARAMETER_ON
        )
```
Checking documentation from peak, this parameter is only available with version `4.2` or higher. My version was `4.1.0.96`

To make it easier for others to solve this issue i implemented a version check here.

Open questions:
1.  Should there be a parameter to avoid version check or similar?
2.  Should it be possible to avoid set of this parameter to make this work with older versions?

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -9,6 +9,8 @@ import platform
 
 from typing import Optional
 
+from packaging import version
+
 from ...message import Message
 from ...bus import BusABC, BusState
 from ...util import len2dlc, dlc2len
@@ -19,6 +21,7 @@ from .basic import (
     PCAN_BITRATES,
     PCAN_FD_PARAMETER_LIST,
     PCAN_CHANNEL_NAMES,
+    PCAN_NONEBUS,
     PCAN_BAUD_500K,
     PCAN_TYPE_ISA,
     PCANBasic,
@@ -26,6 +29,7 @@ from .basic import (
     PCAN_ALLOW_ERROR_FRAMES,
     PCAN_PARAMETER_ON,
     PCAN_RECEIVE_EVENT,
+    PCAN_API_VERSION,
     PCAN_DEVICE_NUMBER,
     PCAN_ERROR_QRCVEMPTY,
     PCAN_ERROR_BUSLIGHT,
@@ -57,6 +61,8 @@ from .basic import (
 
 # Set up logging
 log = logging.getLogger("can.pcan")
+
+MIN_PCAN_API_VERSION = version.parse("4.2.0")
 
 
 try:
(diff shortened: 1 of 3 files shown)


pkess:
Hello, it looks like I have to fix something here. But can you give some feedback on the basic concept here and the open questions?

pkess:
Well, the real unit test cases succeed now. But i don't know why code coverage is shown to be decreased. At least i added test cases for the new lines introduced here.

can/interfaces/pcan/pcan.py
```diff
@@ -304,6 +312,21 @@ def bits(n):
...
+
+        return version.parse(value.decode("ascii"))
+
+    def check_api_version(self):
+        apv = self.get_api_version()
+        if apv < MIN_PCAN_API_VERSION:
+            raise CanInitializationError(
+                f"Minimum version of pcan api is {MIN_PCAN_API_VERSION}."
+                f" Installed version is {apv}. Consider upgrade of pcan basic package"
+            )
```
hardbyte, line 328: We could also log a warning and try to continue anyway. Do you know more about what is imposing the current version requirement? Will every user always need that same minimum or is it only for some functionality?

Overall though, this seems like a good approach to me.

hardbyte (approved):

can/interfaces/pcan/pcan.py
pkess: Ok, a warning might be sufficient here. I will change this.
There is a document inside the PCAN Basic package: https://www.peak-system.com/fileadmin/media/files/pcan-basic.zip
Here it is described what version of the API is required for several functions to work. For the Parameter `PCAN_ALLOW_ ERROR_FRAMES` the version `4.2.0` or higher is required so i made this the minimum api version. I did not check if there were other parameters used with higer minimum version. It could be a good idea to attach the minimum version to every parameter so that it could be checked, but i think this is to much for this pr at the moment.

test/test_pcan.py
```diff
@@ -34,6 +37,17 @@ def tearDown(self) -> None:
...
+    def _mockGetValue(self, Channel, Parameter):
+        """
+        This method is used as mock for GetValue method of PCANBasic object.
+        Only a subset of parameters are supported.
+        """
+        if Parameter == PCAN_API_VERSION:
+            return PCAN_ERROR_OK, self.PCAN_API_VERSION_SIM.encode("ascii")
+        raise NotImplementedError(
+            f"No mock return value specified for parameter {Parameter}"
+        )
```
felixdivo, line 49: It's convention to use lowercase variables. :)

```suggestion
    def _mockGetValue(self, channel, parameter):
        """
        This method is used as mock for GetValue method of PCANBasic object.
        Only a subset of parameters are supported.
        """
        if parameter == PCAN_API_VERSION:
            return PCAN_ERROR_OK, self.PCAN_API_VERSION_SIM.encode("ascii")
        raise NotImplementedError(
            f"No mock return value specified for parameter {parameter}"
        )
```

felixdivo (approved):

PR #1195, opened 2021-12-17
Add preserve timestamps to virtual

simontegelid:
Add an option to virtual interfaces to preserve message timestamps on transmissions. This is useful in test setups fed by log files.


felixdivo (approved):
Looks useful!

pierreluctg (approved):

Issue #585, opened 2019-05-16
Test logger.py


hardbyte:
See coverage of 40% at release 3.2 - https://codecov.io/gh/hardbyte/python-can/src/7268f49f4ca349464e29bc680e9e42c66e313455/can/logger.py

Ideally this is >90%

felixdivo:
#1036 Factors our common functionality and introduces slightly better coverage.

pkess:
I think this can be closed as converage is 100% now. If you see more tests do be added here i would like to implement this.

felixdivo:
Hey @pkess I've just now seen the **_great_** PR #1196. Sure, you are right!

PR #1196, opened 2021-12-19
Add unittest for logger execution

pkess:
According to issue #585 i started implementing some more unit test for the logger.py module. This is only a preview at the moment. I would like to check code coverage to get at least above 90 percent


pkess:
Finally it seems like the tests are successfull now and logger.py is covered 100%. I hope you like this

hardbyte:
Thanks for the contribution @pkess - much appreciated!

Issue #584, opened 2019-05-16
Test player.py


hardbyte:
See coverage of ~50% at release 3.2 - https://codecov.io/gh/hardbyte/python-can/src/release-3.2.0/can/player.py

Ideally this is >90% covered by a CI executed test on each supported version of Python.

Ralph-Ne:
Hi hardbyte,

I've got some problems with the newest version of the player in python-can.

The call is the same: python -m can.player -b 250000 -c can0 -v -i socketcan FILE_TO_PLAY

If I use version 2.2.1dev0, downloaded over a year ago, the player works fine, even without the -g option.

If I use the newest version 3.3.1, the player ends with:

Timestamp: 1504177560.972875        ID: 0000    S E DLC:  0                                Channel: 0
Timestamp: 1504177560.973315        ID: 0000    S E DLC:  0                                Channel: 0
Traceback (most recent call last):
  File "/usr/lib/python3.7/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.7/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/lib/python3.7/site-packages/can/player.py", line 98, in <module>
    main()
  File "/usr/lib/python3.7/site-packages/can/player.py", line 89, in main
    bus.send(m)
  File "/usr/lib/python3.7/site-packages/can/interfaces/socketcan/socketcan.py", line 633, in send
    raise can.CanError("Transmit buffer full")
can.CanError: Transmit buffer full
[canplayer@canplayer-pc ~]$

If I use the -g option it works, if I do -g 0.01: python -m can.player -b 250000 -c can0 -v -i socketcan -g 0.01 FILE_TO_PLAY,

-g 0.001 or -g 0.0001 lead to the same error as above.

Can you imagine what I've done wrong? Do you have any advice for me to solve the problem?


Kind regards

Ralph

karlding:
@Ralph-Ne can you open a new issue with the problem you're describing, and we'll triage from there? This issue is tracking some code coverage improvements.

Thanks!

felixdivo:
As discussed on the mailing list, we will mainly release what is there on release day. Only large issues are coordinated. Therefore, this issue is removed from the milestone.

pkess:
I think with #1198 this issue can be closed now. If you want additional tests just let me know

PR #1198, opened 2021-12-20
584 add test player

pkess:
This implements test for the issue #584. It is not finished yet but i will continue working on this

### DIFF
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -26,7 +26,9 @@ class TestPlayerScriptModule(unittest.TestCase):
         self.mock_virtual_bus.shutdown = Mock()
 
         self.baseargs = [sys.argv[0], "-i", "virtual"]
-        self.logfile = os.path.join(os.path.dirname(__file__), "data", "test_CanMessage.asc")
+        self.logfile = os.path.join(
+            os.path.dirname(__file__), "data", "test_CanMessage.asc"
+        )
 
     def assertSuccessfullCleanup(self):
         self.MockVirtualBus.assert_called_once()
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -64,8 +64,12 @@ class TestPlayerScriptModule(unittest.TestCase):
         self.assertEqual(self.MockSleep.call_count, 2)
         if sys.version_info.major > 3 or sys.version_info.minor >= 8:
             # The args argument was introduced with python 3.8
-            self.assertTrue(msg1.equals(self.mock_virtual_bus.send.mock_calls[0].args[0]))
-            self.assertTrue(msg2.equals(self.mock_virtual_bus.send.mock_calls[1].args[0]))
+            self.assertTrue(
+                msg1.equals(self.mock_virtual_bus.send.mock_calls[0].args[0])
+            )
+            self.assertTrue(
+                msg2.equals(self.mock_virtual_bus.send.mock_calls[1].args[0])
+            )
         self.assertSuccessfulCleanup()
 
     def test_play_virtual_verbose(self):
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -23,7 +23,7 @@ class TestPlayerScriptModule(unittest.TestCase):
         self.MockVirtualBus = patcher_virtual_bus.start()
         self.addCleanup(patcher_virtual_bus.stop)
         self.mock_virtual_bus = self.MockVirtualBus.return_value
-        self.mock_virtual_bus.shutdown = Mock()
+        self.mock_virtual_bus.__enter__ = Mock(return_value=self.mock_virtual_bus)
 
         # Patch time sleep object
         patcher_sleep = mock.patch("can.io.player.sleep", spec=True)
@@ -37,11 +37,12 @@ class TestPlayerScriptModule(unittest.TestCase):
 
     def assertSuccessfullCleanup(self):
         self.MockVirtualBus.assert_called_once()
+        self.mock_virtual_bus.__exit__.assert_called_once()
 
     def test_play_virtual(self):
         sys.argv = self.baseargs + [self.logfile]
         can.player.main()
-        # TODO: add test two messages sent
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 2)
         self.assertEqual(self.MockSleep.call_count, 2)
         self.assertSuccessfullCleanup()
 
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -15,7 +15,6 @@ import can
 import can.player
 
 
-
 class TestPlayerScriptModule(unittest.TestCase):
     def setUp(self) -> None:
         # Patch VirtualBus object
--- a/test/test_player.py
+++ b/test/test_player.py
@@ -41,8 +41,29 @@ class TestPlayerScriptModule(unittest.TestCase):
     def test_play_virtual(self):
         sys.argv = self.baseargs + [self.logfile]
         can.player.main()
-        self.assertEqual(self.mock_virtual_bus.send.call_count, 2)
+        msg1 = can.Message(
+            timestamp=2.501,
+            arbitration_id=0xC8,
+            is_extended_id=False,
+            is_fd=False,
+            is_rx=False,
+            channel=1,
+            dlc=8,
+            data=[0x9, 0x8, 0x7, 0x6, 0x5, 0x4, 0x3, 0x2],
+        )
+        msg2 = can.Message(
+            timestamp=17.876708,
+            arbitration_id=0x6F9,
+            is_extended_id=False,
+            is_fd=False,
+            is_rx=True,
+            channel=0,
+            dlc=8,
+            data=[0x5, 0xC, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0],
+        )
         self.assertEqual(self.MockSleep.call_count, 2)
+        self.assertTrue(msg1.equals(self.mock_virtual_bus.send.mock_calls[0].args[0]))
+        self.assertTrue(msg2.equals(self.mock_virtual_bus.send.mock_calls[1].args[0]))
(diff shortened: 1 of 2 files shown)


pkess:
I think this is ready for review and merge now

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+        patcher_sleep = mock.patch("can.io.player.sleep", spec=True)
+        self.MockSleep = patcher_sleep.start()
+        self.addCleanup(patcher_sleep.stop)
+
+        self.baseargs = [sys.argv[0], "-i", "virtual"]
+        self.logfile = os.path.join(
+            os.path.dirname(__file__), "data", "test_CanMessage.asc"
+        )
+
+    def assertSuccessfullCleanup(self):
```
zariiii9003, line 39:
```suggestion
    def assertSuccessfulCleanup(self):
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+import unittest
+from unittest import mock
+from unittest.mock import Mock
+import os
+import sys
+import io
+import can
+import can.player
+
+from .config import *
```
zariiii9003, line 17:
```suggestion
```

This is unnecessary

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+    def assertSuccessfullCleanup(self):
+        self.MockVirtualBus.assert_called_once()
+        self.mock_virtual_bus.__exit__.assert_called_once()
+
+    def test_play_virtual(self):
+        sys.argv = self.baseargs + [self.logfile]
+        can.player.main()
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 2)
+        self.assertEqual(self.MockSleep.call_count, 2)
+        self.assertSuccessfullCleanup()
```
zariiii9003, line 48:
```suggestion
        self.assertSuccessfulCleanup()
```

test/test_player.py
zariiii9003, line 39:
```suggestion
    def assertSuccessfulCleanup(self):
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_play_virtual_verbose(self):
+        sys.argv = self.baseargs + ["-v", self.logfile]
+        with unittest.mock.patch("sys.stdout", new_callable=io.StringIO) as mock_stdout:
+            can.player.main()
+        self.assertIn("09 08 07 06 05 04 03 02", mock_stdout.getvalue())
+        self.assertIn("05 0c 00 00 00 00 00 00", mock_stdout.getvalue())
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 2)
+        self.assertEqual(self.MockSleep.call_count, 2)
+        self.assertSuccessfullCleanup()
```
zariiii9003, line 58:
```suggestion
        self.assertSuccessfulCleanup()
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_play_virtual_exit(self):
+        self.MockSleep.side_effect = [None, KeyboardInterrupt]
+
+        sys.argv = self.baseargs + [self.logfile]
+        can.player.main()
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 1)
+        self.assertEqual(self.MockSleep.call_count, 2)
+        self.assertSuccessfullCleanup()
```
zariiii9003, line 67:
```suggestion
        self.assertSuccessfulCleanup()
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_play_skip_error_frame(self):
+        logfile = os.path.join(
+            os.path.dirname(__file__), "data", "logfile_errorframes.asc"
+        )
+        sys.argv = self.baseargs + ["-v", logfile]
+        can.player.main()
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 9)
+        self.assertEqual(self.MockSleep.call_count, 12)
+        self.assertSuccessfullCleanup()
```
zariiii9003, line 77:
```suggestion
        self.assertSuccessfulCleanup()
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def test_play_error_frame(self):
+        logfile = os.path.join(
+            os.path.dirname(__file__), "data", "logfile_errorframes.asc"
+        )
+        sys.argv = self.baseargs + ["-v", "--error-frames", logfile]
+        can.player.main()
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 12)
+        self.assertEqual(self.MockSleep.call_count, 12)
+        self.assertSuccessfullCleanup()
```
zariiii9003, line 87:
```suggestion
        self.assertSuccessfulCleanup()
```

test/test_player.py
```diff
@@ -0,0 +1,91 @@
...
+        )
+
+    def assertSuccessfullCleanup(self):
+        self.MockVirtualBus.assert_called_once()
+        self.mock_virtual_bus.__exit__.assert_called_once()
+
+    def test_play_virtual(self):
+        sys.argv = self.baseargs + [self.logfile]
+        can.player.main()
+        self.assertEqual(self.mock_virtual_bus.send.call_count, 2)
```
zariiii9003, line 46:
```suggestion
        msg1 = can.Message(
            timestamp=2.501,
            arbitration_id=0xC8,
            is_extended_id=False,
            is_fd=False,
            is_rx=False,
            channel=1,
            dlc=8,
            data=[0x9, 0x8, 0x7, 0x6, 0x5, 0x4, 0x3, 0x2],
        )
        msg2 = can.Message(
            timestamp=17.876708,
            arbitration_id=0x6F9,
            is_extended_id=False,
            is_fd=False,
            is_rx=True,
            channel=0,
            dlc=8,
            data=[0x5, 0xC, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0],
        )
        self.mock_virtual_bus.send.assert_has_calls([mock.call(msg1), mock.call(msg2)])
```

zariiii9003 (changes requested):
Hello Peter, thank you for your contribution and sorry for the late reply! Tests are always welcome.
I think it makes sense to check the call arguments of the mock. Otherwise this looks good to me, just some minor cosmetic issues.

pkess:
Thank you for your review. i changed the code now accordingly

pkess:
Unfortunatelly the compare operator was not implemented for messages. I implemented this now. But do you want this to be added with this pr?

zariiii9003:
> Unfortunatelly the compare operator was not implemented for messages. I implemented this now. But do you want this to be added with this pr?

There is a `Message.equals()` method for this purpose. I'd say remove your last commit since that is out of scope for this PR.

pkess:
> > Unfortunatelly the compare operator was not implemented for messages. I implemented this now. But do you want this to be added with this pr?
> 
> There is a `Message.equals()` method for this purpose. I'd say remove your last commit since that is out of scope for this PR.

I knew about the `equals` function and i used it in the compare operator. But i think it was necessary to get your proposal with `assert_has_calls` working. What should i do with this? should i implement an own compare for the executed calls?

zariiii9003:
> I knew about the `equals` function and i used it in the compare operator. But i think it was necessary to get your proposal with `assert_has_calls` working. What should i do with this? should i implement an own compare for the executed calls?

I wonder how i did not notice that, sorry. Nevertheless,  there was a discussion in #413 and `__eq__` was removed.

We could this instead:
```python3
        self.assertTrue(self.mock_virtual_bus.send.mock_calls[0].args[0].equals(msg1))
        self.assertTrue(self.mock_virtual_bus.send.mock_calls[1].args[0].equals(msg2))
```
It's ugly but it does the job.

pkess:
> > I knew about the `equals` function and i used it in the compare operator. But i think it was necessary to get your proposal with `assert_has_calls` working. What should i do with this? should i implement an own compare for the executed calls?
> 
> I wonder how i did not notice that, sorry. Nevertheless, there was a discussion in #413 and `__eq__` was removed.
> 
> We could this instead:
> 
> ```python
>         self.assertTrue(self.mock_virtual_bus.send.mock_calls[0].args[0].equals(msg1))
>         self.assertTrue(self.mock_virtual_bus.send.mock_calls[1].args[0].equals(msg2))
> ```
> 
> It's ugly but it does the job.

Ok, i implemented it this way now. I used the created message object for the `equals` function to be sure there will be no error method does not exist when the called argument is not correct.

pkess:
Unfortunately some tests fail now. i will try to fix this

pkess:
> Unfortunately some tests fail now. i will try to fix this

It looks like the problem is that the args property was introduced with python3.8. So i only compare this in higher versions now. I think this is ok as the behaviour should not change with python version

zariiii9003 (approved):

test/test_player.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+# coding: utf-8
```
felixdivo, line 2: I'd suggest these changes for consistency.

```suggestion
#!/usr/bin/env python
```

felixdivo (approved):
This is a really nice PR! Thanks for tackling this "technical debt" and making sure the code actually works in unit tests! And most importantly, these serve as great regression tests for future changes.

Do you agree that this resolves #584?

test/test_player.py
pkess: I applied your suggestion. But it might be a good idea to keep at least all test files to the same standard header. Should i prepare a pr for this?

felixdivo: If you want, you can do that! 🥳

I already tried such cleanup efforts before but over time, inconsistencies keep coming back if they are not enforced by CI tools.

felixdivo: E.g. there was #1075

Issue #1175, opened 2021-11-26
python-can 4.0.x can/util.py timing configuration do not allow int


nedlukasz:
tseg1, tseg2, sjw cannot be configured as int.

Issue is located in file util.py in function _create_bus_config in line 235:
`timing_conf[key] = int(config[key], base=0)`

For example:
```python
config["tseg1"] = int(5)
timing_conf["tseg1"] = int(config["tseg1"], base=0)  #  here will be exception
#  TypeError: int() can't convert non-string with explicit base
```

**Expected behavior**
For config[key] specified as string use int(config[key], base=0)
For config[key] specified as not string use int(config[key])

Tbruno25:
I'll take this issue!

PR #1204, opened 2022-01-01
Add str conversion to config vals

Tbruno25:
Closes out issue #1175


zariiii9003:
Thank you for the fix! Could you add a test that fails without your fix?

Tbruno25:
@zariiii9003 added!

felixdivo (approved):
Nice!

zariiii9003 (approved):

Issue #859, opened 2020-06-18
Filter documentation improvements


hardbyte:
Needs a bit more explanation and examples.

Links:
- http://www.cse.dmu.ac.uk/~eg/tele/CanbusIDandMask.html

felixdivo:
Doesn't [this](https://python-can.readthedocs.io/en/develop/bus.html#can.BusABC.set_filters) suffice?

hardbyte:
It is certainly better than it was but still no, not really. It serves as a good reference for those familiar with the library. If I recall someone asked for some examples of using can filters - including what happens to example messages.


More generally python-can could really do with a few [documentation guides](https://docs-guide.readthedocs.io/en/latest/doctypes/) to compliment the example code and reference docs.

PR #1208, opened 2022-01-11
Add utility function create_filter

### LINKED ISSUE #1190: Assigning Can Filters for mcp2515 CAN bus module on python
so I am using the following method to filter messages between 2 controllers,but the issue with this method is that I can only filter up to 5     
can_ids, so I was wondering if there is a better method to do that, knowing I am quite new to this and to python even.
bus.set_filters([{"can_id": 0x591, "can_mask": 0xFFFFFFF, "extended": False},\
    {"can_id": 0x592"can_mask": 0xFFFFFFF, "extended": False},\
    {"can_id": 0x593, "can_mask": 0xFFFFFFF, "extended": False},\
    {"can_id": 0x594, "can_mask": 0xFFFFFFF, "extended": False},\
    {"can_id": 0x595, "can_mask": 0xFFFFFFF, "extended": False}])

zariiii9003:
Closes #1190 
Also somewhat relevant to #859

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -31,7 +31,7 @@ from .io import CanutilsLogReader, CanutilsLogWriter
 from .io import CSVWriter, CSVReader
 from .io import SqliteWriter, SqliteReader
 
-from .util import set_logging_level
+from .util import set_logging_level, create_filter
 
 from .message import Message
 from .bus import BusABC, BusState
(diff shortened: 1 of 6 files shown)


felixdivo:
What filter does `create_filter` compute? One that (with two rules) covers at least the given IDs (and some others maybe too)? I think this can be stated in the docstring more clearly.

felixdivo:
See also  #1225

PR #1209, opened 2022-01-12
Add ZLG interface

keelung-yang:


felixdivo:
Hey @keelung-yang, I did see your proposal (🥳!) but I currently do not have time to review it. Sorry for that 😒.

keelung-yang:
@felixdivo Thanks. It's OK for waiting and improving my code in actual usage before merging.  
And please note that I've one line code (-> list[int] in timing module) using Python 3.9 and several lines using Python 3.8. I want to wait for requiring Python 3.8 as minimal version, before someone wants to use this ZLG interface.

zhuyu4839:
> Hey @keelung-yang, I did see your proposal (🥳!) but I currently do not have time to review it. Sorry for that 😒.

ZLG can interface has added by #1351

hardbyte:
@keelung-yang would you mind reviewing https://github.com/hardbyte/python-can/pull/1351 and comparing?

keelung-yang:
> @keelung-yang would you mind reviewing #1351 and comparing?

Hi @hardbyte & @zhuyu4839 
Seems zhuyu's implement is depended on his zlgcan-driver project: https://github.com/zhuyu4839/zlgcan-driver, which trying to  wrap different APIs and platforms(Linux/Windows) for ZLG CAN libraries.

ZLG should take response for universal APIs to access their various CAN devices, like others, such as kvaser and so on. But they don't want, according to the communications with their technical assistants and developer. Even for the same device, such as USBCAN(FD),  there are 3 different bundles of API just for CAN or CANFD, Linux or Windows. That's why zhuyu's API wrapping is very long, almost 1500 lines: https://github.com/zhuyu4839/zlgcan-driver/blob/main/zlgcan/zlgcan.py

ZLG tried to add their offical python-can interface implement before releasing python-can v4.0. But they gave up: https://manual.zlg.cn/web/#/169/6070

For my implement, I don't want to be disarray by ZLG's native libraries. And on the other hand, on Windows, ZLG has official GUI app and development libraries which reqiring VC runtime 2005、2008、2010、2012、2013(https://manual.zlg.cn/web/#/152/5332).  So, let's goto Linux since I have to use them in my daily work, and share if others want too.

In short, ZLG APIs are in chaos, users should take care of oneself. 

For python-can, it's hard to implement universal APIs to support almost all ZLG CAN devices, both in CAN and CANFD, Linux and Windows. Only when ZLG want to do that.

-----------------------------------------
P.S. for zlgcan-driver project, containing .so/.dll from ZLG is not a good idea since they should be published by the vendor ZLG.

zhuyu4839:
I have separated the code for different platforms.

FroyoRab:
what's up with checks? I can't see details.
it say "The logs for this run have expired and are no longer available."
can do it again? hope to help merge.

zariiii9003:
> what's up with checks? I can't see details. it say "The logs for this run have expired and are no longer available." can do it again? hope to help merge.

I reran the tests. I'd prefer if this was implemented using the [plugin interface](https://python-can.readthedocs.io/en/develop/plugin-interface.html)

keelung-yang:
I see this log in some checks:
```
FAILED test/test_interface_zlg.py::ZlgTestCase::testRecv - can.exceptions.CanInterfaceNotImplementedError: 
Cannot import module can.interfaces.zlg for CAN interface 'zlg': libusbcanfd.so: cannot open shared object file: No such file or directory
```
That's because I do have no any .so file uploaded into repository.
Because it will be out of date if ZLG update `libusbcanfd.so` officially.
And on the other hand, this may cause some security issue I guess, since I don't know how the  `libusbcanfd.so`  is generated.
So I want users to copied it from ZLG official site instead of using the one uploaded by me.
And I've make this clear in the doc/interfaces/zlg.rst

zariiii9003:
There is python-can plugin, which adds zlg support: https://github.com/jesses2025smith/zlgcan-driver, closing.

PR #1212, opened 2022-01-14
Change DLC to DL in Message.__str__()

### LINKED ISSUE #1183: Current DLC didn't support FD
### Describe the bug

Technically DLC is only 4 bits long. DLC values ranging from 1001 to 1111 are used to specify the data lengths of 12, 16, 20, 24, 32, 48, and 64 bytes.  Currently Message.DLC is actually the data length instead of DLC. It is ok for classic CAN but for CAN FD it will cause the DLC value up to 64.

Timestamp:    30969.422452        ID: 03b3    S     F BS       DLC: 16    44 00 00 0c e6 00 00 00 00 00 00 00 00 00 00 00

zariiii9003:
Closes #1183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,6 +138,7 @@ Other API changes and improvements
 * Speed up interface plugin imports by removing pkg_resources (#1110)
 * Allowing for extra config arguments in can.logger (#1142, #1170)
 * Add changed byte highlighting to viewer.py (#1159)
+* Change DLC to DL in Message.\_\_str\_\_() (#1212)
 
 Other Bugfixes
 --------------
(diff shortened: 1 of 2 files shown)


pierreluctg (approved):

felixdivo:
Shouldn't we also think about the `dlc` attribute before changing this in only part of the code?

zariiii9003:
> Shouldn't we also think about the `dlc` attribute before changing this in only part of the code?

That would mean an API change and a deprecation period for the `Message.dlc` attrbute. i don't really think there's a need for it. This PR is just a cosmetic change.

felixdivo:
Yeah, I just thought that it might cause more confusion than it helps, but I don't have a strong opinion on this one.

felixdivo (approved):

felixdivo:
Okay, 4.0.0 is a good point in time for this 👍

PR #1213, opened 2022-01-14
Test python 3.10 on Windows

zariiii9003:
windows-curses is available for python 3.10 now, so we can include 3.10 for all operating systems.


felixdivo:
Oh, we had the same idea. I didn't see your PR in time, though and implemented pretty much the same over here: #1214.

zariiii9003:
Closing in favor of #1214

Issue #528, opened 2019-03-05
What to do in 2020 - Python 2 EOL


felixdivo:
> Being the last of the 2.x series, 2.7 will have an extended period of maintenance. Specifically, 2.7 will receive bugfix support until January 1, 2020. After the last release, 2.7 will receive no support.

It's no secret that [Python 2 is reaching its end of life](https://pythonclock.org/).

What do we do from then on? Should we just document which version was the last to support Python 2 + 3 and then do a (possibly incremental) cleanup of the code and only support Python 3 from then on? We could do that in a major version bump (like maybe 4.0).

I would vote for exactly that. Do we all agree on that?

felixdivo:
cc @hardbyte @christiansandberg

hardbyte:
Yeah I'm in favor of making a python-can v3.2 release be the final release supporting python2 (with possible bug fixes accepted to a 3.x up until python 2 EOL), that frees us up to do a major version bump to v4 where we could drop python2 support and possibly introduce type annotations throughout the core.

felixdivo:
Adding typing sounds great indeed! This works especially well because [Python 3.4 will reach EOL 2019-03-16](https://devguide.python.org/#status-of-python-branches). So we would support Python 3.5+ only, which enables us to freely use the `typing` module.

But we cannot use the "new" typing syntax yet, since it is [only supported by Python 3.6+](https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#type-hints-cheat-sheet-python-3), which will be [supported until 2020-09-13](https://devguide.python.org/#status-of-python-branches).

danielhrisca:
Why not jump straight to Python >= 3.6?

felixdivo:
Because Python 3.5 is supported until `2020-09-13`.

hardbyte:
That doesn't mean we have to support Python3.5 in python-can v4. Any Python3.5 users can always use python-can<=3.2.

I work on another [codebase](https://github.com/data61/clkhash/blob/master/clkhash/bloomfilter.py#L146) using comment based typing and it is a bit of a pain so I'd prefer to make the jump to the new typing syntax.

danielhrisca:
I think Python 3.6 brings some compelling features: f-strings, ordered dicts with better performance, annotations, improved asyncio

felixdivo:
Okay, you two convinced me. The users on "archaic" systems may still stay on 2.3.x 👌

felixdivo:
[...]

PR #1214, opened 2022-01-15
Update workflow versions, platforms, library and tool versions + README

felixdivo:
I think most changes do not require explanation. The README was changed since Python 3.6 reached EOL as discussed in (the end of the discussion of) #528.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -44,7 +44,6 @@ setup(
     classifiers=[
         # a list of all available ones: https://pypi.org/classifiers/
         "Programming Language :: Python",
-        "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -73,7 +72,7 @@
     version=version,
     packages=find_packages(exclude=["test*", "doc", "scripts", "examples"]),
     scripts=list(filter(isfile, (join("scripts/", f) for f in listdir("scripts/")))),
-    author="Python CAN contributors",
+    author="python-can contributors",
```
felixdivo, line 75: I renamed this since it's more familiar for most, I suppose?

hardbyte (approved):

felixdivo:
@zariiii9003 Do you approve?

zariiii9003 (approved):
> @zariiii9003 Do you approve?

Yeah i was just a little hesitant to drop 3.6 already.

felixdivo:
> Yeah i was just a little hesitant to drop 3.6 already.

I think there is no point in maintaining an EOL version. People can still install it at their own risk. In the worst case, they have to stick to a certain commit. 🤷

PR #1221, opened 2022-01-18
1210 Add compressed file methods

### LINKED ISSUE #1210: can.LogReader error on opening "filename.with.too.many.dots.blf"
### Describe the bug
in python3 repl
```python
>> can.LogReader("./2021.11.23.test.blf") 
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/username/Documents/various/python-can/can/io/player.py", line 87, in __new__
    raise ValueError(
ValueError: No read support for this unknown log format ".11.23.test.blf"
```
### To Reproduce
Open a log file with more then one period in it.

### Expected behavior
The file should be opened as a blf file
If the file is renamed to "test.blf" the conversion works properly

### Additional context

OS and version: Ubuntu20.04 in WSL `Linux TL0213507 4.4.0-19041-Microsoft #1237-Microsoft Sat Sep 11 14:32:00 PST 2021 x86_64 x86_64 x86_64 GNU/Linux`
Python version: python 3.8.10
python-can version: python develop commit 2e24af08326ecd69fba9f02fed7b9c26f233c92b

Tbruno25:
This closes #1210 by adding `LogReader.decompress` / `Logger.compress` methods.

Changes:
- ~~add `io/gz.py`~~
- ~~add `GzipReader` class~~
- ~~add `GzipReader` to `LogReader` message readers~~
- edit `LogReader` to check for last suffix only
- edit `LogReader` docstring 
- add `decompress` method to `LogReader`
- add unittest for compressed player
- edit `Logger` to check for last suffix only
- edit `Logger` docstring 
- add `compress` method to `Logger`
- add unittest for compressed logger
- add multiple dot suffix to `test_extension_matching`
- remove `GzipASCReader` / `GzipASCWriter` and subsequent unittests


can/io/gz.py
```diff
@@ -0,0 +1,37 @@
...
+        real_suffix = pathlib.Path(file).suffixes[-2]
+        file_mode = "b" if real_suffix in [".blf"] else "t"
+
+        with tempfile.NamedTemporaryFile(
+            mode=f"w{file_mode}", suffix=real_suffix
+        ) as log:
+            with gzip.open(file, f"r{file_mode}") as f:
+                log.write(f.read())
+
+            return LogReader(log.name)
```
felixdivo, line 37: I think that after this return, the temporary file will get closed by *might* be removed.

Instead, we could make this class inherit from `MessageReader`, which inherits from `BaseIOHandler` which has a `stop()` method in which we can close resources. But then, we would have to abandon the `__new__` approach.

can/io/gz.py
```diff
@@ -0,0 +1,37 @@
...
+        :param file: a path-like object or as file-like object to read from.
+        """
+        # Needed to avoid circular import error
+        from .player import LogReader
+
+        real_suffix = pathlib.Path(file).suffixes[-2]
+        file_mode = "b" if real_suffix in [".blf"] else "t"
+
+        with tempfile.NamedTemporaryFile(
+            mode=f"w{file_mode}", suffix=real_suffix
```
felixdivo, line 32: I think for the copy procedure, we do not need to decode and re-encode the file, we can stay in binary `file_mode`, right?

can/io/gz.py
felixdivo: It would be nice to be able to pass a file name *and* a an optional file-pointer to `LogReader`, one to resolve the reader type and one for the actual content. That way, we could read the file "lazily" by directly passing the `gzip.open()` result and avoid the temporary file. It would even get closed automatically by the specific reader.

can/io/gz.py
Tbruno25: you're right!

I was experimenting with using the `io` module which required I select between `StringIO` or `BytesIO` -- if a `tempfile` is used it's not needed since we don't need to read the file and can use something like `shutil` to copy.

can/io/gz.py
Tbruno25: I like this idea as well and it seems the most logical. 

However if we're going to alter `LogReader`, I  think it makes more sense for `GzipReader` to become a method within `LogReader` since technically it's not _reading_ anything. This would also avoid the circular import issue which seems like a bit of a 'code  smell' imo.

felixdivo: That sounds pretty simple now that you say it!

Tbruno25: Added! Would love feedback before I work on the tests.

can/io/player.py
```diff
@@ -88,6 +92,23 @@ def __new__(  # type: ignore
...
+    def unzip(zipfile: "can.typechecking.StringPathLike"):
+        """
+        Return the suffix and io object of the decompressed file.
+        """
+        real_suffix = pathlib.Path(zipfile).suffixes[-2].lower()
+        file = gzip.open(zipfile, "rt")
+
+        # Re-open in binary mode if file not readable.
+        try:
+            file.read()
```
felixdivo, line 105: > When size is omitted or negative, the entire contents of the file will be read and returned; it’s your problem if the file is twice as large as your machine’s memory.

This could be problematic, right?

can/io/player.py
```diff
@@ -88,6 +92,23 @@ def __new__(  # type: ignore
                 f'No read support for this unknown log format "{suffix}"'
             ) from None
 
+    @staticmethod
+    def unzip(zipfile: "can.typechecking.StringPathLike"):
```
felixdivo, line 96: I'd suggest to rename this to `decompress()`, since it's not about `.zip` files and we could easily extend this to other compression types if needed.

can/io/player.py
felixdivo: Do we need a `mode` argument or shall we infer it from `real_suffix`?

can/io/player.py
felixdivo:
```suggestion
    def unzip(zipfile: "can.typechecking.StringPathLike") -> Tuple[str, typing.IO[typing.Any]]:
```

felixdivo:
Looks simple and good! 😄

Once this is merged, I suggest opening the first pre-release.

can/io/player.py
Tbruno25: will do!

can/io/player.py
Tbruno25: > Do we need a `mode` argument or shall we infer it from `real_suffix`

We could but I guess I was trying to avoid hard coding it. 

One alternative could be to move the readers `mode` out of `__init__` and make it a class attribute instead. This way we could call something like  `getattr(LogReader.message_readers[suffix], "mode")`. Maybe I'm overthinking this though!

zariiii9003:
This does not really close #1210 yet. The same issue exists in `can.io.logger`. Also there should be a test, which reads and writes files with multiple dots.

can/io/player.py
```diff
@@ -88,6 +92,25 @@ def __new__(  # type: ignore
...
+        """
+        real_suffix = pathlib.Path(filename).suffixes[-2].lower()
+        file = gzip.open(filename, "rt")
+
+        try:
+            for line in file:
+                str(line)
+            file.seek(0)
+        except UnicodeDecodeError:
+            file = gzip.open(filename, "rb")
```
felixdivo, line 110: Isn't this required?

```suggestion
            file.close()  # Close the text file pointer
            file = gzip.open(filename, "rb")
```

can/io/player.py
felixdivo: We can still do this later 😄 I really like your proposal since its elegantly simple!

felixdivo:
> This does not really close #1210 yet. The same issue exists in `can.io.logger`. Also there should be a test, which reads and writes files with multiple dots.

@Tbruno25 Do you have time for that too?

Tbruno25:
> > This does not really close #1210 yet. The same issue exists in `can.io.logger`. Also there should be a test, which reads and writes files with multiple dots.
> 
> @Tbruno25 Do you have time for that too?

Definitely! I can probably wrap this up this weekend.

can/io/player.py
Tbruno25: Not sure it's _required_, but definitely a good idea to not rely on the gc.

felixdivo:
Awesome!

As soon as that is merged, I'll update the changelog, publish a pre-release and create an announcement-issue.

Tbruno25:
> This does not really close #1210 yet. The same issue exists in `can.io.logger`. Also there should be a test, which reads and writes files with multiple dots.

I'm not sure we need to explicitly read/write with multiple dots -- just that _only_ the last suffix is used. I added a variant to `test_extension_matching` which I think satisfies this, although we could make it a separate unittest as well if required.

can/io/player.py
Tbruno25: I ended up inferring `mode` from `real_suffix` as suggested but I definitely think we should implement this later!

Tbruno25:
`LogReader.decompress` and `Logger.compress` are almost identical except for the `mode`. It probably makes sense in the future to unify them as a single function that both classes either import/inherit.

felixdivo:
> `LogReader.decompress` and `Logger.compress` are almost identical except for the `mode`. It probably makes sense in the future to unify them as a single function that both classes either import/inherit.

Yeah, maybe. one could also easily add zip/... support too. For now, I think it is okay as is since it only requires three lines.

felixdivo:
There are still some typing errors left after the cleanup of @zariiii9003 in #1224, would you mind fixing them (they were skipped to avoid merge conflicts)?

zariiii9003:
> There are still some typing errors left after the cleanup of @zariiii9003 in #1224, would you mind fixing them (they were skipped to avoid merge conflicts)?

GzipASCReader and GzipASCWriter can be deleted, right?

pierreluctg:
> > There are still some typing errors left after the cleanup of @zariiii9003 in #1224, would you mind fixing them (they were skipped to avoid merge conflicts)?
> 
> GzipASCReader and GzipASCWriter can be deleted, right?

Yes, they can be removed.

felixdivo (approved):
Wonderful, thank you @Tbruno25!

felixdivo:
@zariiii9003 Do you agree on merging this?

zariiii9003:
> @zariiii9003 Do you agree on merging this?

I'll review it after work.

felixdivo:
Ahh, one thing: [The linter complains](https://github.com/hardbyte/python-can/runs/4947835740?check_suite_focus=true). Seems like this is unrelated though, so we can do this in a subsequent PR. I'll open an issue for this.

can/io/player.py
```diff
@@ -27,12 +27,13 @@ class LogReader(BaseIOHandler):
 
     The format is determined from the file format which can be one of:
       * .asc
-      * .asc.gz
       * .blf
       * .csv
       * .db
       * .log
 
+      Or any of the above compressed using gzip (.gz)
```
zariiii9003, line 35: Same as for the Logger

can/io/player.py
```diff
@@ -27,12 +27,13 @@ class LogReader(BaseIOHandler):
 
     The format is determined from the file format which can be one of:
```
zariiii9003, line 28:
```suggestion
    The format is determined from the file suffix which can be one of:
```

zariiii9003 (approved):
Well done! The docstring could be a little more precise though.

@felixdivo I do not agree with the typehints here but i will follow up with another PR that takes care of the IO type annotations in general. There are some inconsistencies that are masked by casting.

Tbruno25:
> Well done! The docstring could be a little more precise though.
> 
> @felixdivo I do not agree with the typehints here but i will follow up with another PR that takes care of the IO type annotations in general. There are some inconsistencies that are masked by casting.

I'm interested to see the solution regarding type hints! I don't have much experience with `mypy` but I couldn't seem to get it happy unless I used the very vague `Any`.

Issue #1133, opened 2021-09-29
New release (4.0.0)


danielhrisca:
Hello all, is there some target release date for version 4.0.0 final?

felixdivo:
Now, #1046 is "only" waiting for reviews of the changes and merges. The implementation is done.

felixdivo:
Yay, #1046 is complete. I'll provide a release candidate soon (when time permits).

felixdivo:
#1216 adds a changelog

felixdivo:
Waits for a regression to be fixed by #1221. Then there will be version `4.0.0-rc.0` and an announcement issue.

felixdivo:
Okay, so the blocking #1221 and #1230 are resolved. Everything is implemented for the release.

felixdivo:
A new release candidate is available, [`4.0.0rc0`](https://github.com/hardbyte/python-can/releases/tag/4.0.0rc0)! :partying_face:

You can install it with `pip install python-can==4.0.0rc0`.

felixdivo:
Btw, #1249 is not included in that release. It will be in the next one or the full release, depending of whether we need a second *rc* build at all.

felixdivo:
See  #1250. Thank you for your patience. :smiley:

PR #1226, opened 2022-01-22
Add typing for can.Message

brent-stone:
Add mypy typing hints for can.Message for downstream projects.

### DIFF
--- a/can/message.pyi
+++ b/can/message.pyi
@@ -1,0 +1,41 @@
+from typing import Optional, Any
+
+class Message:
+    timestamp: float
+    arbitration_id: int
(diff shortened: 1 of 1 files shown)


felixdivo:
Hey @brent-stone, thank you for your proposal! However, in what regard is this file different from normal type hints? [We do have them](https://github.com/hardbyte/python-can/blob/develop/can/message.py#L51) on the `develop` branch (although they are only released next week or so, see #1133).

brent-stone:
great work with the class. That dev branch update should fix the typing problem in the current version. Thanks. I'll close the pull request.

PR #1229, opened 2022-01-25
Enable mypy for vector interface

zariiii9003:


can/interfaces/vector/canlib.py
```diff
@@ -10,22 +10,16 @@
...
     HAS_EVENTS = True
 except ImportError:
-    try:
-        # Try pywin32 package
-        from win32event import WaitForSingleObject, INFINITE
-
-        HAS_EVENTS = True
-    except ImportError:
-        # Use polling instead
-        HAS_EVENTS = False
```
felixdivo, line 28: I'm unsure about this change, should we really abandon this fallback? The module seems to be private, are we sure it will stay around?

If yes, we should remove this from the docs:

> If you are using Python 2.7 it is recommended to install pywin32, otherwise a slow and CPU intensive polling will be used when waiting for new messages.

`pywin32` is still specified in the `setup.py`, but I suppose it's used elsewhere.

can/interfaces/vector/canlib.py
```diff
@@ -73,20 +67,20 @@ def __init__(
         channel: Union[int, Sequence[int], str],
         can_filters: Optional[CanFilters] = None,
         poll_interval: float = 0.01,
-        receive_own_messages: bool = False,
-        bitrate: int = None,
+        receive_own_messages: Optional[bool] = False,
```
felixdivo, line 70: Why should we allow `None`?

can/interfaces/vector/canlib.py
```diff
@@ -73,20 +67,20 @@ def __init__(
...
+        bitrate: Optional[int] = None,
         rx_queue_size: int = 2 ** 14,
-        app_name: str = "CANalyzer",
-        serial: int = None,
-        fd: bool = False,
-        data_bitrate: int = None,
+        app_name: Optional[str] = "CANalyzer",
+        serial: Optional[int] = None,
+        fd: Optional[bool] = False,
```
felixdivo, line 75: Why should we allow `None`? It either is FD-enabled or not.

can/interfaces/vector/canlib.py
```diff
@@ -657,7 +655,7 @@ def _detect_available_configs() -> List[AutoDetectedConfig]:
                     "vector_channel_config": channel_config,
                 }
             )
-        return configs
+        return configs  # type: ignore
```
felixdivo, line 658: Could we instead specify `configs: List[AutoDetectedConfig] = []` in the beginning of the method? If not, this is fine.

felixdivo (approved):
Great! 😄

can/interfaces/vector/canlib.py
zariiii9003: `_winapi` is only available in CPython. pywin32 is also only available for CPython. So unless CPython removes `_winapi` (why would they?) that code is pointless. I can add a test for that unlikely case :smile:

pywin32 is still used in BroadcastManager.

can/interfaces/vector/canlib.py
zariiii9003: I tried, but TypedDict does not allow extra keys.

can/interfaces/vector/canlib.py
zariiii9003: fixed, thank you!

can/interfaces/vector/canlib.py
zariiii9003: Fixed.

can/interfaces/vector/canlib.py
zariiii9003: I added the test and apparently PyPy supports `_winapi` too.

can/interfaces/vector/canlib.py
felixdivo: Okay, apparently this is not supported. Didn't know that.

can/interfaces/vector/canlib.py
felixdivo: 🥳 Then maybe let that test run there too? Else, nice work!

zariiii9003: The test runs on PyPy too.

felixdivo: Ah, sorry, of course!

felixdivo (approved):

felixdivo:
Two mypy errors have popped up:

```
can/interfaces/vector/canlib.py:22: error: Incompatible types in assignment (expression has type "None", variable has type "Callable[[int, int], int]")
can/interfaces/vector/canlib.py:22: error: Incompatible types in assignment (expression has type "None", variable has type "int")
```

Issue #1230, opened 2022-01-26
Turn the CI green before the next release


felixdivo:
Most of the work was done in #1224 and  #1221, but maybe small stuff is left and we want to have a clean release.

zariiii9003:
There are also some deprecation warnings that we could fix

```
can/interfaces/__init__.py:36
  /home/runner/work/python-can/python-can/can/interfaces/__init__.py:36: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    if "can.interface" in entry:

test/notifier_test.py::AsyncNotifierTest::test_asyncio_notifier
  /home/runner/work/python-can/python-can/test/notifier_test.py:43: DeprecationWarning: There is no current event loop
    loop = asyncio.get_event_loop()
```

PR #1233, opened 2022-01-27
Fix entry_points deprecation

felixdivo:
Part of  #1230.


zariiii9003 (approved):

can/interfaces/__init__.py
```diff
@@ -32,14 +32,16 @@
...
+    except TypeError:
+        # Fallback for Python <3.10
+        # See https://docs.python.org/3/library/importlib.metadata.html#entry-points, "Compatibility Note"
+        entries = entry_points().get("can.interface", [])
+
+    BACKENDS.update(
+        {interface.name: tuple(interface.value.split(":")) for interface in entries}
+    )
 except ImportError:
     from pkg_resources import iter_entry_points
```
zariiii9003, line 46: Do we still need this?

can/interfaces/__init__.py
```diff
@@ -32,14 +32,16 @@
...
-            {
-                interface.name: tuple(interface.value.split(":"))
-                for interface in entry["can.interface"]
-            }
-        )
+    try:
+        entries = entry_points(group="can.interface")
+    except TypeError:
+        # Fallback for Python <3.10
+        # See https://docs.python.org/3/library/importlib.metadata.html#entry-points, "Compatibility Note"
```
zariiii9003, line 39: We could also use `importlib_metadata` from PyPI as fallback

felixdivo: That would also be possible, yep.
As it's just so little effort, I thought it's simpler to just handle the fallback here.

can/interfaces/__init__.py
felixdivo: Probably not for long, see #1234. 😃

PR #1238, opened 2022-01-27
Improve IO typing

zariiii9003:


felixdivo (approved):
I think we can merge this as-is and fix the pylint issues in a subsequent PR to keep the diffs simple  and clear.

felixdivo:
Nice! We're getting there ...

PR #1241, opened 2022-01-28
Prepare 4.0.0-rc.0

felixdivo:
I did not update the [Acknowledgements](https://python-can.readthedocs.io/en/develop/history.html#acknowledgements) section since it's hard to get an overview over the last few years. Even after assembling the changelog, which already took forever. One needs to also update the `CONTRIBUTORS.txt`. **But** both can IMHO also be updated just before the final `4.0.0`, and do not need to be ready for the pre-release. The changelog was more important, since people can check the changes and change their code. I'll open an issue for it, but I'm not eager to actually do it ...

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -8,7 +8,7 @@ messages on a can bus.
 import logging
 from typing import Dict, Any
 
-__version__ = "4.0.0-dev.2"
+__version__ = "4.0.0-rc.0"
 
 log = logging.getLogger("can")
 
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -49,7 +49,7 @@ The CANalyst-II interface was contributed by Shaoyu Meng in 2018.
 Support for CAN within Python
 -----------------------------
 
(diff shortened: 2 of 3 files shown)


felixdivo:
@hardbyte I checked [the docs for creating a new release](https://python-can.readthedocs.io/en/develop/development.html#creating-a-new-release) and found that I am not able to GPG-sign the wheels (since of course, I don't have your private key). Can you do that instead?

felixdivo:
Or should I just use [this](https://github.com/hardbyte/python-can/blob/develop/.travis.yml#L77) Travis CI job instead? Semms like "no", since that only works when releasing from the `master` branch ([doc](https://docs.travis-ci.com/user/deployment/pypi/#deploying-specific-branches)).

hardbyte:
I'm traveling(!) until mid next week and don't have gpg keys on me. Happy for you to release binaries without my signature or without any signature though.

felixdivo:
Okay, then enjoy it :smile: I'll be unavailable 'till then too and will probably release it ~Thursday.

Issue #1245, opened 2022-01-29
[io][asc]: parser / writer using DLC as data length


jazi007:
### Describe the bug
**Reader**
https://github.com/hardbyte/python-can/blob/develop/can/io/asc.py#L197: using DLC (data length code) from ASC as dlc (Data length) in python-can message. I think `data_length` must be used instead

**Writer**
https://github.com/hardbyte/python-can/blob/develop/can/io/asc.py#L384: inverse is done here. instead `len2dlc(msg.dlc)` 

### To Reproduce
asc_bug.py at python-can root

```python
from pathlib import Path
from can import LogReader

if __name__ == "__main__":
    path = Path(__file__).parent / "test" / "data" / "test_CanFdMessage64.asc"
    reader = LogReader(path)
    for msg in reader:
        print(msg)
```
output:
```bash
Timestamp:       30.506898        ID: 04ee    S Rx     F    EI    DL: 15    a1 02 03 04 00 00 00 00 00 00 00 00 00 00 00    Channel: 3
Timestamp:       31.506898    ID: 1c4d80a7    X Rx     F BS       DL: 15    b1 02 03 04 00 00 00 00 00 00 00 00 00 00 00    Channel: 3
```

In the `test_CanFdMessage64.asc` frames are 64 bytes length

### Expected behavior
frames should have 64 bytes length

### Additional context
used python-can is develop branch

PR #1246, opened 2022-01-29
[IO][ASC]: fix data length

jazi007:
Fix https://github.com/hardbyte/python-can/issues/1245


can/io/asc.py
```diff
@@ -381,8 +390,8 @@ def on_message_received(self, msg: Message) -> None:
                 symbolic_name="",
                 brs=1 if msg.bitrate_switch else 0,
                 esi=1 if msg.error_state_indicator else 0,
-                dlc=msg.dlc,
-                data_length=len(data),
+                dlc=len2dlc(msg.dlc),
```
felixdivo, line 393: Yeah, `msg.dlc` is confusing (#1212). This change and the one to `test/logformats_test.py` looks good.

felixdivo (approved):
Looks good overall! Thanks for the analysis!

PR #1252, opened 2022-02-05
Pass flags instead of flags_t type for USB2CAN

jacobian91:
It looks like flags_t the type was being used instead of the flags argument.

### DIFF
--- a/can/interfaces/usb2can/usb2canInterface.py
+++ b/can/interfaces/usb2can/usb2canInterface.py
@@ -94,7 +94,7 @@ class Usb2canBus(BusABC):
         channel=None,
         dll="usb2can.dll",
         flags=0x00000008,
-        *args,
+        *_,
         bitrate=500000,
         **kwargs,
     ):
@@ -119,9 +119,7 @@ class Usb2canBus(BusABC):
         connector = f"{device_id}; {baudrate}"
         self.handle = self.can.open(connector, flags)
 
-        super().__init__(
-            channel=channel, dll=dll, flags=flags, bitrate=bitrate, *args, **kwargs
-        )
+        super().__init__(channel=channel, **kwargs)
 
     def send(self, msg, timeout=None):
         tx = message_convert_tx(msg)
--- a/can/interfaces/usb2can/usb2canInterface.py
+++ b/can/interfaces/usb2can/usb2canInterface.py
@@ -9,7 +9,6 @@ from typing import Optional
 from can import BusABC, Message, CanInitializationError, CanOperationError
 from .usb2canabstractionlayer import Usb2CanAbstractionLayer, CanalMsg, CanalError
 from .usb2canabstractionlayer import (
-    flags_t,
     IS_ERROR_FRAME,
     IS_REMOTE_FRAME,
     IS_ID_TYPE,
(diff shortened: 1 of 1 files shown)


jacobian91:
This is the traceback when encountering this error:

```python
Unable to create CAN node Traceback (most recent call last):

File "...\site-packages\can\interfaces\usb2can\usb2canabstractionlayer.py", line 141, in open
    result = self.__m_dllBasic.CanalOpen(config_ascii, flags)
             │                           │             └ <class 'ctypes.c_ulong'>
             │                           └ b'F8BAFED8; 500'
             └ <can.interfaces.usb2can.usb2canabstractionlayer.Usb2CanAbstractionLayer object at 0x0000020F03E5D370>

ctypes.ArgumentError: argument 2: <class 'TypeError'>: Don't know how to convert parameter 2
```

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -9,7 +9,6 @@
 from can import BusABC, Message, CanInitializationError, CanOperationError
 from .usb2canabstractionlayer import Usb2CanAbstractionLayer, CanalMsg, CanalError
 from .usb2canabstractionlayer import (
-    flags_t,
```
jacobian91, line 12: The only reason I removed this is that with this change it is unused in the file.

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -118,10 +117,10 @@ def __init__(
         self.channel_info = f"USB2CAN device {device_id}"
 
         connector = f"{device_id}; {baudrate}"
-        self.handle = self.can.open(connector, flags_t)
+        self.handle = self.can.open(connector, flags)
 
         super().__init__(
-            channel=channel, dll=dll, flags_t=flags_t, bitrate=bitrate, *args, **kwargs
+            channel=channel, dll=dll, flags=flags, bitrate=bitrate, *args, **kwargs
```
felixdivo, line 123: Since `BusABC` does not know these, we can also ignore them here.

```suggestion
            channel=channel, bitrate=bitrate, *args, **kwargs
```

felixdivo (approved):

felixdivo:
Thanks @jacobian91!

can/interfaces/usb2can/usb2canInterface.py
jacobian91: I've also removed bitrate as it looks like `BusABC` doesn't use this either, so now this line is `super().__init__(channel=channel, **kwargs)`.
I've removed the `*args` as well to prevent errors from positional vs keyword argument ordering. the `BusABC` argument `can_filters` could get passed into `Usb2canBus` via `kwargs`.

can/interfaces/usb2can/usb2canInterface.py
```diff
@@ -94,7 +94,7 @@ def __init__(
         channel=None,
         dll="usb2can.dll",
         flags=0x00000008,
-        *args,
+        *_,
```
jacobian91, line 97: Optional style change, this documents that all additional positional arguments get ignored but won't throw an exception when passed in.

felixdivo (approved):

PR #1257, opened 2022-02-08
ASCReader bugfix

### LINKED ISSUE #1256: Error when replaying ASC file
I get the following error when I replay an ASC file:

```
09:07:47  self = <can.io.asc.ASCReader object at 0x0000020898E659D0>
09:07:47  
09:07:47      def __iter__(self) -> Generator[Message, None, None]:
09:07:47          self._extract_header()
09:07:47      
09:07:47          for line in self.file:
09:07:47              line = line.strip()
09:07:47      
09:07:47              if not re.match(
09:07:47                  r"\d+\.\d+\s+(\d+\s+(\w+\s+(Tx|Rx)|ErrorFrame)|CANFD)",
09:07:47                  line,
09:07:47                  re.ASCII | re.IGNORECASE,
09:07:47              ):
09:07:47                  # line might be a comment, chip status,
09:07:47                  # J1939 message or some other unsupported event
09:07:47                  continue
09:07:47      
09:07:47              msg_kwargs: Dict[str, Union[float, bool, int]] = {}
09:07:47              try:
09:07:47                  _timestamp, channel, rest_of_message = line.split(None, 2)
09:07:47  >               timestamp = float(_timestamp) + self.start_time
09:07:47  E               AttributeError: 'ASCReader' object has no attribute 'start_time'
```

the error occurs [here](https://github.com/hardbyte/python-can/blob/develop/can/io/asc.py#L225)

This is the input file
[CAN.txt](https://github.com/hardbyte/python-can/files/8021586/CAN.txt)

_Originally posted by @danielhrisca in https://github.com/hardbyte/python-can/issues/1250#issuecomment-1032280811_

zariiii9003:
ASCReader.start_time was never set, when no trigger block was present.
Closes #1256

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -34,8 +34,6 @@ class ASCReader(MessageReader):
 
     file: TextIO
 
-    FORMAT_START_OF_FILE_DATE = "%a %b %d %I:%M:%S.%f %p %Y"
-
     def __init__(
         self,
         file: Union[StringPathLike, TextIO],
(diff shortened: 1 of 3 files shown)


zariiii9003:
> Oh wow, this is surprisingly large. Looks good to me!

Yeah, i tested a few more files and noticed that the date conversion failed almost every time. To be honest i don't like the ASCReader implementation that much. [This C++ ASC implementation](https://bitbucket.org/tobylorenz/vector_asc/src/master/src/Vector/ASC/File.cpp) seems much more elegant. We still have some room for improvement.

PR #1260, opened 2022-02-18
Release/4.0.0

hardbyte:
Merges `master` into `develop` (accepting develop for all conflicts). 

Might take the opportunities to get with the times and rename `master` -> `main` in github too.


hardbyte:
I don't think there is any point of requesting a review on a 30k diff!

PR #1270, opened 2022-03-02
Improve gs_usb usability and fix loopback frames

chemicstry:
This improves the `gs_usb` interface by adding ability to specify device index instead of USB bus/address. The previous method was really inconvenient, because bus/address changes with each USB enumeration. This required libusb dependency in end applications to scan for adapters.

Additionaly, this fixes `is_extended_id` and loopback (echo) frames, by correctly marking `is_rx` flag.

### DIFF
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -1,7 +1,7 @@
 from typing import Optional, Tuple
 
 from gs_usb.gs_usb import GsUsb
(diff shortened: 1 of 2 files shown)


chemicstry:
Anything preventing this from being merged?

zariiii9003:
> Anything preventing this from being merged?

Just a lack of active maintainers...

zariiii9003 (approved):
Looks good to me.

PR #1271, opened 2022-03-02
[IO][canutils]: use common CAN interface names in generated logfile

hartkopp:
CAN interfaces in canutils logfiles are usually named 'can0', 'can32' or
'vcan8'. This allows to split logfiles just by performing 'grep' or to
rename CAN interfaces easily with 'sed'.

The current implementation of the canutils log file writer just provides
channel numbers for CAN interfaces. This patch adds the string 'can' to
the channel number to make it look like a usual canutils logfiles that can
be preprocessed as described above.

Signed-off-by: Oliver Hartkopp <socketcan@hartkopp.net>

### DIFF
--- a/can/io/canutils.py
+++ b/can/io/canutils.py
@@ -160,6 +160,8 @@ class CanutilsLogWriter(FileIOMessageWriter):
             timestamp = msg.timestamp
 
         channel = msg.channel if msg.channel is not None else self.channel
+        if isinstance(channel, int) or isinstance(channel, str) and channel.isdigit():
+            channel = f"can{channel}"
 
         framestr = "(%f) %s" % (timestamp, channel)
 


can/io/canutils.py
```diff
@@ -161,7 +161,7 @@ def on_message_received(self, msg):
 
         channel = msg.channel if msg.channel is not None else self.channel
 
-        framestr = "(%f) %s" % (timestamp, channel)
+        framestr = "(%f) can%s" % (timestamp, channel)
```
zariiii9003, line 164: msg.channel might can be a string like `"can0"`. You'd get `"cancan0"` here.

hartkopp: Good point!
Unfortunately I'm not really familiar with Python, but would it make sense to add some `isdigit()` vodoo to catch the case that we are only dealing with numbers and add `"can"` only in these cases?

zariiii9003: you could use
```python3
if isinstance(channel, int):
    ...
```

hartkopp: Hi @zariiii9003 ,
I looked at the docs and also at some code - but I don't have Python knowledge that goes beyond my original patch :-/
Would you mind to provide a patch/PR that adds `"can"` before the `channel` only when channel is a simple number?
Thanks!

hardbyte: Done

hartkopp: Big thanks!

zariiii9003 (approved):

PR #1278, opened 2022-03-09
Fix timestamp handling in udp_multicast on macOS

### LINKED ISSUE #1275: Error with Multicast IP Interface
Hello everyone, I'm fairly new to the python-can environment and I'm trying to get the basic example for the Multicast IP Interface. 
I've installed the python-can module using pip, I've copied the example for this specific CAN connection from the documentation ([https://python-can.readthedocs.io/en/develop/interfaces/udp_multicast.html#bus-class-documentation](https://python-can.readthedocs.io/en/develop/interfaces/udp_multicast.html#bus-class-documentation)), but I seem to get an error that has to do with the bus.py file in the interface installation folder. 

The error I'm getting is the following:

```
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/can/interfaces/udp_multicast/bus.py", line 241, in _create_socket    sock.setsockopt(socket.SOL_SOCKET, SO_TIMESTAMPNS, 1)
OSError: [Errno 42] Protocol not available
```

This feels really weird since the same function is called just a few lines prior in the bus.py file, but it does not raise an error.

I'm on MacOS using python3.10 and python-can version 4.0.0

Is there something I'm doing wrong? Any help would be really appreciated.

felixdivo:
Fix #1275.


felixdivo:
@Gudoghe Can you check the fix? 🤔 

You should be able to install it with `pip install git https://github.com/hardbyte/python-can.git@felixdivo-fix-1275`?

felixdivo:
You should now also be able to run the tests, e.g. with:

```sh
pip install tox
tox -e py
```

Could you post the results for the test suite parts `BasicTestUdpMulticastBusIPv4` and `BasicTestUdpMulticastBusIPv6`? They should now run on your system (and not be skipped any more).

Gudoghe:
@felixdivo The command to install the fix seems to return an error 404, maybe there's some problem with the url?

felixdivo:
Ah, jup, there's a missing `+`, sorry. This works for me: `pip install git+https://github.com/hardbyte/python-can.git@felixdivo-fix-1275`.

Gudoghe:
The patch and tox were installed successfully but the `tox -e py` command gives me the following error:
`ERROR: tox config file (either pyproject.toml, tox.ini, setup.cfg) not found`

Sorry I'm quite new to this tool and environment.

felixdivo:
Strange ... [there should be one](https://github.com/hardbyte/python-can/blob/develop/tox.ini#L22). Are you sure that you are running it from the repository root? `pwd` (on Unix) should give you something ending in `/python-can`.

Alternatively, you can also try `tox -e gh` one it works.

Gudoghe:
Alright I managed to clone the patched repo and install python can from there. I then ran the tests on the cloned repository and all the tests regarding `BasicTestUdpMulticastBusIPv4` and `BasicTestUdpMulticastBusIPv6` failed, I'm attaching a screenshot. 

Moreover I tried to run again the example from the Multicast page and the error i get back changed to the following:

> Traceback (most recent call last):
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/can/interfaces/udp_multicast/bus.py", line 264, in _create_socket
>     sock.bind(("", self.port))
> OSError: [Errno 48] Address already in use


Schermata 2022-03-11 alle 16 09 12

felixdivo:
Uhm, that's strange. Can you find out what is using that address, or use another one (suitable for multicast, like [the default one](https://python-can.readthedocs.io/en/master/interfaces/udp_multicast.html#can.interfaces.udp_multicast.UdpMulticastBus.DEFAULT_GROUP_IPv6))? I can't really debug it from here 🙃. But at least the exception from last time is avoided. I suppose the tests fail due to similar problems?

Gudoghe:
I've tried looking at all processes running with the command `ps -fA` but nothing seems to be taking up that port. 
I've tried changing the port to be assigned in the range of 43001-43187, as to my understanding this seems to be the proper range ([https://www.speedguide.net/port.php?port=43113](url)), but I now run into a new error:

>  Traceback (most recent call last):
>   File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/can/interfaces/udp_multicast/bus.py", line 273, in _create_socket
>     sock.setsockopt(socket.IPPROTO_IPV6, socket.IPV6_JOIN_GROUP, request)
> OSError: [Errno 49] Can't assign requested address

I'm really at a blank for a solution. If you had any suggestions on what to try next that'd be greatly appreciated.
Thank you again for all the support!

felixdivo:
You're welcome. :smiley:  I am currently retreating from this repo but since it wrote that interface I figured that I should have a look at this.

Hm ... Are you sure that multicast works elsewhere? I cold not find a simple test command, but I did find [this article](https://blog.znsio.com/2009/11/08/enabling-multicast-on-your-macos-unix/). It's from 2009, but looks relevant.

felixdivo:
I'm a bit confused: In the exception above you use IPv6 (and it fails) and now you checked IPv4 multicast (and it works). Maybe, you can try IPv4 multicast by using something like [`DEFAULT_GROUP_IPv4`](https://python-can.readthedocs.io/en/master/interfaces/udp_multicast.html#can.interfaces.udp_multicast.UdpMulticastBus.DEFAULT_GROUP_IPv4)? Maybe macOS only has problems with IPv6 multicast?

felixdivo:
Okay, so this became kinda stale. I suggest merging at lest these changes, since they at least resolve some problems. If more occur, we can still keep discussing them and add more patches.

Side note: @zariiii9003 I wrote about reducing my contributions and why I do that [on the mailing list](https://groups.google.com/g/python-can/c/Dio_wKBzbas/m/sw2S-2MgAQAJ) 🙂

can/interfaces/udp_multicast/bus.py
```diff
@@ -200,8 +207,13 @@ def __init__(
 
         # used in recv()
         self.received_timestamp_struct = "@ll"
-        ancillary_data_size = struct.calcsize(self.received_timestamp_struct)
-        self.received_ancillary_buffer_size = socket.CMSG_SPACE(ancillary_data_size)
+        self.received_timestamp_struct_size = struct.calcsize(
+            self.received_timestamp_struct
+        )
+        if self.timestamp_nanosecond:
```
zariiii9003, line 213: `self.timestamp_nanosecond` should be declared in `__init__`

can/interfaces/udp_multicast/bus.py
```diff
@@ -320,21 +340,37 @@ def recv(
                 self.max_buffer, self.received_ancillary_buffer_size
```
zariiii9003, line 340: The attribute `self.received_ancillary_buffer_size` does not exist if `self.timestamp_nanosecond == False`

can/interfaces/udp_multicast/bus.py
```diff
@@ -320,21 +340,37 @@ def recv(
...
-            seconds, nanoseconds = struct.unpack(
-                self.received_timestamp_struct, cmsg_data
-            )
-            if nanoseconds >= 1e9:
-                raise can.CanError(
-                    f"Timestamp nanoseconds field was out of range: {nanoseconds} not less than 1e9"
+            # fetch timestamp; this is configured in _create_socket()
+            if self.timestamp_nanosecond:
+                # Very similar to timestamp handling in can/interfaces/socketcan/socketcan.py -> capture_message()
+                assert len(ancillary_data) == 1, "only requested a single extra field"
```
zariiii9003, line 346: assert statements are ignored if python is run with the [-o option](https://docs.python.org/3/using/cmdline.html#cmdoption-O)

can/interfaces/udp_multicast/bus.py
```diff
@@ -320,21 +340,37 @@ def recv(
...
+                cmsg_level, cmsg_type, cmsg_data = ancillary_data[0]
+                assert (
```
zariiii9003, line 348: same as above

zariiii9003 (changes requested):

can/interfaces/udp_multicast/bus.py
felixdivo: Good catch

can/interfaces/udp_multicast/bus.py
felixdivo: Discussed above.

can/interfaces/udp_multicast/bus.py
felixdivo: Okay, you are right. But that's how it is done in *socketcan* (and in a single line in `can/io/blf.py`) too. Shall I change those too?

zariiii9003: No, at least not in this PR. But generally production code should not contain assert statements.

felixdivo: Jup, you're right. Maybe a linter could check that.

felixdivo:
[Black has a problem](https://github.com/hardbyte/python-can/runs/6085042693?check_suite_focus=true#step:5:14). We should just update (https://github.com/psf/black/issues/2964). I added a commit for that to this PR.

zariiii9003:
@felixdivo Do you know why the UdpMulticast tests are failing randomly?

felixdivo:
Nope ... the type of error also confuses me.

hardbyte (approved):

PR #1281, opened 2022-03-11
%d format is for a number, not str

pierreluctg:
Failure to initiate the Neousys DLL was causing `TypeError: %d format: a number is required, not str` error since the log line argument is not a number.


felixdivo (approved):
Looks good, but such problems should be detected by the test suite. A regression test would probably be a good idea.

Issue #1282, opened 2022-03-14
`send_periodic` does not accept `on_error` callback, resulting in race condition


andebjor:
### Describe the bug
When setting up a periodic send task there is no way to configure the `on_error` callback in the task. Therefore, if the very first send attempt in the task fails, the task is aborted. This can happen before the user is able to add a callback to the newly created task. Thus there is a race condition between creating the task and sending the first message, and adding the callback.

### To Reproduce
Open a bus and create a periodic task. If sending to the bus fails, the task is aborted due to no callback being configured.

### Expected behavior
An argument is added to `send_periodic` such that the callback can be set in the constructor of the task.

PR #1283, opened 2022-03-14
Add parameter for `on_error` to `BusABC.send_periodic`

andebjor:
If there is an error while sending messages in a (background) periodic
task, then the `on_error` callback is called, if set. If the callback is
configured and returns `True` then the task continues, otherwise it is
aborted.

While it is possible to update a task with a callback after the task has
been created, this procedure is prone to a race condition where the
callback might not be configured in time for the first send event. Thus,
if the first send fails and the callback has not yet been configured,
the task will abort.

This commit solves the race condition issue by adding an argument to
`BusABC.send_periodic` to specify the callback. By including the
callback in the constructor it will be deterministically active for all
sends in the task. This fixes issue #1282.

This commit also adds myself to the CONTRIBUTORS list.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -2,7 +2,7 @@
 Contains the ABC bus implementation and its documentation.
 """
 
-from typing import cast, Any, Iterator, List, Optional, Sequence, Tuple, Union
+from typing import cast, Any, Callable, Iterator, List, Optional, Sequence, Tuple, Union
 
 import can.typechecking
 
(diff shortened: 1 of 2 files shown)


can/bus.py
```diff
@@ -232,7 +240,7 @@ def send_periodic(
         # Create a backend specific task; will be patched to a _SelfRemovingCyclicTask later
         task = cast(
             _SelfRemovingCyclicTask,
-            self._send_periodic_internal(msgs, period, duration),
+            self._send_periodic_internal(msgs, period, duration, on_error),
```
zariiii9003, line 243: This will fail for subclasses that reimplement `_send_periodic_internal()`

PR #1289, opened 2022-04-07
Mf4 io

### LINKED ISSUE #506: add MF4 io writer

driftregion:
I rebased https://github.com/hardbyte/python-can/pull/554/

Closes #506


felixdivo:
[mf4-io](https://github.com/hardbyte/python-can/tree/mf4-io) can probably be deleted, should this be merged.

zariiii9003:
@driftregion can i force push to this PR?
I fixed a few things [here](https://github.com/zariiii9003/python-can/tree/mf4-io). I can open a new PR, too.

driftregion:
@zariiii9003 sure, go ahead.

zariiii9003:
@danielhrisca would it be possible to stream to a file directly?  
Also: i accessed a private attribute to get the current file size, is there a better way to do that?

danielhrisca:
> @danielhrisca would it be possible to stream to a file directly?

The only way to `stream` to the file is to use the `extend` method, but this is performance bottleneck for high data rates.

> Also: i accessed a private attribute to get the current file size, is there a better way to do that?
I don't think so. Why do you need that information?

zariiii9003:
@danielhrisca 
The SizedRotatingLogger starts a new file when a give file size is reached. That's why i added the `file_size` function to the logger classes.

The BlfWriter compresses and writes data in chunks so maybe that could be possible here too?

setup.py
```diff
@@ -36,6 +36,10 @@
     "viewer": [
         'windows-curses;platform_system=="Windows" and platform_python_implementation=="CPython"'
     ],
+    "mf4": [
+        'asammdf>=6.0.0;platform_python_implementation=="CPython" and python_version<"3.11"',
+        'numpy;platform_python_implementation=="CPython" and python_version<"3.11"',
```
hardbyte, line 41: `asammdf` will declare the version of numpy it needs. Not sure about the platform and python version requirement either?

doc/listeners.rst
```diff
@@ -191,3 +191,32 @@ The following class can be used to read messages from BLF file:
 
 .. autoclass:: can.BLFReader
     :members:
+    
+  
+MF4 (Measurement Data Format v4)
+--------------------------------
+
+Implements support for MF4 (Measurement Data Format v4) which is a proprietary
+format from ASAM, widely used in many automotive software (Vector CANape, ETAS INCA, dSPACE ControlDesk, etc.).
```
hardbyte, line 200:
```suggestion
format from ASAM (Association for Standardization of Automation and Measuring Systems), widely used in 
many automotive software (Vector CANape, ETAS INCA, dSPACE ControlDesk, etc.).
```

can/io/mf4.py
```diff
@@ -0,0 +1,468 @@
...
+        self._std_buffer = np.zeros(1, dtype=STD_DTYPE)
+        self._err_buffer = np.zeros(1, dtype=ERR_DTYPE)
+        self._rtr_buffer = np.zeros(1, dtype=RTR_DTYPE)
+
+
+class MF4Reader(MessageReader):
+    """
+    Iterator of CAN messages from a MF4 logging file.
+
+    The MF4Reader only supports MF4 files that were recorded with python-can.
```
hardbyte, line 264: Why is that?

setup.py
zariiii9003: I think i added this so CI does not fail on PyPy and CPython 3.11. But there's probably a better solution for that. I admit this is ugly.

can/io/mf4.py
zariiii9003: It's a limitation of asammdf. See [here](https://github.com/danielhrisca/asammdf/issues/744#issuecomment-1207942077)

zariiii9003:
@danielhrisca merged after only 3.5 years 💪

danielhrisca:
It's like a good wine

PR #1309, opened 2022-05-06
fix: conversion for port number

Thepowa753:
If reading the port number (for remote can) from a config file, the dict contains a string, that crashes becouse it require an integer.

### DIFF
--- a/can/util.py
+++ b/can/util.py
@@ -215,12 +215,12 @@ def _create_bus_config(config: Dict[str, Any]) -> typechecking.BusConfig:
         if isinstance(config["port"], str):
         	if config["port"].isnumeric():
         		config["port"] = int(config["port"])
-        		if not 0 < config["port"] < 65535:
-        			raise ValueError("Port config must be inside 0-65535 range!")
         	else:
         		raise ValueError("Port config must be a number!")
         elif not isinstance(config["port"], int):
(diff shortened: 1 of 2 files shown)


hardbyte (approved):

can/util.py
```diff
@@ -211,7 +211,16 @@ def _create_bus_config(config: Dict[str, Any]) -> typechecking.BusConfig:
...
+    if "port" in config:
+        if isinstance(config["port"], str):
+            if config["port"].isnumeric():
+                config["port"] = int(config["port"])
+            else:
+                raise ValueError("Port config must be a number!")
+        elif not isinstance(config["port"], int):
+            raise TypeError("Port config must be string or integer!")
+    if not 0 < config["port"] < 65535:
+        raise ValueError("Port config must be inside 0-65535 range!")
```
hardbyte, line 223: This needs to be inside the `if "port" in config:` block

Thepowa753: oh, thank you, I'm an idiot😅

Thepowa753:
I would need some help in figuring out how to fix mypy error `No overload variant of "int" matches argument type "object"`.
I had to add the forcing to `int()` in `0 < int () < 25565` because otherwise it couldn't find the `<` operand between int and object. (I think it's because it doesn't take into account the checks made before.)

PR #1327, opened 2022-06-06
Enhance `can.logger` to consider the `append` option

### LINKED ISSUE #1326: Enhance the `can.logger` to check to see if the log file exists and append if there rather than overwrite

Consider `can.logger` being used on an embedded Linux system. A `python can.logger ...` process is started as a daemon (`/etc/init.d/`) in boot and killed in shutdown. When the system is restarted, the existing log file on the device and its contents is currently overwritten. This results in a loss of data. 

Upon executing `python can.logger ...` the program should check to see if the specified log file exists. If it does exist, then the file should be appended to. 

An optional argument could be added (perhaps `-a` or `-w`) to give the user the option as to whether an existing file, if found, should be overwritten or appended to. 

### Additional context

P.S. Feel free to assign this issue to me. I should be around to it by Sunday or Monday.

j-c-cook:
The goal here appears to be to give access to the boolean `append` option of `Printer`. 

https://github.com/hardbyte/python-can/blob/0d7f65f74b8b015fb650e9910a920323023f310d/can/io/printer.py#L16-L40

closes #1326

### DIFF
--- a/can/logger.py
+++ b/can/logger.py
@@ -166,6 +166,7 @@ def main() -> None:
     parser.add_argument(
         '-a',
         '--append',
+        # action=argparse.BooleanOptionalAction,  # Use when Python>=3.9
         dest='append_mode',
         type=bool,
         help='Whether to overwrite or append to an existing log file if it exists.',
@@ -209,6 +210,29 @@ def main() -> None:
 
     print(f"Connected to {bus.__class__.__name__}: {bus.channel_info}")
     print(f"Can Logger (Started on {datetime.now()})")
+
+    # argparser.py in Python 3.7.11 does not properly consider boolean arguments passed
+    # in from the command line. If a boolean is passed in, the variable must be converted
+    # to an integer 0 (False) or 1 (True). Once the project goes to a minimum of
+    # Python 3.9, the BooleanOptionalAction, this will no longer be required (Please do
+    # not use this as a reason to advocate a minimum Python version of 3.9).
+    boolean_args = ['-a', '--append']
+    args = sys.argv[1:]
+    for i, bool_arg in enumerate(boolean_args):
+        for j, arg in enumerate(args):
+            if bool_arg == arg:
+                # Make sure the length of args is long enough to check the action
+                if len(args) >= j+1:
+                    action = args[j+1]
+                else:
+                    break
+                # Consider the action, and update the results to accurately embody request
+                if action == '0' or action == 'False':
+                    results.append_mode = False
+                elif action == '1' or action == 'True':
+                    results.append_mode = True
+                else:
+                    pass
 
     options = {'append': results.append_mode}
     if results.file_size:
--- a/can/logger.py
+++ b/can/logger.py
@@ -157,10 +157,12 @@ def main() -> None:
         default=None,
     )
 
-    parser.add_argument("--a",
-                        "--append",
-                        help="Append to the log file if it already exists.",
-                        action="store_true")
+    parser.add_argument(
+        "--a",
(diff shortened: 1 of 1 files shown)


can/logger.py
```diff
@@ -201,12 +211,36 @@ def main() -> None:
...
+                    action = args[j+1]
+                else:
+                    break
+                # Consider the action, and update the results to accurately embody request
+                if action == '0' or action == 'False':
+                    results.append_mode = False
+                elif action == '1' or action == 'True':
+                    results.append_mode = True
+                else:
+                    pass
```
j-c-cook, line 235: I added a manual `argparse` to search for `boolean_args` and correct the `results` if necessary. 

NOTE: Only required until Python>=3.9
NOTE: Please do not use this as a reason to advocate Python>=3.9.

j-c-cook:
@hardbyte @zariiii9003 This accomplishes what was descriped in #1327 and is ready for review.

zariiii9003:
Could you fix the linters?

j-c-cook:
@zariiii9003 The `*lack` format is failing here (https://github.com/hardbyte/python-can/runs/6761006357?check_suite_focus=true), but when I run it locally the checks pass. 

```
$ black --check --verbose logger.py
Identified `/home/user/Documents/python-can` as project root containing a .git directory.
Sources to be formatted: "can/logger.py"
logger.py wasn't modified on disk since last run.

All done! ✨ 🍰 ✨
1 file would be left unchanged.
```

zariiii9003:
I guess you are using a different black version. Run `pip install -r requirements-lint.txt`

j-c-cook:
@zariiii9003 Do you want for me to add in tests to try to resolve the https://github.com/hardbyte/python-can/pull/1327/checks?check_run_id=6775694757 failure?

can/logger.py
```diff
@@ -201,12 +219,42 @@ def main() -> None:
     print(f"Connected to {bus.__class__.__name__}: {bus.channel_info}")
     print(f"Can Logger (Started on {datetime.now()})")
 
+    # argparser.py in Python 3.7.11 does not properly consider boolean
```
zariiii9003, line 222: You don't need any of this. Look at how the `--fd` argument is handled. It is also a boolean.

j-c-cook: I agree. I have modified my changes. This simplifies things. Thank you.

can/logger.py
```diff
@@ -154,12 +157,20 @@ def main() -> None:
         default=None,
     )
 
+    parser.add_argument(
+        "--a",
```
zariiii9003, line 161:
```suggestion
        "-a",
```

can/logger.py
```diff
@@ -154,12 +157,20 @@ def main() -> None:
...
+        "--append",
```
zariiii9003, line 162:
```suggestion
        "--append",
        dest="append",
```

can/logger.py
```diff
@@ -201,12 +212,13 @@ def main() -> None:
     print(f"Connected to {bus.__class__.__name__}: {bus.channel_info}")
     print(f"Can Logger (Started on {datetime.now()})")
 
+    options = {"append": results.a}
```
zariiii9003, line 215:
```suggestion
    options = {"append": results.append}
```

zariiii9003 (changes requested):

j-c-cook:
@zariiii9003 I committed your recommendations. Next time you do that I'll make sure to accept them from this UI rather than manually change them.

zariiii9003 (approved):

PR #1333, opened 2022-06-19
Fix fileno error on Windows (Serial bus)

MattWoodhead:
I was having issues with the Async example from the docs using the Serial interface: `io.UnsupportedOperation: fileno`

I looked back through the issues and found that several people have hit the fileno error over the years (#760, #877, #902). I have implemented the most recent fix from #1313 on the serial_bus.py interface and added a test case to the test_serial.py file.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,13 +1,17 @@
 python-can
 ==========
 
-|release| |python_implementation| |downloads| |downloads_monthly| |formatter|
+|release| |conda forge release| |python_implementation| |downloads| |downloads_monthly| |formatter|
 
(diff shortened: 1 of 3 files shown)


zariiii9003:
Thank you! Could you undo the Readme changes? The conda-forge version seems unmaintained.

MattWoodhead:
Sorry @zariiii9003, I have reverted the readme change. That must have snuck in when I rebased my local repository. I made a pull request to the conda-forge feedstock repository some time ago, but as you noted it appears to be unmainained.

zariiii9003 (approved):

PR #1343, opened 2022-07-07
Exclude repository-configuration from git-archive

umlaeute:
tracking repository configuration (like `.gitignore`) and CI-configuration (`.github/`, `.travis.yml`) within a repository is nice, practical and sometimes unavoidable.

however, such configs do not place nicely when downstreams import the archives into their own git-based workflows.

e.g. for the Debian packaging of `python-can`, we definitely do not want to hide any build-artifacts via `.gitignore` (our workflow is very different from that of upstream development).
we also do not want to accidentally trigger our CI (our CI is supposed to test the packaging, rather than the development).
otoh, we also want to follow the upstream sources as close as possible (that is: without adding/changing/removing any files from the upstream releases).

the typical fix for this is to simply not include the repository-configuration when exporting a snapshot, such as is created automatically by github when downloading a "Release Tarball" (which is really just running `git archive` in the background).

the `.gitattributes` file in this PR does exactly this: it excludes a number of repository- and CI-configuration files from `git archive`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+.git*      	export-ignore
+.*.yml  	export-ignore


zariiii9003:
I'm not sure i understand your use case. Debian packaging clones the python-can github repo? Why wouldn't they use the source distribution on PyPI?

umlaeute:
We use the source tarballs from github (*not* a clone of GH repository).

iirc, the reason being that the we want to run all the checks (as found in the upstream development, but often stripped from pypi packages) but avoid binary artifacts (often found in pypi packages).

this is a generic answer (I'm currently afk and cannot check the specific reason for python-can)

otoh: do you have a specific usecase why you want to include `.gitignore` and `.github/` in the release tarballs?

zariiii9003 (approved):

PR #1344, opened 2022-07-07
Add py.typed file and distribute it upon installation

### LINKED ISSUE #1340: Include a py.typed file so mypy works for downstream packages
@felixdivo We might need to include a [py.typed](https://www.python.org/dev/peps/pep-0561/#id13) file so mypy works for downstream packages. But i'm not sure if it causes problems since `can.interfaces` is not really well annotated.

_Originally posted by @zariiii9003 in https://github.com/hardbyte/python-can/issues/1227#issuecomment-1019561010_

felixdivo:
This should allow downstream packages and "end users" to use type checkers more easily.

Fixes #1340.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,7 @@ setup(
         "": ["README.rst", "CONTRIBUTORS.txt", "LICENSE.txt", "CHANGELOG.md"],
         "doc": ["*.*"],
         "examples": ["*.py"],
+        "can": ["py.typed"],
     },
     # Installation
     # see https://www.python.org/dev/peps/pep-0345/#version-specifiers


felixdivo:
@rumpelsepp Do you mind checking if this change works for you? This should work:

```bash
pip uninstall python-can
pip install git+https://github.com/felixdivo/python-can.git@add-py.typed-file-marker
```

felixdivo:
I guess that the failing test has nothing to do with the change.

zariiii9003:
Is the `setup.py` entry sufficient for it to be included in the source distribution?

felixdivo:
> Is the `setup.py` entry sufficient for it to be included in the source distribution?

Yes, that appears to work. I opened a `python:3` Docker image, produced a source distribution with `python setup.py sdist bdist_wheel` and then inspected the generated distribution file `dist/python-can-4.0.0.tar.gz`. It contains the marker file.

zariiii9003 (approved):

PR #1345, opened 2022-07-08
Add Parameter for Enabling PCAN Auto Bus-Off Reset

1atabey1:
Allows enabling of automatic hardware reinitialisation in case of a bus off event by adding the parameter `auto_reset` (defaults to `False` in order to not change current default behaviour)


1atabey1:
Hmm looks like some tests are failing for the UdpMulticastBus and to be very honest I have no clue how that bus works and how I could have broken anything there... Can someone with a bit more experience have a brief look and help a newbie out? :)

(I noticed that @felixdivo has done a lot of work on the virtual UDP bus, do you have any idea what's going wrong here?)

1atabey1:
Okay no worries, I'll see if I can make some sense out of it!

lumagi:
Hey @1atabey1, @felixdivo 
I have the same failure reports on my pull request. I don't believe it has anything to do with the pull requests because for my second commit the test passed just fine. I rather think it is a race condition in the test runner that only fails badly part of the time.

I did some investigation, the failing tests all do the following:
1. Send out CAN message on bus 1 with arbitration ID `i`
2. Receive message on bus 2 and assert that it is equal to the original message
3. Clear out the reception queue of bus 1 by receiving one message non-blocking. This operation is supposed to clear out the datagram that is looped back to the sending socket during step 1
4. Increment arbitration ID of the original message to `i + 1` and send it out on bus 2
5. Receive message on bus 1 and assert that it is equal to the original message with incremented arbitration ID `i + 1`

All tests fail because the arbitration ID of the message received during step 5 is still equal to `i` and not `i + 1`. The only reason I can think of is that the looped back datagram is not yet available in the reception buffer of bus 1 in step 3 and hence not cleared out. Then, during step 5 of the test, the code reads the message with arbitration ID `i` instead of `i + 1` and fails.

Unfortunately, I am unable to reproduce this bug locally. I will try to create a new pull request to debug this further.

can/interfaces/pcan/pcan.py
```diff
@@ -259,6 +264,14 @@ def __init__(
                     "Ignoring error. PCAN_ALLOW_ERROR_FRAMES is still unsupported by OSX Library PCANUSB v0.10"
                 )
 
+        if kwargs.get("auto_reset", False):
```
zariiii9003, line 267: Could you add the parameter to the `__init__()` method instead of using `kwargs.get`?
I do not know why some of the other parameters are not `__init__`-arguments but i thinks that's bad for readability.

1atabey1:
sorry, was too busy last month unfortunately.. Thank you two for finishing this up @zariiii9003 @lumagi !

PR #1346, opened 2022-07-08
Add device_id parameter to PcanBus constructor

lumagi:
This pull request adds support for constructing PCAN bus instances based on the device ID.

Before, it was only possible to select a certain bus based on the channel name (i.e. `PCAN_USBBUSx`). This can be troublesome since the order of the devices is not necessarily consistent across reboots or USB reconnects. With this patch, it is possible to choose the device based on its device ID. The device ID can be freely set through the API or PCAN-View.

Things to discuss before a merge:
* At the moment, the implementation tests all possible PCAN channels. A more suitable method which queries the driver for active channels was introduced in version 4.4.0. I refrained from using it because I believe this addition was fairly recent and using it would make the feature incompatible to older API versions.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -103,6 +103,7 @@ class PcanBus(BusABC):
     def __init__(
         self,
         channel="PCAN_USBBUS1",
+        device_id=None,
         state=BusState.ACTIVE,
         bitrate=500000,
         *args,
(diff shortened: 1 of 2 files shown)


lumagi:
@zariiii9003 Is there anything else you need me to do for the merge?

zariiii9003 (approved):
Looks good to me.

PR #1360, opened 2022-08-04
Fix the sized rotating logger size bug for the blf writer

### LINKED ISSUE #1359: Sized Rotating Logger bug for BLFWriter file_size (-s) argument
### Describe the bug

The `BaseRotatingLogger` performs a check to see if it [should_rollover(msg)](https://github.com/hardbyte/python-can/blob/b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35/can/io/logger.py#L204), which is an abstract method that invokes the `SizedRotatingLogger` function [should_rollover(msg)](https://github.com/hardbyte/python-can/blob/b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35/can/io/logger.py#L324-L331). Inside of the `should_rollover` function, the `writer.tell()` [function](https://docs.python.org/3/library/io.html#io.TextIOBase.tell) is utilized to see how large the file being written to is. The issue when the `writer` is of type [BLFWriter](https://github.com/hardbyte/python-can/blob/b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35/can/io/blf.py#L354) is that `writer.tell()` always returns a size of 144 bytes, until the buffer grows to be larger than the [max_container_size](https://github.com/hardbyte/python-can/blob/b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35/can/io/blf.py#L362) and the buffer is [flushed](https://github.com/hardbyte/python-can/blob/b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35/can/io/blf.py#L533-L534). Then the function `writer.tell()` returns something around 26,500 bytes (25.8 KiB). As long as the `-s` option is larger than the header (144 bytes), then the file written will always be around 25.8 KiB.  

### To Reproduce

Enable virtual can: `sudo ip link add dev vcan0 type vcan && sudo ip link set up vcan0`
Spam the virtual bus with random can messages 1 ms apart: `cangen vcan0 -e -I 18EAFF26 -L i -v -v -v -g 1`
Log the virtual bus: `python -m can.logger -c vcan0 -i socketcan -b 250000 -f file.blf -s 150`

### Expected behavior

I expected for a `blf` file of size 150 bytes to be written. 

### Additional context

OS and version: Ubuntu 20.04.4 LTS
Python version: Python 3.7.17
python-can version: b9d9d01c1c87bc2a4d08b709a0763cdc850f8c35

### Details

I have added print statements in the `io/blf.py` and `io/logger.py` modules to highlight the bug here: 977e1e0c6dc85b709eb16b6fea9a719144440603.

j-c-cook:
I see two options here:
  
  1. The `-s` option for the blf rolling logger will **not** define the file size to be written, but rather the buffer size prior to compression
  2. A mathematical equation is to be written that computes the _expected_ file size of the buffer based on the compression integer passed to [zlib.compress](https://docs.python.org/3/library/zlib.html#zlib.compress). (This would require either some knowledge of how the compression functions, or could be an equation fit made from a table of collected values based on a range of compressed buffer sizes vs. the input integer.)
 
I plan to move forward shortly with a commit that will implement the solution (1). 

closes #1359

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -531,7 +531,6 @@ class BLFWriter(FileIOMessageWriter):
         self._buffer_size += obj_size + padding_size
         self.object_count += 1
         if self._buffer_size >= self.max_container_size:
-            print("Flush the buffer to a file. Larger than max container!")
             self._flush()
 
     def _flush(self):
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -17,7 +17,7 @@ import zlib
 import datetime
 import time
 import logging
-from typing import List, BinaryIO, Generator, Union, Tuple, Optional, cast
+from typing import List, BinaryIO, Generator, Union, Tuple, Optional, cast, Any
 
 from ..message import Message
 from ..util import len2dlc, dlc2len, channel2int
@@ -358,9 +358,6 @@ class BLFWriter(FileIOMessageWriter):
 
     file: BinaryIO
 
-    #: Max log container size of uncompressed data
-    max_container_size = 128 * 1024
-
     #: Application identifier for the log writer
     application_id = 5
 
@@ -370,6 +367,7 @@ class BLFWriter(FileIOMessageWriter):
         append: bool = False,
         channel: int = 1,
         compression_level: int = -1,
+        **options: Any,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to write to
@@ -400,6 +398,11 @@ class BLFWriter(FileIOMessageWriter):
         self.compression_level = compression_level
         self._buffer: List[bytes] = []
         self._buffer_size = 0
+        #: Max log container size of uncompressed data
+        if "max_container_size" in options:
+            self.max_container_size = options["max_container_size"]
+        else:
+            self.max_container_size = 128 * 1024  # bytes
         if append:
             # Parse file header
             data = self.file.read(FILE_HEADER_STRUCT.size)
@@ -530,6 +533,7 @@ class BLFWriter(FileIOMessageWriter):
 
         self._buffer_size += obj_size + padding_size
         self.object_count += 1
+        a = len(zlib.compress(memoryview(b"".join(self._buffer)), 1))
         if self._buffer_size >= self.max_container_size:
             self._flush()
 
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -531,6 +531,7 @@ class BLFWriter(FileIOMessageWriter):
         self._buffer_size += obj_size + padding_size
         self.object_count += 1
         if self._buffer_size >= self.max_container_size:
+            print("Flush the buffer to a file. Larger than max container!")
             self._flush()
 
     def _flush(self):
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -533,7 +533,6 @@ class BLFWriter(FileIOMessageWriter):
 
         self._buffer_size += obj_size + padding_size
         self.object_count += 1
-        a = len(zlib.compress(memoryview(b"".join(self._buffer)), 1))
         if self._buffer_size >= self.max_container_size:
             self._flush()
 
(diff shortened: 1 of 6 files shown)


can/logger.py
```diff
@@ -215,6 +215,7 @@ def main() -> None:
 
     options = {"append": results.append}
     if results.file_size:
+        options["max_container_size"] = results.file_size  # bytes
```
j-c-cook, line 218: I'm now passing in `max_container_size` as an `option`.

can/io/blf.py
```diff
@@ -370,6 +367,7 @@ def __init__(
         append: bool = False,
         channel: int = 1,
         compression_level: int = -1,
+        **options: Any,
```
j-c-cook, line 370: Added `**options` to BLFWriter API.

can/io/blf.py
```diff
@@ -400,6 +398,11 @@ def __init__(
         self.compression_level = compression_level
         self._buffer: List[bytes] = []
         self._buffer_size = 0
+        #: Max log container size of uncompressed data
+        if "max_container_size" in options:
+            self.max_container_size = options["max_container_size"]
+        else:
+            self.max_container_size = 128 * 1024  # bytes
```
j-c-cook, line 405: There is now an adjustable max buffer size.

j-c-cook:
@zariiii9003 Ready for review.

can/logger.py
```diff
@@ -170,8 +170,9 @@ def main() -> None:
         "--file_size",
         dest="file_size",
         type=int,
-        help="Maximum file size in bytes. Rotate log file when size threshold "
-        "is reached.",
+        help="Maximum file size in bytes (or for the case of blf, maximum "
+        "buffer size before compression and flush to file). Rotate log "
+        "file when size threshold is reached.",
```
hardbyte, line 175: 👍🏼

can/io/blf.py
hardbyte: Assuming that `compression_level` is specific to the BLFWriter, perhaps it should be moved into the `options` dict as well?

hardbyte (approved):

zariiii9003:
@hardbyte i see you approved this but personally i think bringing implementation details of the BlfWriter (`options["max_container_size"] = results.file_size # bytes`) into the logger script is not a pretty solution.

What do you think about adding a `size` property to all FileIOMessageWriter subclasses? That way we could delegate size estimation to the logger implementations.

can/io/blf.py
j-c-cook: Arguments specific to a `Writer` could be made more accessible if able to be set by `**options` . Currently, the only way to set `compression_level` would be to bypass the `Logger` (or `RollingLogger`) object and call the `BLFWriter` object itself. 

When going about these changes, I'll want to make sure that the API is either unaffected by this improvement, or that a deprecation warning is added (similar to what I did [here](https://github.com/j-c-cook/python-can/blob/5c232483e97d92b002886ada109f4f6500ffec25/can/io/logger.py#L433-L439) for `SizedRotatingLogger`).

zariiii9003: It is specific to the BlfWriter (and soon Mf4Writer) but i think it's more readable when it is not hidden away in `**options`

j-c-cook: I have confirmed that `compression_level` can be modified by passing via the argument in `**kwargs`. @zariiii9003's CLI arguments https://github.com/zariiii9003/python-can/tree/extra_args2 will provide the ability to modify variables such as `compression_level`. For example:
```
python -m can.logger -i socketcan -c vcan0 -b 250000 -f file.blf -s 100000 --compression-level=1 --max-container-size=200000
```
Once https://github.com/hardbyte/python-can/pull/1366 is merged, I will close this PR. The combination of https://github.com/hardbyte/python-can/pull/1367 and https://github.com/hardbyte/python-can/pull/1366 will resolve everything I had intended from this PR.

PR #1361, opened 2022-08-06
Raise appropriate error message when append is not possible

### LINKED ISSUE #1318: Enhance the error message when requesting file format `.db` with sized rotating logger
### Describe the bug

The `can.logger` does not accept `.db` as a format to place the log in, but the API documentation states that `.db` is an acceptable format:
```
    The format is determined from the file suffix which can be one of:
      * .asc: :class:`can.ASCWriter`
      * .blf :class:`can.BLFWriter`
      * .csv: :class:`can.CSVWriter`
      * .db: :class:`can.SqliteWriter`
      * .log :class:`can.CanutilsLogWriter`
      * .txt :class:`can.Printer`
```

### To Reproduce

Enable virtual can: `sudo ip link set vcan0 up type can`
Spam the virtual bus with random can messages 1 ms apart: `cangen vcan0 -e -I 18EAFF26 -L i -v -v -v -g 1`
Log the virtual bus: `python -m can.logger -c vcan0 -i socketcan -b 250000 -f file.db -s 50`

### Expected behavior

The `can.logger` script should successfully log messages. 

### Additional context

OS and version: Ubuntu 20.04.4 LTS
Python version: Python 3.7.17
python-can version: 4.0.0

```python
  File "/home/agco/Documents/python-can/scripts/can_logger.py", line 11, in <module>
    main()
  File "/home/agco/Documents/python-can/can/logger.py", line 206, in main
    base_filename=results.log_file, max_bytes=results.file_size
  File "/home/agco/Documents/python-can/can/io/logger.py", line 322, in __init__
    self._writer = self._get_new_writer(self.base_filename)
  File "/home/agco/Documents/python-can/can/io/logger.py", line 230, in _get_new_writer
    "The Logger corresponding to the arguments is not a FileIOMessageWriter or "
Exception: The Logger corresponding to the arguments is not a FileIOMessageWriter or can.Printer
```
</details>

j-c-cook:
- Handle "append" by adding **options to SqlWriter 
- Check logger instance for SqliteWriter in rotating logger setup and raise appropriate error.

closes #1318

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -352,6 +352,7 @@ class ASCWriter(FileIOMessageWriter):
         self,
         file: Union[StringPathLike, TextIO],
         channel: int = 1,
+        **kwargs,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to write to
@@ -360,6 +361,13 @@ class ASCWriter(FileIOMessageWriter):
         :param channel: a default channel to use when the message does not
                         have a channel set
         """
+        if "append" in kwargs:
+            if kwargs["append"]:
+                raise Exception(
+                    f"{self.__class__.__name__} is currently not "
+                    f"equipped to append messages to an existing "
+                    f"file."
+                )
         super().__init__(file, mode="w")
 
         self.channel = channel
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -352,7 +352,7 @@ class ASCWriter(FileIOMessageWriter):
         self,
         file: Union[StringPathLike, TextIO],
         channel: int = 1,
-        **kwargs,
+        **kwargs: Any,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to write to
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -362,7 +362,7 @@ class ASCWriter(FileIOMessageWriter):
                         have a channel set
         """
         if kwargs.get("append", False):
-            raise TypeError(
+            raise ValueError(
                 f"{self.__class__.__name__} is currently not equipped to "
                 f"append messages to an existing file."
             )
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -363,7 +363,7 @@ class ASCWriter(FileIOMessageWriter):
         """
         if "append" in kwargs:
             if kwargs["append"]:
-                raise Exception(
+                raise TypeError(
                     f"{self.__class__.__name__} is currently not "
                     f"equipped to append messages to an existing "
                     f"file."
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -349,7 +349,10 @@ class ASCWriter(FileIOMessageWriter):
     FORMAT_EVENT = "{timestamp: 9.6f} {message}\n"
 
     def __init__(
-        self, file: Union[StringPathLike, TextIO], channel: int = 1, **kwargs: Any,
+        self,
+        file: Union[StringPathLike, TextIO],
+        channel: int = 1,
+        **kwargs: Any,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to write to
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -349,10 +349,7 @@ class ASCWriter(FileIOMessageWriter):
     FORMAT_EVENT = "{timestamp: 9.6f} {message}\n"
 
     def __init__(
-        self,
-        file: Union[StringPathLike, TextIO],
-        channel: int = 1,
-        **kwargs: Any,
+        self, file: Union[StringPathLike, TextIO], channel: int = 1, **kwargs: Any,
     ) -> None:
         """
         :param file: a path-like object or as file-like object to write to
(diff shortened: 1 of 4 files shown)


j-c-cook:
@zariiii9003 ready for review.

can/io/sqlite.py
```diff
@@ -128,7 +128,9 @@ class SqliteWriter(MessageWriter, BufferedReader):
     MAX_BUFFER_SIZE_BEFORE_WRITES = 500
     """Maximum number of messages to buffer before writing to the database"""
 
-    def __init__(self, file: StringPathLike, table_name: str = "messages") -> None:
+    def __init__(
+        self, file: StringPathLike, table_name: str = "messages", **options: Any
```
zariiii9003, line 132:
```suggestion
        self, file: StringPathLike, table_name: str = "messages", **kwargs: Any
```

This is usually called `kwargs` in python projects

can/io/logger.py
```diff
@@ -225,6 +226,11 @@ def _get_new_writer(self, filename: StringPathLike) -> FileIOMessageWriter:
...
         elif isinstance(logger, Printer) and logger.file is not None:
             return cast(FileIOMessageWriter, logger)
+        elif isinstance(logger, can.io.SqliteWriter):
+            raise Exception(
+                "The SqliteWriter is not available in conjunction"
+                "with the sized rotating logger."
+            )
         else:
             raise Exception(
                 "The Logger corresponding to the arguments is not a FileIOMessageWriter or "
```
zariiii9003, line 236: You could improve the existing error message without the `isinstance(logger, can.io.SqliteWriter)` check. You can access the logger class name with `logger.__class__.__name__`, for example inside a f-string.
This way you do not need special cases for any new unsupported loggers

zariiii9003 (changes requested):

can/io/sqlite.py
j-c-cook: > Also: Shouldn't there be an exception when the user demands "append" but the logger does not support it?

I think this is brilliant.

j-c-cook:
@zariiii9003 Thank you for that feedback. I added in the "append" error handling to `ASCWriter` and `SQLWriter` and adjusted the exception based on your [comment](https://github.com/hardbyte/python-can/pull/1361#discussion_r939557531).

can/io/asc.py
```diff
@@ -360,6 +361,13 @@ def __init__(
         :param channel: a default channel to use when the message does not
                         have a channel set
         """
+        if "append" in kwargs:
+            if kwargs["append"]:
+                raise TypeError(
```
j-c-cook, line 366: I'm not sure if this is the best kind of error to use. Maybe a `ValueError`?

zariiii9003: Yes, I'd say ValueError, too

can/io/logger.py
```diff
@@ -227,8 +228,8 @@ def _get_new_writer(self, filename: StringPathLike) -> FileIOMessageWriter:
             return cast(FileIOMessageWriter, logger)
         else:
             raise Exception(
-                "The Logger corresponding to the arguments is not a FileIOMessageWriter or "
-                "can.Printer"
+                f"The Logger, {logger.__class__.__name__}, corresponding to "
```
zariiii9003, line 231: The previous error message wasn't great. We should use simple language.

```python3
raise Exception(
    f"The log format \"{''.join(pathlib.Path(filename).suffixes[-2:])}\" "
    f"is not supported by {self.__class__.__name__}"
)
```
This results in `Exception: The log format ".db" is not supported by SizedRotatingLogger`.
Works for all subclasses, too. We could also point the user to the documentation but that might be a bit redundant.

can/io/asc.py
```diff
@@ -360,6 +361,13 @@ def __init__(
         :param channel: a default channel to use when the message does not
                         have a channel set
         """
+        if "append" in kwargs:
+            if kwargs["append"]:
```
zariiii9003, line 365: you could remove this line if you use `if kwargs.get("append", False):`

can/io/asc.py
```diff
@@ -361,13 +358,11 @@ def __init__(
...
-                raise TypeError(
-                    f"{self.__class__.__name__} is currently not "
-                    f"equipped to append messages to an existing "
-                    f"file."
-                )
+        if kwargs.get("append", False):
+            raise TypeError(
+                f"{self.__class__.__name__} is currently not equipped to "
+                f"append messages to an existing file."
+            )
```
j-c-cook, line 368: Took your advice here.

can/io/logger.py
```diff
@@ -228,7 +228,7 @@ def _get_new_writer(self, filename: StringPathLike) -> FileIOMessageWriter:
             return cast(FileIOMessageWriter, logger)
         else:
             raise Exception(
-                f"The log format \".{filename.split('.')[-1]}\" is not "
+                f"The log format \"{os.path.splitext(filename)[-1]}\" is not "
```
j-c-cook, line 231: I used your recommended error message, but slightly adjusted how the extension is found.

zariiii9003: This does not work for multiple suffixes like `file.asc.gz`

j-c-cook: I fixed it with your version.

j-c-cook:
@zariiii9003 I think this might be ready to go in. Could you give it another look?

zariiii9003 (approved):
Thank you, looks good

PR #1362, opened 2022-08-06
Write 3 ms digits to ascii to resolve CANoe bug

### LINKED ISSUE #1315: Logged .asc file can't open in CANoe 14
### Describe the bug
Logged .asc file by python-can is not able to open in CANoe 14

### To Reproduce
- Record .asc file by python-can
- Try to replay logged file by CANoe 14

### Expected behavior
Can replay log by CANoe 14

### Additional context

OS and version: Window 10
Python version: 3.8
python-can version: 4.0.0(https://github.com/hardbyte/python-can/releases/download/4.0.0/python_can-4.0.0-py3-none-any.whl)

### Root cause
Date information in header of .asc file contain 6 digits of milisecond and somehow violate CANoe 14 rule. 3 digits is accepted
Not work: date Mon May 23 03:29:03.103**xxx** PM 2022
Worked: date Mon May 23 03:29:03.103 PM 2022

### Propose fix
Use same logic to get datetime when write begin trigger block
File [asc.py, line 368-369](https://github.com/hardbyte/python-can/blob/796b52586a5d891e4e2077b423de40cba07f8b44/can/io/asc.py#L368)
Change from
```python
        now = datetime.now().strftime(self.FORMAT_START_OF_FILE_DATE)
        self.file.write(f"date {now}\n")
```
to
```python
        now = datetime.now().timestamp()
        mlsec = repr(now).split(".")[1][:-3]
        formatted_date = time.strftime(
            self.FORMAT_DATE.format(mlsec), time.localtime(now)
        )
        self.file.write(f"date {formatted_date}\n")
```
Tested and can open in CANoe

j-c-cook:
- The ascii writer will now only write 3 micro second decimal places to a log file. 
- The ascii format currently is not setup to append. I recenly added in **options to the logger script. **options are only required for
the rolling logger.

closes #1315
closes #1357

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -344,7 +344,6 @@ class ASCWriter(FileIOMessageWriter):
             "{bit_timing_conf_ext_data:>8}",
         ]
     )
-    # Use trigger start time to replace file start time
     FORMAT_START_OF_FILE_DATE = "%a %b %d %I:%M:%S.%f %p %Y"
     FORMAT_DATE = "%a %b %d %I:%M:%S.{} %p %Y"
     FORMAT_EVENT = "{timestamp: 9.6f} {message}\n"
@@ -411,10 +410,6 @@ class ASCWriter(FileIOMessageWriter):
             formatted_date = time.strftime(
                 self.FORMAT_DATE.format(mlsec), time.localtime(self.last_timestamp)
             )
-            # changed 2022-08-04, moved file start header here to be one.
-            self.file.write(f"date {formatted_date}\n")
-            self.file.write("base hex  timestamps absolute\n")
-            self.file.write("internal events logged\n")
             self.file.write(f"Begin Triggerblock {formatted_date}\n")
             self.header_written = True
             self.log_event("Start of measurement")  # caution: this is a recursive call!
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -346,6 +346,8 @@ class ASCWriter(FileIOMessageWriter):
     )
     # Use trigger start time to replace file start time
     FORMAT_START_OF_FILE_DATE = "%a %b %d %I:%M:%S.%f %p %Y"
+    FORMAT_DATE = "%a %b %d %I:%M:%S.{} %p %Y"
+    FORMAT_EVENT = "{timestamp: 9.6f} {message}\n"
 
     def __init__(
         self,
@@ -366,9 +368,9 @@ class ASCWriter(FileIOMessageWriter):
         # write start of file header
         now = datetime.now().strftime(self.FORMAT_START_OF_FILE_DATE)
         # Note: CANoe requires that the microsecond field only have 3 digits
-        idx = now.index('.')  # Find the index in the string of the decimal
+        idx = now.index(".")  # Find the index in the string of the decimal
         # Keep decimal and first three ms digits (4), remove remaining digits
-        now = now.replace(now[idx+4:now[idx:].index(' ') + idx], '')
+        now = now.replace(now[idx + 4 : now[idx:].index(" ") + idx], "")
         self.file.write(f"date {now}\n")
         self.file.write("base hex  timestamps absolute\n")
         self.file.write("internal events logged\n")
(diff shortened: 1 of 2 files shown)


j-c-cook:
@zariiii9003 ready for review.

can/logger.py
```diff
@@ -213,13 +213,13 @@ def main() -> None:
...
 
-    options = {"append": results.append}
     if results.file_size:
+        options = {"append": results.append}
         logger = SizedRotatingLogger(
             base_filename=results.log_file, max_bytes=results.file_size, **options
         )
     else:
-        logger = Logger(filename=results.log_file, **options)  # type: ignore
+        logger = Logger(filename=results.log_file)  # type: ignore
```
j-c-cook, line 222: This change won't be necessary once https://github.com/hardbyte/python-can/pull/1361 is merged.

can/io/asc.py
```diff
@@ -400,6 +405,10 @@ def log_event(self, message: str, timestamp: Optional[float] = None) -> None:
             formatted_date = time.strftime(
                 self.FORMAT_DATE.format(mlsec), time.localtime(self.last_timestamp)
             )
+            # changed 2022-08-04, moved file start header here to be one.
```
zariiii9003, line 408: Why?

can/logger.py
```diff
@@ -213,13 +213,13 @@ def main() -> None:
     print(f"Connected to {bus.__class__.__name__}: {bus.channel_info}")
     print(f"Can Logger (Started on {datetime.now()})")
 
-    options = {"append": results.append}
     if results.file_size:
+        options = {"append": results.append}
```
zariiii9003, line 217: This seems out of scope for a ASC related PR

j-c-cook: Yes it is. There was an error when `-a` was used without `-s`, so this was a quick fix for that. #1361 resolved the issue. I need to revert the commit that introduced this.

can/io/asc.py
j-c-cook: This appears to be a carry-over from when I `cherry-pick`ed from https://github.com/hardbyte/python-can/pull/1357/commits/ff7581bd20e5c8300a1d6ca64789067ac5387017. It looks like these lines need to be removed.

j-c-cook:
@zariiii9003 I believe this accomplishes what was described in #1315.

j-c-cook:
At a glance, I'm not sure how the test failure of `mypy --python-version 3.8 .` has to do with these minor changes. The `datetime` documentation says [strftime](https://docs.python.org/3/library/datetime.html#datetime.date.strftime) returns a string.

zariiii9003:
> At a glance, I'm not sure how the test failure of `mypy --python-version 3.8 .` has to do with these minor changes. The `datetime` documentation says [strftime](https://docs.python.org/3/library/datetime.html#datetime.date.strftime) returns a string, and I have operated on it with string compatible functions (`index` and `replace`).

I just ran into this issue, too. It's weird because nothing changed. I'll try to fix it.

zariiii9003:
Should be fixed in develop.

Tian-Jionglu:
@j-c-cook as mentioned in #1357, script `can_logconvert.py` makes logged datetime in converted asc file as **converting datetime**. Does anyone else also think it is weird?

zariiii9003:
> @j-c-cook as mentioned in #1357, script `can_logconvert.py` makes logged datetime in converted asc file as **converting datetime**. Does anyone else also think it is weird?

I'm not sure what you mean, could you give an example? Or attach the files maybe?

Tian-Jionglu:
See logs header below:

- ascii logged by python-can can_logger.py
```
date Thu Aug 04 03:17:23.124 PM 2022
base hex  timestamps absolute
internal events logged
Begin Triggerblock Thu Aug 04 03:17:23.120 PM 2022
```
- ascii logged by CANalyzer
```
date Mon Aug 15 09:32:44.632 am 2022
base hex  timestamps absolute
internal events logged
// version 11.0.0
Begin Triggerblock Mon Aug 15 09:32:44.632 am 2022
```
- ascii converted by python-can can_logconvert.py
```
date Sat Aug 13 09:09:58.794 AM 2022
base hex  timestamps absolute
internal events logged
Begin Triggerblock Thu Aug 04 12:44:31.611 PM 2022
```
We can see that in log from CANalyzer, log datetime is the same as trigger beginning datetime.
But in python-can these two datetimes are different, this differs a lot when using `logconvert`.

Tian-Jionglu:
@zariiii9003, what's your opinion?

zariiii9003:
I see what you mean now. Technically the current implementation is correct, The header tells when the file was created and the triggerblock tells the timestamp of the first message.

But i'd like to hear other opinions on this.

PR #1365, opened 2022-08-07
Generalize the RotatingLogger; size, time or first of the two trigger the rollover

### LINKED ISSUE #1364: TimedRotatingLogger

The existing `SizedRotatingLogger` can be used to frequently (or infrequently) write CAN traces to a file by specifying a maximum file size with the `-s` option. The frequency can become periodic if the number of CAN traces being sent on the bus are _nearly_ consistent or the dominant message(s) are periodic. My specific application involves a remote logger running embedded Linux with a cellular connection that transports the rolled over logged files via secure shell protocol. I have been trying to balance the number of files produced vs. total file size. More importantly thus far, I must limit the number of file transfers from a specific device due to the firewall settings on the server where the files are being transported to. Therefore, my application would be well suited for a `TimedRotatingLogger`, where I can define the amount of time in between rollovers. 

The `TimedRotatingLogger` would rollover after the specified amount of time as long as there were CAN messages observed in that time. If no CAN messages were sent on the bus in that time, an empty trace should not be written. 

At this point in time I plan to focus on the implementation of a `TimedRotatingLogger`, but I wonder if both options should be able to be given. For example, if a user supplies a time and file size constraint, the first to occur (timeout or size limit) would trigger the rollover. 

### Additional context

The idea of the `TimedRotatingLogger` was (to my knowledge) originally introduced in #1359 by @zariiii9003 in https://github.com/hardbyte/python-can/issues/1359#issuecomment-1207256273.

j-c-cook:
A `RotatingLogger` object is created that constrains the file by file size, time or both. In the case of when both are constrained, the first that occurs will trigger the rollover. 

- Adds a deprecation warning to the TimedRotatingLogger. The TimedRotatingLogger now inherits from the generalized RotatingLogger.

- [ ] When time is a constraint for the `RotatingLogger`, don't rollover if there have been no messages received since the last rollover
- [ ] Update the doc string under `RotatingLogger`

closes #1364

### DIFF
--- a/can/__init__.py
+++ b/can/__init__.py
@@ -35,7 +35,14 @@ from . import interface
 from .interface import Bus, detect_available_configs
 from .bit_timing import BitTiming
 
-from .io import Logger, SizedRotatingLogger, Printer, LogReader, MessageSync
+from .io import (
+    Logger,
+    SizedRotatingLogger,
+    Printer,
+    LogReader,
+    MessageSync,
+    RotatingLogger,
+)
 from .io import ASCWriter, ASCReader
 from .io import BLFReader, BLFWriter
 from .io import CanutilsLogReader, CanutilsLogWriter
--- a/can/io/logger.py
+++ b/can/io/logger.py
@@ -6,6 +6,8 @@ import os
 import pathlib
 from abc import ABC, abstractmethod
 from datetime import datetime
+import time
+import warnings
 import gzip
 from typing import Any, Optional, Callable, Type, Tuple, cast, Dict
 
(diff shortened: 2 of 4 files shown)


j-c-cook:
The `BLFWriter` will be able to be used like the following:
```
python -m can.logger -i socketcan -c vcan0 -b 250000 -f file.blf -t 600 --compression-level=1 --max-container-size=200000
```
Here, a rollover would occur every 5 seconds, or when the buffer and file size becomes 200000. Note that the default `max_container_size` is 128 * 1024. 

Another option would be:
```
python -m can.logger -i socketcan -c vcan0 -b 250000 -f file.log -t 600 -s 100000
```
A rollover would occur every 5 seconds, or when the file size becomes 100000 bytes.

j-c-cook:
> * [x]  When time is a constraint for the `RotatingLogger`, don't rollover if there have been no messages received since the last rollover

This is inherit to the implementation of logger. When a message is received then the logger is called. The file cannot rollover without there being a message passed to logger. 
https://github.com/hardbyte/python-can/blob/88fc8e5bcf56ab2f9f2041242733ad90bb7101b4/can/logger.py#L246-L251

can/__init__.py
```diff
@@ -35,7 +35,14 @@
...
 
-from .io import Logger, SizedRotatingLogger, Printer, LogReader, MessageSync
+from .io import (
+    Logger,
+    SizedRotatingLogger,
+    Printer,
+    LogReader,
+    MessageSync,
+    RotatingLogger,
+)
```
j-c-cook, line 45: This was a `black` modification.

can/logger.py
```diff
@@ -185,6 +185,17 @@ def main() -> None:
...
     )
 
+    parser.add_argument(
+        "-t",
+        "--file_time",
+        dest="file_time",
+        type=int,
+        help="Maximum period in seconds before rotating log file. (If file_size"
+        "is also given, then the first of the two constraints to occur is"
+        "what causes the rollover.)",
```
felixdivo, line 195: Is this right?

```suggestion
        help="Maximum period in seconds before rotating log file. (If file_size "
        "is also given, then the first of the two constraints to occur is "
        "what causes the rollover.)",
```

can/io/logger.py
```diff
@@ -275,18 +277,19 @@ def do_rollover(self) -> None:
...
     created and written to.
 
-    This behavior can be customized by setting the
-    :attr:`~can.io.BaseRotatingLogger.namer` and
-    :attr:`~can.io.BaseRotatingLogger.rotator`
-    attribute.
+    This behavior can be customized by setting the :attr:`namer` and
+    :attr:`rotator` attribute.
 
     Example::
```
felixdivo, line 294: You also have to adjust the example.

felixdivo (changes requested):
Regarding the discussion in #1364: I generally think that separating what can be split is a fair goal. In this case, however, I think having one class for this makes a lot of sense since it does simplify both user code and the implementation (I think having cooperating sub-classes to have a logger that supports both variants of rotations will not be easier to understand than this).

There are also a few nice small improvements in here that should be merged regardless of the new features!

We should finish the talk about the general API in #1364 before merging here.

can/logger.py
j-c-cook: Yes, this is true of the current implementation. The one exception being that if there is no CAN messages sent on the bus, then the `rollover` condition is not checked for. Therefore, if the CAN bus is inactive and the time limit is passed, a rollover will only occur after a message is received. At first I thought this was problematic, but I'm not sure if there is a benefit of creating empty log files. However, if that is a desired feature, then something would need to change.

can/io/logger.py
j-c-cook: I've updated the example now, thank you.

j-c-cook: @felixdivo Please note that the attributes references are restored to the `BaseRotatingLogger` in https://github.com/hardbyte/python-can/pull/1365/commits/2862573c09a2c7b2271169151642af9dcdfb6c27 due to a failure in the `docs` workflow build.

can/logger.py
j-c-cook: This is also how the TimedRotatingLogger is implemented. 
image

felixdivo: Ah sorry, I meant to ask whether my suggestion to add spaces is right. I could've been clearer. 😅

But thank you for your explanation. I also think that the current implementation is fine in that regard!

PR #1369, opened 2022-08-10
socketcan: Make find_available_interfaces() find slcanX interfaces

msalau:
Hello,

This change allows `can.detect_available_configs(['socketcan'])` find not just `canX` and `vcanX` interfaces, but also `slcanX` interfaces too. Such names are given by default to interfaces that are created using the `slcand` utility.

The module has an option to work with SLCAN adapters directly, but creating a SocketCAN interface instead can be beneficial too (e.g.: make it possible to use Wireshark, can-utils, cantools and etc.). For this reason I added few lines to the doc about how to use SLCAN adapters with socketcan.

Regards,
Maksim

### DIFF
--- a/can/interfaces/socketcan/utils.py
+++ b/can/interfaces/socketcan/utils.py
@@ -38,7 +38,7 @@ def pack_filters(can_filters: Optional[typechecking.CanFilters] = None) -> bytes
     return struct.pack(can_filter_fmt, *filter_data)
 
 
-_PATTERN_CAN_INTERFACE = re.compile(r"(sl|v)?can\d+")
+_PATTERN_CAN_INTERFACE = re.compile(r"(sl|v|vx)?can\d+")
 
 
 def find_available_interfaces() -> Iterable[str]:
--- a/can/interfaces/socketcan/utils.py
+++ b/can/interfaces/socketcan/utils.py
@@ -38,7 +38,7 @@ def pack_filters(can_filters: Optional[typechecking.CanFilters] = None) -> bytes
     return struct.pack(can_filter_fmt, *filter_data)
 
 
-_PATTERN_CAN_INTERFACE = re.compile(r"v?can\d+")
+_PATTERN_CAN_INTERFACE = re.compile(r"(sl|v)?can\d+")
 
 
 def find_available_interfaces() -> Iterable[str]:
--- a/doc/interfaces/socketcan.rst
+++ b/doc/interfaces/socketcan.rst
@@ -57,6 +57,28 @@ existing ``can0`` interface with a bitrate of 1MB:
 
     sudo ip link set can0 up type can bitrate 1000000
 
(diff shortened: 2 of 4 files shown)


can/interfaces/socketcan/utils.py
```diff
@@ -38,7 +38,7 @@ def pack_filters(can_filters: Optional[typechecking.CanFilters] = None) -> bytes
     return struct.pack(can_filter_fmt, *filter_data)
 
 
-_PATTERN_CAN_INTERFACE = re.compile(r"v?can\d+")
+_PATTERN_CAN_INTERFACE = re.compile(r"(sl|v)?can\d+")
```
hartkopp, line 41: Good idea! The slcan driver has been reworked for Linux 5.20 and will therefore show up as a 'normal' CAN interface (e.g. named 'can0') in the future.
But for the current setups the detection of 'slcan0' would be useful.
Btw. when updating the pattern it would make sense to take care of vxcan0 too, which is a special virtual CAN interface (pair) which allows to communicate over network namespace boundaries (analogue to veth), e.g. for setups in Docker containers, etc.

So it should be `re.compile(r"(sl|v|vx)?can\d+")`, right?

msalau: @hartkopp, thanks for feedback!
Sure, I'll update the pattern and the test code.

BTW while experimenting with vcan I discovered that an interface name can be almost anything (even `foo,bar-baz` is accepted):
```
sudo ip link add dev foo,bar-baz type vcan
sudo ip link set up foo,bar-baz
```

I'm confused by the failed test.
The issue seems to be unrelated to the changes in this pull-request.
```
collecting ... ERROR: InvocationError for command 'D:\a\python-can\python-can\.tox\gh\Scripts\pytest.EXE' (exited with code 3221225477)
___________________________________ summary ___________________________________
ERROR:   gh: commands failed
```
Do I have to do something with it?

hartkopp: Yes, it is a normal behaviour that you can name a network device whatever you like.
You can also rename them after creation.
When there is no name provided (e.g. by ìp link add type vcan`) the default `vcan%d` is taken from the vcan driver.

msalau: Done.

msalau:
@hartkopp @zariiii9003 
Please approve CI so the review can move on eventually :)
All the tests passed in my forked repository (only codecov upload failed due to missing token).

Thanks,
Maksim

hartkopp:
> @hartkopp @zariiii9003 Please approve CI so the review can move on eventually :) All the tests passed in my forked repository (only codecov upload failed due to missing token).
> 
> Thanks, Maksim

Looks good to me!
Reviewed-by: Oliver Hartkopp <socketcan@hartkopp.net>

PR #1384, opened 2022-08-27
Pass file mode to compress function

j-c-cook:


zariiii9003 (approved):
I'm not really familiar with the compress function, but it looks okay to me.

PR #1396, opened 2022-09-14
Test load_config()

### LINKED ISSUE #345: Add tests for load_config()

nertpinx:
Few patches for things I found when trying to add tests for `load_config()`.

Closes #345

### DIFF
--- a/test/listener_test.py
+++ b/test/listener_test.py
@@ -15,9 +15,6 @@ import can
 
 from .data.example_data import generate_message
 
-channel = "virtual_channel_0"
-can.rc["interface"] = "virtual"
-
 logging.basicConfig(level=logging.DEBUG)
 
 # makes the random number generator deterministic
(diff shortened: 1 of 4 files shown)


nertpinx:
I marked this as a draft since I am not sure this is the proper way of contributing. Let me know if any changes are needed.  Thanks.

setup.cfg
```diff
@@ -1,5 +1,5 @@
 [metadata]
-license_file = LICENSE.txt
+license_files = LICENSE.txt
```
j-c-cook, line 2: Why?

j-c-cook:
> I marked this as a draft since I am not sure this is the proper way of contributing. Let me know if any changes are needed. Thanks.

Your contribution workflow looks good. Once you think this is ready to go in, mark it as ready for review. 

You’ll need to format your code with the `black` package. First install the linters requirements, `pip install -r requirements-lint.txt` and then run `black —verbose .` in the root directory of `python-can`. This should auto format the 3 files that are causing the formatting check to break.

setup.cfg
nertpinx: It is explained in the commit message which is not immediately seen in GitHub, I can split it into a separate PR, just like the other unrelated patches.

https://github.com/hardbyte/python-can/pull/1396/commits/8e9b8fc49a0f2706b2553981ee08ec9c25258244

j-c-cook: It is good to keep PR’s within the narrowest scope as practical. Perhaps a quote by Einstein somewhat applies; “Everything should be made as simple as possible, but not simpler.” The review process is simpler (and often can occur quicker) if changes fall within a scope described in the related issue.

nertpinx: OK, I removed the other patches and will post them separately.

hardbyte (approved):
Thanks for adding these tests (and your contributions across other PRs)

Issue #1207, opened 2022-01-09
BufferedReader returns oldest messages first


trey-sygnalauto:
### Describe the bug
BufferedReader.get is documented as returning the latest message: "Attempts to retrieve the latest message received by the instance.".  In reality BufferedReader.get returns the oldest message in it's queue.  This is because the queue is implemented using a SimpleQueue which is a FIFO type queue.  

### To Reproduce
Receive CAN message and compare the order read using BufferedReader.get to the output of something like candump.

### Expected behavior
BufferedReader.get should always return the latest message received.  If no message has been received, the most recent item from the queue should be returned.

### Additional context

OS and version: 
Python version: 3.8
python-can version: 3.3.4
python-can interface/s (if applicable):

I implemented a basic LIFO implementation using LIFO queue.  Ultimately I believe a circular LIFO queue would be the best implementation for BufferedReader.  My very basic LIFO queue implementation is shown below.  Note the hacky way I allow new messages when the queue is full by creating a new queue and overwriting the old one.

class LIFOReader(Listener):
    """
    A BufferedReader is a subclass of :class:`~can.Listener` which implements a
    **message buffer**: that is, when the :class:`can.BufferedReader` instance is
    notified of a new message it pushes it into a queue of messages waiting to
    be serviced. The messages can then be fetched with
    :meth:`~can.BufferedReader.get_message`.

    Putting in messages after :meth:`~can.BufferedReader.stop` has be called will raise
    an exception, see :meth:`~can.BufferedReader.on_message_received`.

    :attr bool is_stopped: ``True`` iff the reader has been stopped
    """

    def __init__(self, maxsize = 128):
        # set to "infinite" size
        self.buffer = LifoQueue(maxsize=maxsize)
        self.is_stopped = False

    def on_message_received(self, msg):
        """Append a message to the buffer.

        :raises: BufferError
            if the reader has already been stopped
        """
        if self.is_stopped:
            raise RuntimeError("reader has already been stopped")
        else:
            try:
                self.buffer.put(msg)
            except Full:
                with self.buffer.mutex:
                    self.buffer = LifoQueue(maxsize=self.buffer.maxsize)

    def get_message(self, timeout=0.5):
        """
        Attempts to retrieve the latest message received by the instance. If no message is
        available it blocks for given timeout or until a message is received, or else
        returns None (whichever is shorter). This method does not block after
        :meth:`can.BufferedReader.stop` has been called.

[...]

PR #1397, opened 2022-09-15
Update BufferedReader.get_message docstring

barbacbd:
** Docstring now states that the `get_message` method utilizes a FIFO ordering for grabbing messages from the queue.

#1207

### DIFF
--- a/can/listener.py
+++ b/can/listener.py
@@ -105,13 +105,13 @@ class BufferedReader(Listener):  # pylint: disable=abstract-method
 
     def get_message(self, timeout: float = 0.5) -> Optional[Message]:
         """
-        Attempts to retrieve the latest message received by the instance. If no message is
-        available it blocks for given timeout or until a message is received, or else
-        returns None (whichever is shorter). This method does not block after
-        :meth:`can.BufferedReader.stop` has been called.
+        Attempts to retrieve the message that has been in the queue for the longest amount
+        of time (FIFO). If no message is available, it blocks for given timeout or until a
+        message is received (whichever is shorter), or else returns None. This method does
+        not block after :meth:`can.BufferedReader.stop` has been called.
 
         :param timeout: The number of seconds to wait for a new message.
-        :return: the Message if there is one, or None if there is not.
+        :return: the received :class:`can.Message` or `None`, if the queue is empty.
         """
         try:
             if self.is_stopped:


can/listener.py
```diff
@@ -105,13 +105,14 @@ def on_message_received(self, msg: Message) -> None:
 
     def get_message(self, timeout: float = 0.5) -> Optional[Message]:
         """
-        Attempts to retrieve the latest message received by the instance. If no message is
-        available it blocks for given timeout or until a message is received, or else
-        returns None (whichever is shorter). This method does not block after
+        Attempts to retrieve the message that has been in the queue for the longest amount
+        of time (FIFO). If no message is available, it blocks for given timeout or until a
+        message is received, or else returns None (whichever is shorter). This method does
```
nertpinx, line 110: I think it would be clearer to say "... until a message is recieved (whichever is shorter), or else returns None." otherwise it sounds like the it should return None because it is the shortest thing to do :smile: .

nertpinx (changes requested):
I would just change the one mentioned thing, otherwise it looks fine to me.

can/listener.py
```diff
@@ -105,13 +105,14 @@ def on_message_received(self, msg: Message) -> None:
 
     def get_message(self, timeout: float = 0.5) -> Optional[Message]:
         """
-        Attempts to retrieve the latest message received by the instance. If no message is
-        available it blocks for given timeout or until a message is received, or else
-        returns None (whichever is shorter). This method does not block after
+        Attempts to retrieve the message that has been in the queue for the longest amount
+        of time (FIFO). If no message is available, it blocks for given timeout or until a
+        message is received (whichever is shorter), or else returns None. This method does
+        not block after
```
zariiii9003, line 111: Here is an unnecessary line break. Otherwise this is good,

can/listener.py
```diff
@@ -105,13 +105,14 @@ def on_message_received(self, msg: Message) -> None:
...
         :meth:`can.BufferedReader.stop` has been called.
 
         :param timeout: The number of seconds to wait for a new message.
-        :return: the Message if there is one, or None if there is not.
+        :return: the Message if one exists, otherwise None.
```
zariiii9003, line 115:
```suggestion
        :return: the received :class:`can.Message` or `None`, if the queue is empty.
```
Not sure about the comma placement though.

zariiii9003 (changes requested):

zariiii9003 (approved):
Thank you!

Issue #1378, opened 2022-08-22
GzipFile BLFWriter negative seek error


j-c-cook:
### Describe the bug

A [GZipFile](https://github.com/python/cpython/blob/b69ba3d7233013ea8cb95a3e30e2338f7f76b359/Lib/gzip.py#L121) object implementation in `cpython` is unable to be sought negatively while in write mode. This is problematic for the `BLFWriter`. Upon initialization, the `BLFWriter` writes a default header to the file, in which the start and stop timestamps are not set. Once the `BLFWriter` is told to [stop](https://github.com/hardbyte/python-can/blob/2da28c1a1c87776618a60218b0b97800cf2deb34/can/io/blf.py#L583-L584), the remaining data in the buffer is flushed to the file and then the header must be updated with the start and stop timestamps. This is where the `GZipFile` object has problems. The location of the writer must be moved to the 0'th position of the file. When in write mode, the offset position [cannot be negatively sought](https://github.com/python/cpython/blob/b69ba3d7233013ea8cb95a3e30e2338f7f76b359/Lib/gzip.py#L384-L385).

##### Socket CAN setup

Enable CAN interface:

`sudo ip link add dev vcan0 type vcan && sudo ip link set up vcan0`

Generate random CAN messages 1 millisecond apart:

`cangen vcan0 -e -I 18EAFF26 -L i -v -v -v -g 1`

##### python-can setup

Enable the `Logger`:

`python -m can.logger -c vcan0 -i socketcan -b 250000 -f file.blf.gz`

The issue can be seen with the `SizedRotatingLogger` as well.

`python -m can.logger -c vcan0 -i socketcan -b 250000 -f file.blf.gz -s 150000`

### Expected behavior

A `file.blf.gz` should be created without error.

### Additional context

OS and version: Kubuntu 22.04
Python version: 3.7.13
python-can version: `pip install git+https://github.com/hardbyte/python-can@2da28c1a1c87776618a60218b0b97800cf2deb34`
python-can interface/s (if applicable):

<details><summary>Traceback and logs</summary>

`Logger` trace back:

```python
(pycan) jackcook@rollo:~$ python -m can.logger -c vcan0 -i socketcan -b 250000 -f file.blf.gz
Connected to SocketcanBus: socketcan channel 'vcan0'
Can Logger (Started on 2022-08-27 10:59:58.127917)
^CTraceback (most recent call last):
[...]

PR #1401, opened 2022-09-29
Modify `file_size` help doc string

j-c-cook:
In hindsight, I'm not sure that the statement, "(or for the case of blf, maximum buffer size before compression and flush to file)" should be in the help doc string for the logger. 

The `BLFWriter` `file_size` function is a bit more nuanced than is described there. Compression is not accounted for when checking the size of `_buffer_size`, and therefore if the rollover is caused by the `_buffer_size` being exceeded, then the resulting file created will have a smaller file size than the `file_size` requested. If `max_container_size << max_bytes` then the size of the file will approach the specified `max_bytes`. 

- [x] Provide an explanation for the nuanced file size rollover in the `BLFWriter` doc string

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -359,6 +359,13 @@ class BLFReader(MessageReader):
 class BLFWriter(FileIOMessageWriter):
     """
     Logs CAN data to a Binary Logging File compatible with Vector's tools.
+
+    .. note::
+        The `buffer_size` will always stay under the `max_container_size`, which
+        is enforced at the end of the `_add_object` function. If the
+        `max_container_size > :attr:io.logger.RotatingLogger.max_bytes` then
+        the resulting file size will be less than described due to an
+        unaccounted compression of the buffer when writing to the file.
     """
 
     file: BinaryIO
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -359,13 +359,6 @@ class BLFReader(MessageReader):
 class BLFWriter(FileIOMessageWriter):
     """
     Logs CAN data to a Binary Logging File compatible with Vector's tools.
-
-    .. note::
-        The `buffer_size` will always stay under the `max_container_size`, which
-        is enforced at the end of the `_add_object` function. If the
-        `max_container_size > :attr:io.logger.RotatingLogger.max_bytes` then
-        the resulting file size will be less than described due to an
-        unaccounted compression of the buffer when writing to the file.
     """
 
     file: BinaryIO
--- a/can/io/logger.py
+++ b/can/io/logger.py
@@ -327,9 +327,10 @@ class SizedRotatingLogger(BaseRotatingLogger):
 
         # BLFWriter specific
         _128_kb = 128 * 1024
-        if kwargs.get('max_container_size', 1) and max_bytes > 0:
-            kwargs['max_container_size'] = \
+        if kwargs.get("max_container_size", 1) and max_bytes > 0:
+            kwargs["max_container_size"] = (
                 _128_kb if max_bytes > _128_kb else max_bytes * 0.1
+            )
 
         self._writer = self._get_new_writer(self.base_filename)
 
--- a/can/io/logger.py
+++ b/can/io/logger.py
@@ -325,6 +325,12 @@ class SizedRotatingLogger(BaseRotatingLogger):
         self.base_filename = os.path.abspath(base_filename)
         self.max_bytes = max_bytes
 
+        # BLFWriter specific
+        _128_kb = 128 * 1024
+        if kwargs.get('max_container_size', 1) and max_bytes > 0:
+            kwargs['max_container_size'] = \
+                _128_kb if max_bytes > _128_kb else max_bytes * 0.1
+
         self._writer = self._get_new_writer(self.base_filename)
 
     def should_rollover(self, msg: Message) -> bool:
--- a/can/io/logger.py
+++ b/can/io/logger.py
@@ -325,12 +325,6 @@ class SizedRotatingLogger(BaseRotatingLogger):
         self.base_filename = os.path.abspath(base_filename)
         self.max_bytes = max_bytes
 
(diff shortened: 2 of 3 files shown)


j-c-cook:
@zariiii9003 Do I have this explanation correct of how the `file_size` function behaves for `BLFWriter`? Do you think this explanation should be moved from the `logger` help doc string and into the `BLFWriter` function? To me it seems (1) perhaps it over complicates the `logger` help doc string and (2) its not an adequate description of the behavior.

hardbyte (approved):

can/io/blf.py
```diff
@@ -359,6 +359,13 @@ def _parse_data(self, data):
 class BLFWriter(FileIOMessageWriter):
     """
     Logs CAN data to a Binary Logging File compatible with Vector's tools.
+
+    .. note::
+        The `buffer_size` will always stay under the `max_container_size`, which
+        is enforced at the end of the `_add_object` function. If the
+        `max_container_size > :attr:io.logger.RotatingLogger.max_bytes` then
+        the resulting file size will be less than described due to an
+        unaccounted compression of the buffer when writing to the file.
```
zariiii9003, line 368: This note references implentation details. I think this could be quite confusing for new users who are reading the docs. 
Instead of this, i would prefer a note in the file_size argument: The resulting file_size is dependent on the log format implementation and cannot be guaranteed. Something like that...

j-c-cook: I've updated the help statement for the file size based on your input. The one thing I don't like about it much is that it may create a sense of uncertainty. When `file_size` is the limiting factor, even with `BLFWriter` the resulting files created should be similar in size to one another. I think the statement may convey a wrong message; that the file size may vary wildly.

can/logger.py
```diff
@@ -193,9 +193,10 @@ def main() -> None:
         "--file_size",
         dest="file_size",
         type=int,
-        help="Maximum file size in bytes (or for the case of blf, maximum "
-        "buffer size before compression and flush to file). Rotate log "
-        "file when size threshold is reached.",
+        help="Maximum file size in bytes. Rotate log file when size threshold "
+        "is reached. (The resulting file sizes will be consistent, but is not "
+        "guaranteed to be exactly what is specified here due to the rollover "
+        "conditions being logger implementation specific.)",
```
j-c-cook, line 199: @zariiii9003 What do you think?

can/io/logger.py
```diff
@@ -325,6 +325,12 @@ def __init__(
         self.base_filename = os.path.abspath(base_filename)
         self.max_bytes = max_bytes
 
+        # BLFWriter specific
+        _128_kb = 128 * 1024
+        if kwargs.get('max_container_size', 1) and max_bytes > 0:
+            kwargs['max_container_size'] = \
+                _128_kb if max_bytes > _128_kb else max_bytes * 0.1
+
```
j-c-cook, line 333: @zariiii9003 This would get rid of the nuance. I'm anticipating that you are not going to like the `BLFWriter` specific code inside of the `SizedRotatingLogger` constructor, but let me know.

can/logger.py
```diff
@@ -193,9 +193,10 @@ def main() -> None:
         "--file_size",
         dest="file_size",
         type=int,
-        help="Maximum file size in bytes (or for the case of blf, maximum "
-        "buffer size before compression and flush to file). Rotate log "
-        "file when size threshold is reached.",
+        help="Maximum file size in bytes. Rotate log file when size threshold "
+        "is reached. (The resulting file sizes will be consistent, but is not "
```
zariiii9003, line 197:
```suggestion
        "is reached. (The resulting file sizes will be consistent, but are not "
```

can/logger.py
zariiii9003: Looks good, thanks.

can/io/logger.py
zariiii9003: Yes, i don't like this here 😄 
Also i do not understand what this does tbh, what's the purpose of this? And why is the default value of `kwargs.get('max_container_size', 1)`  `=1`?

j-c-cook: If `max_container_size` is not defined then set the `max_container_size`. If `max_container_size` is not defined then set the value to be `128 kb` unless the `file_size` is less than `128 kb`. In that case set the `max_container_size` to 10% of the `file_size`. 

I'm reverting the commits because I agree that its not a good location.

zariiii9003:
@j-c-cook Can i merge this?

j-c-cook:
> @j-c-cook Can i merge this?

Yes. As long as you think it’s appropriate to have the details contained in the parenthesis. That it does not crowd up the help description printout. I’ve wondered if it’s more than necessary, and if we should just keep it at https://github.com/hardbyte/python-can/pull/1401#discussion_r984000232.

zariiii9003:
There are parameters with longer descriptions so i think it's fine. 
Alternatively we could disable  support for problematic suffixes like gzip and soonish mf4. The [original implementation ](https://github.com/hardbyte/python-can/blob/5441c5a1809145396e5946769da8d6888455d2b7/can/io/logger.py#L117)restricted the supported file formats, we could go back to that.

j-c-cook:
> There are parameters with longer descriptions so i think it's fine. 

In that case, merge at will.

> Alternatively we could disable support for problematic suffixes like gzip and soonish mf4. The [original implementation ](https://github.com/hardbyte/python-can/blob/5441c5a1809145396e5946769da8d6888455d2b7/can/io/logger.py#L117)restricted the supported file formats, we could go back to that.

I do agree that this may be useful, but think it should possibly be apart of a separate PR since we may have some details to resolve. For example, I think it would be useful if the dictionary was in the following format, given that some implementations do work with `gzip`:
```
supported_writers = {
    BLFWriter: [".blf"],
    CanutilsLogWriter: [".log", ".log.gz"]
}
```

Note: #1378 is currently a known open issue relating to the topic.

PR #1408, opened 2022-10-10
setup.cfg: Use license_files instead of license_file

nertpinx:
Without this (and recent enough setuptools) I get this warning:

/home/nert/dev/python-can/.tox/.package/lib/python3.10/site-packages/setuptools/config/setupcfg.py:508: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.


zariiii9003 (approved):
Thanks

PR #1410, opened 2022-10-11
Add usage of select  instead of pulling for Linux based platform (PCAN Interface)

BKaDamien:
Avoid pulling of "time.sleep(0.001)" in pcan interface method _recv_internal, and use select python function instead.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -493,9 +493,13 @@ class PcanBus(BusABC):
                         return None, False
                 elif HAS_EVENTS and IS_LINUX:
                     recv, _, _ = select.select([self._recv_event], [], [], timeout_s)
+                    log.error(recv)
+                    log.error(IS_LINUX)
                     if self._recv_event not in recv:
+                        log.error("Nothing received")
                         return None, False
                     result = self._read_message()
+                    log.error(f"oh recived something {result}")
                 elif timeout is not None and time.perf_counter() >= end_time:
                     return None, False
                 else:
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -474,7 +474,6 @@ class PcanBus(BusABC):
             # We will utilize events for the timeout handling
             timeout_ms = int(timeout * 1000) if timeout is not None else INFINITE
         elif HAS_EVENTS and IS_LINUX:
-            # We will utilize events for the timeout handling
             timeout_s = timeout if timeout != None else None
         elif timeout is not None:
             # Calculate max time
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -1,7 +1,6 @@
 """
 Enable basic CAN over a PCAN USB device.
 """
-
 import logging
 import time
 from datetime import datetime
@@ -18,6 +17,7 @@ from ...exceptions import CanError, CanOperationError, CanInitializationError
 
 
 from .basic import (
+    PLATFORM,
     PCAN_BITRATES,
     PCAN_FD_PARAMETER_LIST,
     PCAN_CHANNEL_NAMES,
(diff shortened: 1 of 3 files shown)


BKaDamien:
Hi python-can maintainers/developers,

I need help regarding my changes related to the introduction of select for linux platform:

The issue come from test_recv_no_message from test module test_pcan, when I run the sae code everything is fine the test_recv_no_message is passed. 

below my setup:
test run on : Ubuntu 20.04.3 LTS
python :  3.8.10

But since the beginning of this PR, the Ubuntu tests  are constantly failing at test_recv_no_message. Based on the trace from pytest it seems that despite of the select mock the return is not executed  and the code try to access an attribute from a None Value(see link below).
https://github.com/hardbyte/python-can/actions/runs/3273689147/jobs/5402946921

Thanks for your support

felixdivo:
Looks good! I haven't tested it with hardware though. We should probably do that.

felixdivo (approved):

BKaDamien:
> Looks good! I haven't tested it with hardware though. We should probably do that.

Tested on my side, this update will be a little less consumer than the polling.

felixdivo:
Sorry for being this pedantic, but on which platforms did you try this? There appear to be 4 variants here: Windows and Linux and each with and without polling. For each of the 4, sending and receiving has to work. :slightly_smiling_face:

BKaDamien:
> Sorry for being this pedantic, but on which platforms did you try this? There appear to be 4 variants here: Windows and Linux and each with and without polling. For each of the 4, sending and receiving has to work. 🙂

Based on my changes:
Tested on Windows -> OK (use WaitForSingleObject windows base event)
Test on Ubuntu -> OK (use python select base event)
For the other MAC, .... -> Just use the polling

felixdivo:
Okay, thank you for documenting that here. 🙂 I think this is sufficient, since the polling implementation id not really change and there are also unit tests that make sure nothing gets too broken.

I'd propose to merge it as it.

BKaDamien:
> Okay, thank you for documenting that here. 🙂 I think this is sufficient, since the polling implementation id not really change and there are also unit tests that make sure nothing gets too broken.
> 
> I'd propose to merge it as it.

You are the boss :)

BKaDamien:
Arf I wrongly click and close I am sorry

felixdivo:
@hardbyte You can merge it before or after the 4.1.0 release. If you add it before, we'd have to amend the changelog.

can/interfaces/pcan/pcan.py
```diff
@@ -437,11 +456,26 @@ def set_device_number(self, device_number):
...
+
     def _recv_internal(self, timeout):
 
-        if HAS_EVENTS:
+        if HAS_EVENTS and IS_WINDOWS:
             # We will utilize events for the timeout handling
             timeout_ms = int(timeout * 1000) if timeout is not None else INFINITE
+        elif HAS_EVENTS and IS_LINUX:
+            # We will utilize events for the timeout handling
+            timeout_s = timeout if timeout != None else None
```
zariiii9003, line 478: This seems redundant

can/interfaces/pcan/pcan.py
```diff
@@ -64,7 +64,8 @@
 log = logging.getLogger("can.pcan")
 
 MIN_PCAN_API_VERSION = version.parse("4.2.0")
-
+IS_WINDOWS = PLATFORM == "Windows"
+IS_LINUX = PLATFORM == "Linux"
```
felixdivo, line 68: We could also move these to `can/interfaces/pcan/basic.py` and just import `IS_WINDOWS` & `IS_LINUX`, and also change the if around `import winreg` in there to then use `if IS_WINDOWS` instead. Small thing tough.

can/interfaces/pcan/pcan.py
BKaDamien: I don't think so select accept timeout in seconds or None.
On windows WaitForSingleObject expects a timeout in ms

zariiii9003:
@BKaDamien I cleaned up a bit, please take a look.

BKaDamien:
> @BKaDamien I cleaned up a bit, please take a look.

It's fine. Thanks for the chnages.

zariiii9003 (approved):

PR #1415, opened 2022-10-21
Add CANine interface

yconst:
This is a USB CDC interface that is derived from slcan. We developed it because it offers much better performance in several Windows machines that have issues with serial port responsiveness, and as such VCP devices lag significantly.

It uses the same commands as slcan but communicates in binary packets of predefined size.

It is compatible with our own CANine adapter, as well as Canable (v0.4) devices. 

[Firmware repo](https://github.com/tinymovr/CANine)
[Webdfu-based updater](https://tinymovr.github.io/CANine/)


yconst:
Is this considered for merging? Is there something further that we need to do?

zariiii9003:
> Is this considered for merging? Is there something further that we need to do?

Following the discussion in #1437, did you consider using the [plugin interface](https://python-can.readthedocs.io/en/develop/plugin-interface.html#plugin-interface)?

yconst:
That's what we ended up doing.

zariiii9003:
@yconst You could create a PR for a doc entry [here](https://python-can.readthedocs.io/en/develop/plugin-interface.html#example-interface-plugins) and an optional dependency in setup.py

We'll release 4.2 soon

yconst:
@zariiii9003 We did it here: https://github.com/hardbyte/python-can/pull/1556

PR #1385, opened 2022-08-27
Address blf GzipFile negative seek error

j-c-cook:
The current implementation of the `BLFWriter` requires that the file be negatively sought (https://github.com/hardbyte/python-can/pull/1379#discussion_r952598015). The `cpython` implementation does not allow for `GzipFiles` to be negatively sought while in write mode. Modifying `gzip` is outside of the scope of this project. I see two options here:
1. Modify the `BLFWriter` to close, seek then re-open (@christiansandberg says this was previously being done)
2. Inform the user that the `GzipFile` format is not compatible with the `BLFWriter` (later enhancing `python-can` so that zip formats that can be negatively sought are available)

I'm proposing option 2. 

closes #1378

PR #1363, opened 2022-08-07
Update CHANGELOG for v4.1.0 release

zariiii9003:

PR #1427, opened 2022-11-07
Fixing memory leak in neoVI bus where message_receipts grows with no limit

pierreluctg:
Fixing memory leak in neoVI bus where `message_receipt`s grows with no limit. `message_receipts` was growing without bound on msg Rx side

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -318,8 +318,9 @@ class NeoViBus(BusABC):
             if is_tx:
                 if bool(ics_msg.StatusBitField & ics.SPY_STATUS_GLOBAL_ERR):
                     continue
-                if ics_msg.DescriptionID:
-                    receipt_key = (ics_msg.ArbIDOrHeader, ics_msg.DescriptionID)
+
+                receipt_key = (ics_msg.ArbIDOrHeader, ics_msg.DescriptionID)
+                if ics_msg.DescriptionID and receipt_key in self.message_receipts:
                     self.message_receipts[receipt_key].set()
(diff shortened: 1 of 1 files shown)


felixdivo:
We should add some typing information for constructs like `self.message_receipts` to make it easier to read what's in these generic containers. 😄

felixdivo (approved):
Looks great!

felixdivo:
At least for 4.0.0 we did not have a real time estimate. I just sat down at some point and proposed a date (and missed it ^^).

I'm a fan of releasing more often than previously, since I personally find it very annoying if you contribute a bugfix to a repo and they do not release it for ages. So you are forced to install from source or stuff like that.

So now would be a good time for the next release, but since I will not prepare it I don't want to push hard. Maybe include #1385 since it seems to be already finished.

BTW: There is #1363 by @zariiii9003 for 4.1.0, which might be a better place to discuss it than here.

pierreluctg:
@felixdivo there will be no 4.0.1 release? If this is the case can we close the milestone to avoid confusion?

felixdivo:
It certainly did confuse me. If there will be no such release, we should reassign all existing (also closed & merged!) issues/PRs to 4.1.0. @zariiii9003 should I do that? I think you have a good overview over what makes sense now.

zariiii9003:
We agreed on changing it to 4.1.0 because of the `can.logger` API changes. If you don't mind updating those PRs, that would be cool :)

Once #1385 and #1363 are merged, we should create a new release.

felixdivo:
Sounds good :+1: will do, latest Saturday

j-c-cook:
#1385 is migrated to #1429.

PR #1435, opened 2022-11-15
Add with statement to example in README.md

piotrszleg:
Not using with expression in code might cause lack of cleanup and hard to trace errors, also because Bus doesn't override `__del__`. 

So it's a good idea to provide a 100% percent correct example in README for people who don't go into examples folder and read into the code.

Not sure if a sleep call is not needed after the notifier though.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,7 +89,8 @@ Example usage
     # import the library
     import can
 
-    # create a bus instance
+    # create a bus instance using 'with' statement,
+    # this will cause bus.shutdown() to be called on the block exit;
     # many other interfaces are supported as well (see documentation)
     with can.Bus(interface='socketcan',
                   channel='vcan0',
--- a/README.rst
+++ b/README.rst
@@ -91,21 +91,21 @@ Example usage
 
     # create a bus instance
     # many other interfaces are supported as well (see documentation)
-    bus = can.Bus(interface='socketcan',
+    with can.Bus(interface='socketcan',
                   channel='vcan0',
-                  receive_own_messages=True)
+                  receive_own_messages=True) as bus:
 
-    # send a message
-    message = can.Message(arbitration_id=123, is_extended_id=True,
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -91,21 +91,21 @@ Example usage
 
     # create a bus instance
```
hardbyte, line 92:
```suggestion
    # create a bus instance (using a context manager to call `Bus.shutdown()` after use)
```

felixdivo:
This is really funny since [the docs](https://python-can.readthedocs.io/en/master/) already give a good example. ^^ Thanks for the contribution!

Do you think we should add a [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) which calls `shutdown()`?

README.rst
piotrszleg: I added a similar comment but with "with" statement instead of context manager, because people might know about it from using files but not know the name "context manager".

piotrszleg:
> Do you think we should add a [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) which calls `shutdown()`?

In this case I think there should also be a `_is_shutdown` property to make sure shutdown is not called more than once.

But I think most of users, including me, prefer to shutdown it manually and not depend on garbage collection.

felixdivo:
> > Do you think we should add a [`__del__`](https://docs.python.org/3/reference/datamodel.html#object.__del__) which calls `shutdown()`?
> 
> In this case there should also be a `_is_shutdown` property to make sure shutdown is not called more than once.
> 
> But I think most of users, including me, prefer to shutdown it manually and not depend on garbage collection.

Yeah, or the "double calling" should be taken care of by the subclass. You approach is probably simplest.

It was just meant like a "last resort". I think you should never really rely on `__del__` being called. But if it exists, it might ensure cleanup at least in a few cases.

felixdivo:
See  #1448 for a discussion on `__del__`.

PR #1440, opened 2022-11-17
Allow restarting of transmission tasks for socketcan

hardbyte:
I was going through the examples and found that restarting BCM tasks was broken.

The error is due to a requirement that the transmission task NOT exist in kernel space when sending the restart message. I'm not sure how long it has been broken for.

@hartkopp would you mind checking my logic?

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -388,7 +388,9 @@ class CyclicSendTask(
             can_id=self.task_id,
             nframes=0,
         )
-        log.debug(f"Reading properties of (cyclic) transmission task id={self.task_id}", )
+        log.debug(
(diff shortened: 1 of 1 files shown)


can/interfaces/socketcan/socketcan.py
```diff
@@ -349,7 +349,7 @@ def __init__(
         self.task_id = task_id
         self._tx_setup(self.messages)
 
-    def _tx_setup(self, messages: Sequence[Message]) -> None:
+    def _tx_setup(self, messages: Sequence[Message], raise_if_task_exists=True) -> None:
```
zariiii9003, line 352:
```suggestion
    def _tx_setup(self, messages: Sequence[Message], raise_if_task_exists: bool =True) -> None:
```

PR #1443, opened 2022-11-21
Use furo html theme for documentation

zariiii9003:
What do you think?

The output can be seen [here](https://python-can--1443.org.readthedocs.build/en/1443/).


felixdivo (approved):
Why not 👍 😃

pierreluctg (approved):

hardbyte (approved):

Issue #1444, opened 2022-11-23
OverflowError: mktime argument out of range


jung-hunsoo:
### Describe the bug

Exception `OverflowError: mktime argument out of range` occurred while reading BLF file on Windows OS.

### To Reproduce

```python
import os
import can

blf_path = os.path.join("my_can_data.blf")
loaded_blf = can.BLFReader(blf_path)
```

### Expected behavior

A given BLF file is read without exception.

### Additional context

OS and version: Windows 11 Pro
Python version: 3.10.6
python-can version: 4.0.0
python-can interface/s (if applicable): n/a

jung-hunsoo:
No exception occurred when running the same code on Linux environment.

PR #1445, opened 2022-11-23
Adjust minimum year to 1971

jung-hunsoo:
- If there's a case incoming date data is incorrect, adjust year to start from 1971.
- This resolves issue of #1444

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -119,8 +119,13 @@ def timestamp_to_systemtime(timestamp: float) -> TSystemTime:
 
 def systemtime_to_timestamp(systemtime: TSystemTime) -> float:
     try:
+        year_adjusted = (
+            systemtime[0]
+            if systemtime[0] >= 1971
+            else systemtime[0] + (1971 - systemtime[0])
+        )
         t = datetime.datetime(
-            systemtime[0],
+            year_adjusted,
             systemtime[1],
             systemtime[3],
             systemtime[4],


jung-hunsoo:
Thanks @felixdivo.
With the fixed code, the file is loaded properly. Please be understood that it's hard to share the captured image. Instead anyone can pull and test the code from https://github.com/jung-hunsoo/python-can.git@patch-1.

felixdivo:
I don't understand why it is hard to share the captured image. Is it very large?

Further, I don't quite get why the problem occurs. `datetime.datetime(1960, 1, 1)`  works fine for me, and the change expression `systemtime[0] + (1971 - systemtime[0])` is just `1971` right? And why that boundary, it looks oddly close to [Unix time](https://en.wikipedia.org/wiki/Unix_time)?

zariiii9003:
> I don't understand why it is hard to share the captured image. Is it very large?
> 
> Further, I don't quite get why the problem occurs. `datetime.datetime(1960, 1, 1)` works fine for me, and the change expression `systemtime[0] + (1971 - systemtime[0])` is just `1971` right? And why that boundary, it looks oddly close to [Unix time](https://en.wikipedia.org/wiki/Unix_time)?

I think this only occurs on Windows

can/io/blf.py
```diff
@@ -119,8 +119,13 @@ def timestamp_to_systemtime(timestamp: float) -> TSystemTime:
...
 def systemtime_to_timestamp(systemtime: TSystemTime) -> float:
     try:
+        year_adjusted = (
+            systemtime[0]
+            if systemtime[0] >= 1971
+            else systemtime[0] + (1971 - systemtime[0])
+        )
         t = datetime.datetime(
-            systemtime[0],
+            year_adjusted,
```
zariiii9003, line 128:
```suggestion
            max(systemtime[0], 1971),
```

This should be sufficient?

zariiii9003: Also, shouldn't it be 1970? 🤔

felixdivo: Yes, that were my thoughts too. According to [the docs of `datetime.datetime`](https://docs.python.org/3/library/datetime.html#datetime.datetime), small numbers (e.g., year 1960) should be possible though.

felixdivo:
I think we should close this as it is currently stuck.

Issue #1262, opened 2022-02-22
Cycle time fluctuating on Windows 10 (using Vector inferface)


ALPfulben:
### Describe the bug
The cycle times of CAN messages are fluctuating under Windows when using the send_periodic() function.
Expected cycle time 0.02s; Actual cycle time: 0.015s, 0.035s, 0.038s, 0.015s,...

When using Linux (Raspberry Pi) with the same Python code the cycles times are very accurate.

### To Reproduce
```python
bus = can.ThreadSafeBus(interface="vector", channel=0, app_name="CANoe", bitrate=500000)
msg_relay_ctrl_00 = can.Message(arbitration_id=0xABC, is_extended_id=False, data=[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
bus.send_periodic(msg_relay_ctrl_00, 0.02)  # 20ms 
```

### Expected behavior
Cycle time of the CAN message should be 20ms +- 10%.

### Observed behavior 
Measured cycle times are fluctuating (e.g. 0.014940s, 0.036936s, etc.), see log below.

### Additional context
When executing a CPU intensive task (for example watching a video) the cycle time is more accurate.

OS and version: Windows 10 Enterprise 2016
Python version: 3.9 (64-bit)
python-can version: 4.0.0
python-can interface: vector

<details><summary>Log & Code snippet</summary>

Some measured cycle times (average of the value is approx. 0.02s). Measured with CANoe.

0.014940
0.016039
0.037955
0.015012
0.016029
0.014916
0.016187
0.037411
0.015168
0.016315
0.015999
0.015839
0.036936
0.015417
0.015179
0.016255
0.015503
0.037425
0.014502
0.016747
0.014950
0.015247
0.038727

```python
bus = can.ThreadSafeBus(interface="vector", channel=0, app_name="CANoe", bitrate=500000)
msg_relay_ctrl_00 = can.Message(arbitration_id=0xABC, is_extended_id=False, data=[0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00])
bus.send_periodic(msg_relay_ctrl_00, 0.02)  # 20ms 
```
</details>

zariiii9003:
Hello, thank you for reporting.
Can you please compare the timing between python-can 3.3.4 and 4.0.0?

[...]

PR #1449, opened 2022-11-24
Use high resolution timer on Windows

zariiii9003:
Improve timing precision for BroadcastManager without calling timeBeginPeriod.

If merged, this should stay in develop for some time. I'm not sure if catching `OSError` is sufficient to make this compatible to old Windows versions.

Related: #1262

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     types: [ published ]
   pull_request:
   push:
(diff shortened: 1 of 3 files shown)


hardbyte:
> If merged, this should stay in develop for some time. I'm not sure if catching OSError is sufficient to make this compatible to old Windows versions.

We've never had a problem with that!!

hardbyte (approved):

Issue #1458, opened 2022-12-12
Opening a PCanBus in FD mode with Bus.__new__ breaks with python-can 4.0.0 / 4.1.0


lumagi:
### Describe the bug
I am using the PCAN interface in an FD configuration with custom bit timings. I provide the parameters, i.e. `f_clock`, `nom_brp`, `data_brp`, ..., directly as kwargs to `Bus.__new__` (see example below). With the transition from version `v3.3.4` to `v4.1.0`, I can no longer instantiate the bus and always receive the error message below.

I took a look at the code in question, and it appears like this might be a corner case that was forgotten about when adding the `BitTiming`class. When instantiating the `PcanBus` for a CAN-FD bus, it is mandatory to supply the `f_clock` parameter as well as the bit timings directly to the constructor. The `bitrate` parameter is ignored because the PCAN driver configures the interface directly based on the timing parameters (see [pcan.py](https://github.com/hardbyte/python-can/blob/25d30b529dcc28cbe75320d8315b57233fd10cc1/can/interfaces/pcan/pcan.py#L136)).
The exception is thrown in `_create_bus_config` in [util.py](https://github.com/hardbyte/python-can/blob/25d30b529dcc28cbe75320d8315b57233fd10cc1/can/util.py#L252). The function in question assumes that whenever certain timing-related arguments (such as `f_clock`, `tseg1`, `tseg2` or `brp`) are present in kwargs, the `bitrate` argument must be present as well. For the case of the `PcanBus`, this is not the case since the interface implementation does not use the `BitTiming` class nor requires `bitrate` to be set when running in CAN-FD mode.

I did some searching in the repository and found issue #614 by @christiansandberg . I also added a comment there, but didn't receive a response.

### To Reproduce
See code example below

### Expected behavior
The constructor should not require the `bitrate` parameter to be set and instantiate the bus correctly.

### Additional context

OS and version: Windows
Python version: 3.9
python-can version: 4.1.0
python-can interface/s (if applicable): PEAK PCAN-USB FD

<details><summary>Traceback and logs</summary>

```
Traceback (most recent call last):
  File "<...>", line 23, in <module>
    main()
  File "<...>", line 12, in main
    bus = Bus(**params)
  File "<...>", line 108, in __new__
    kwargs = load_config(config=kwargs, context=context)
  File "<...>", line 192, in load_config
    bus_config = _create_bus_config(config)
  File "<...>", line 252, in _create_bus_config
[...]

PR #1460, opened 2022-12-18
Fix Bus.__new__ for PEAK CAN-FD interfaces

lumagi:
This pull request removes unused code in Bus.__new__ that is responsible for  creating BitTiming instances, but is incompatible with PEAK devices. See issue #1458 .
    
The BitTiming class is an attempt at unifying the various timing parameters of the individual interfaces. The idea is that instead of manually supplying multiple parameters that make up the timing definition of the interface, one can instead supply a single instance of the BitTiming class, which will also automatically calculate derivative values from its input.
    
At the moment, this class is only used by two interfaces: CANtact and CANanalystii. Both either accept a single bitrate or a BitTiming instance. The latter will overrule the former. For reference, a pull request that would implement BitTiming for PEAK devices has not been merged, see #625 .
    
The code that is removed with this commit is part of the generic Bus.__new__ constructor. The removed code searches the set of kwargs parameters for timing-related values. If it finds at least one such value, it creates a BitTiming class instance and adds it to the list of parameters. However, it breaks compatibility with the PEAK interface, see issue #1458. Additionally, the code in question is generic and applies to all interfaces. Instantiating a class here is prone to issues since it must be generic enough to fit all use cases. A better approach would be to simply forward the parameters as was done previously and leave it up to the individual interfaces to handle things properly.

### DIFF
--- a/can/util.py
+++ b/can/util.py
@@ -233,25 +233,6 @@ def _create_bus_config(config: Dict[str, Any]) -> typechecking.BusConfig:
     if "data_bitrate" in config:
         config["data_bitrate"] = int(config["data_bitrate"])
 
-    # Create bit timing configuration if given
-    timing_conf = {}
-    for key in (
-        "f_clock",
-        "brp",
-        "tseg1",
-        "tseg2",
-        "sjw",
-        "nof_samples",
-        "btr0",
-        "btr1",
-    ):
-        if key in config:
-            timing_conf[key] = int(str(config[key]), base=0)
-            del config[key]
-    if timing_conf:
-        timing_conf["bitrate"] = config["bitrate"]
-        config["timing"] = can.BitTiming(**timing_conf)
-
     return cast(typechecking.BusConfig, config)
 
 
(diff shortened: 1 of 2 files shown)


zariiii9003 (approved):

can/util.py
```diff
@@ -233,25 +233,6 @@ def _create_bus_config(config: Dict[str, Any]) -> typechecking.BusConfig:
...
-        "tseg1",
-        "tseg2",
-        "sjw",
-        "nof_samples",
-        "btr0",
-        "btr1",
-    ):
-        if key in config:
-            timing_conf[key] = int(str(config[key]), base=0)
-            del config[key]
```
zariiii9003, line 250: Have you tried just removing this line with `del` instead of everything?

lumagi: For my specific issue with the PCAN interface, it's actually line 252 that's causing the immediate problem (although the `del` would also be an issue further down the road). Since it detects that `f_clock` is present in `kwargs`, it automatically tries to create the `timing_conf` and simply assumes that the config must also contain a `bitrate` key in line 252. This is not always the case for PCAN.

There are multiple reasons for me suggesting the removal of this piece of code:
* It causes issues with at least two interfaces
* It doesn't look like the code is actively used by any interface or anyone has picked up on it since it was merged
* The contract between `Bus.__new__` and the interface implementations is very weak. It basically says: If you name your parameters in the correct way, we'll automagically create the `BitTiming` instance for you. But only for Classic CAN, not CAN-FD.
* The author itself stated in his commit that it (obviously) doesn't support CAN-FD.

PR #1462, opened 2022-12-21
Add deprecation warning for 'bustype' parameter

zariiii9003:


can/interface.py
```diff
@@ -88,6 +88,7 @@ class Bus(BusABC):  # pylint: disable=abstract-method
     """
 
     @staticmethod
+    @deprecated_args_alias(bustype="interface")
```
felixdivo, line 91:
```suggestion
    @deprecated_args_alias(bustype="interface")  # Deprecated since python-can 4.2
```

felixdivo (approved):
Great idea to finally clean this up!

Issue #1461, opened 2022-12-19
Unable to Load PCANBasic Module on Windows on ARM


sanssecours:
### Describe the bug

Executing python-can code based on the PCANBasic API on Windows on ARM does not seem to work. The code line:

https://github.com/hardbyte/python-can/blob/f3136fb5e85f11ecb9ba83744a955a706f9dbde2/can/interfaces/pcan/basic.py#L661

is unable to load the DLL:

```
FileNotFoundError: Could not find module 'PCANBasic' (or one of its dependencies). Try using the full path with constructor syntax.
```

This error happens both on the x64 version of Python `3.10` and the (experimental) ARM version of Python `3.11`.

### To Reproduce

1. Install Windows on ARM (e.g. using [Parallels Desktop](https://www.parallels.com) on an Apple Silicon based Mac)
2. Install [PCAN driver](https://www.peak-system.com/quick/DrvSetup)
3. Install python-can:

    ```sh
    pip install python-can[pcan]
    ```

4. Run the following Python code:
   
   ```py
   from can.interfaces.pcan.basic import PCANBasic
   PCANBasic()
   ```

### Expected behavior

The code works without problems.

### Actual behavior

Execution fails with the following stack trace:

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\rene\AppData\Local\Programs\Python\Python310\lib\site-packages\can\interfaces\pcan\basic.py", line 661, in __init__
    self.__m_dllBasic = windll.LoadLibrary("PCANBasic")
  File "C:\Users\rene\AppData\Local\Programs\Python\Python310\lib\ctypes\__init__.py", line 452, in LoadLibrary
    return self._dlltype(name)
  File "C:\Users\rene\AppData\Local\Programs\Python\Python310\lib\ctypes\__init__.py", line 374, in __init__
    self._handle = _dlopen(self._name, mode)
FileNotFoundError: Could not find module 'PCANBasic' (or one of its dependencies). Try using the full path with constructor syntax.

Could not find module 'PCANBasic' (or one of its dependencies). Try using the full path with constructor syntax.
```

### Additional context

OS and version: Windows 11 on ARM 
Python version: Python 3.10.8
python-can version: python-can-4.0.0
python-can interface/s (if applicable): [PCAN-USB FD](https://www.peak-system.com/PCAN-USB-FD.365.0.html)

For now I use the following **(very ugly) workaround** to fix the problem:

```py
from re import match
[...]

PR #1463, opened 2022-12-21
PCAN: Fix Detection of Library on Windows on ARM

sanssecours:
Before this commit detection of the PCAN DLL would fail on Windows on ARM, regardless if you used the native Python version, or the x64 version of Python. After this fix detection should work properly as long as the PCAN library for your version of Python is listed first in the `PATH` variable. The default:

1. `C:\Program Files\PEAK-System\PEAK-Drivers 4\APIs\ARM64\` before
2. `C:\Program Files\PEAK-System\PEAK-Drivers 4\APIs\x64\`

should work if you use the (native) ARM version of Python. If you reorder these paths, then loading the library works in the x64 version of Python.

I tested the fix on the native version of Python `3.11` and the x64 version of Python `3.10` in Window on ARM, where everything seemed to work as expected. For more information, please take a look at issue #1461.

### DIFF
--- a/can/interfaces/pcan/basic.py
+++ b/can/interfaces/pcan/basic.py
@@ -658,7 +658,7 @@ class PCANBasic:
         #
         if platform.system() == "Windows":
             # Loads the API on Windows
-            self.__m_dllBasic = windll.LoadLibrary("PCANBasic")
+            self.__m_dllBasic = windll.LoadLibrary(find_library("PCANBasic"))
             aReg = winreg.ConnectRegistry(None, winreg.HKEY_LOCAL_MACHINE)
             try:
                 aKey = winreg.OpenKey(aReg, r"SOFTWARE\PEAK-System\PEAK-Drivers")


felixdivo (approved):

sanssecours:
Looks like the test suite [does not like  the Window emoji (`🪟`) on Windows](https://github.com/hardbyte/python-can/actions/runs/3748530924/jobs/6369914247):

```
...
UnicodeEncodeError: 'charmap' codec can't encode character '\U0001fa9f' in position 21: character maps to <undefined>
[44](https://github.com/hardbyte/python-can/actions/runs/3748530924/jobs/6369914247#step:5:45)
  gh: FAIL code 2 (13.53 seconds)
[45](https://github.com/hardbyte/python-can/actions/runs/3748530924/jobs/6369914247#step:5:46)
  evaluation failed :( (13.75 seconds)
```

At least that is what I would assume from the output  of the Python code:

```py
print("\U0001fa9f")
```

Should I reopen the pull request from a different different branch (that only contains ASCII characters)?

felixdivo:
I'd suggest just merging this. @hardbyte?

zariiii9003:
Where is the emoji coming from? pip?

zariiii9003:
I merged it and now the CI fails, oops.

PR #1464, opened 2022-12-21
Tiny type narrowing

felixdivo:
- Slightly narrows the type of `can.interfaces.BACKENDS` from `Dict[str, Tuple[str, ...]]` to `Dict[str, Tuple[str, str]]`
- Will behave more robust if too many `:` are given in entry-point declaration


felixdivo:
Wow, the official documentation on that module is so bad ... I'll merge this with  #1465 and do it as one. It's more changes than expected.

PR #1465, opened 2022-12-21
Improve interface registry: Fix DeprecationWarning + better typing + more robust

felixdivo:
- Previously, the `entry_points().get("can.interface", [])}` expression issued the following: `DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.`. This patch fixes it. I also tested it locally. To reproduce the error before the change, simply run `python -W error -c 'import can.interfaces'`.
- Slightly narrows the type of `can.interfaces.BACKENDS` from `Dict[str, Tuple[str, ...]]` to `Dict[str, Tuple[str, str]]`
- Will behave more robust if too many `:` are given in entry-point declaration

### DIFF
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -36,7 +36,11 @@ if sys.version_info >= (3, 8):
     from importlib.metadata import entry_points
 
     # See https://docs.python.org/3/library/importlib.metadata.html#entry-points, "Compatibility Note".
-    entries = entry_points(group="can.interface") if sys.version_info >= (3, 10) else entry_points().get("can.interface", ())
+    entries = (
+        entry_points(group="can.interface")
+        if sys.version_info >= (3, 10)
+        else entry_points().get("can.interface", ())
+    )
     BACKENDS.update(
         {interface.name: tuple(interface.value.split(":")) for interface in entries}
     )
--- a/can/interfaces/__init__.py
+++ b/can/interfaces/__init__.py
@@ -3,7 +3,7 @@ Interfaces contain low level implementations that interact with CAN hardware.
 """
 
 import sys
-from typing import Dict, Tuple
+from typing import Dict, Iterable, Tuple
 
 # interface_name => (module, classname)
 BACKENDS: Dict[str, Tuple[str, ...]] = {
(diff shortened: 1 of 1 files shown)


zariiii9003:
Github actions is slow today 🤔

felixdivo:
> Github actions is slow today 🤔

Jup ... But I am too lazy to develop this locally ^^

can/interfaces/__init__.py
```diff
@@ -3,42 +3,57 @@
...
-    "systec": ("can.interfaces.systec", "UcanBus"),
-    "seeedstudio": ("can.interfaces.seeedstudio", "SeeedBus"),
-    "cantact": ("can.interfaces.cantact", "CantactBus"),
-    "gs_usb": ("can.interfaces.gs_usb", "GsUsbBus"),
-    "nixnet": ("can.interfaces.nixnet", "NiXNETcanBus"),
-    "neousys": ("can.interfaces.neousys", "NeousysBus"),
-    "etas": ("can.interfaces.etas", "EtasBus"),
-    "socketcand": ("can.interfaces.socketcand", "SocketCanDaemonBus"),
-}
+BACKENDS: Dict[str, Tuple[str, str]] = cast(
```
zariiii9003, line 9: Locally, it works without the cast. Are you sure it's necessary?

felixdivo: I'm setting it up locally and push something proper ^^ Maybe I messed something up between the changes.

can/interfaces/__init__.py
```diff
@@ -50,7 +47,8 @@
     else:
         BACKENDS.update(
             {
-                interface.name: tuple(interface.value.split(":"))
+                # This cast in wrong if interface.value is formatted badly, but we just fail later
+                interface.name: cast(Tuple[str, str], tuple(interface.value.split(":")))
```
zariiii9003, line 51:
```suggestion
                interface.name: cast(Tuple[str, str], tuple(interface.value.split(":", maxsplit=1)))
```

felixdivo: Yup, thanks!

can/interfaces/__init__.py
felixdivo: Yeah, works locally without 👍

can/interfaces/__init__.py
zariiii9003: Alternatively we could use some helper function so it actually fails when there are more or less than two values:

```python3
        def _read_entry_point(value: str) -> Tuple[str, str]:
            """Helper function to satisfy mypy."""
            _module, _class = value.split(":")
            return _module, _class
```
But it's very verbose

zariiii9003 (approved):
Looks good 👍

felixdivo:
Thanks for having a look. 😃

PR #1471, opened 2022-12-27
Update pylint

zariiii9003:
- Update pylint to version 2.15.9
- check can.io.* files with pylint
- check can.interfaces.socketcan.* with pylint


can/io/trc.py
```diff
@@ -211,12 +214,12 @@ def __init__(
...
+        if isinstance(file, str):
             self.filepath = os.path.abspath(file)
-        elif type(file) is TextIOWrapper:
+        elif isinstance(file, TextIOWrapper):
             self.filepath = "Unknown"
             logger.warning("TRCWriter: Text mode io can result in wrong line endings")
-            logger.debug(
+            logger.debug(  # pylint: disable=logging-fstring-interpolation
                 f"TRCWriter: Text mode io line ending setting: {file.newlines}"
             )
```
felixdivo, line 224: Is there a reason for not doing:

```suggestion
            logger.debug(
                "TRCWriter: Text mode io line ending setting: %s",
                file.newlines,
            )
```

.pylintrc
```diff
@@ -76,6 +76,7 @@ disable=invalid-name,
         fixme,  # We deliberately use TODO/FIXME inline
         abstract-class-instantiated,  # Needed for can.Bus
         duplicate-code,  # Needed due to https://github.com/PyCQA/pylint/issues/214
+        keyword-arg-before-vararg,
```
felixdivo, line 79: I am a bit hesitant to adding new excepts, especially since this specific earning has caused problems in the past. I get these warnings locally:

```
************* Module can.io.generic
can/io/generic.py:31:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
can/io/generic.py:92:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.canutils
can/io/canutils.py:135:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.csv
can/io/csv.py:94:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.printer
can/io/printer.py:28:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.sqlite
can/io/sqlite.py:35:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
can/io/sqlite.py:137:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.logger
can/io/logger.py:320:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.asc
can/io/asc.py:37:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
can/io/asc.py:352:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
************* Module can.io.blf
can/io/blf.py:372:4: W1113: Keyword argument before variable positional arguments list in the definition of __init__ function (keyword-arg-before-vararg)
```

I think we should rather change these IO handlers.

can/interfaces/socketcan/socketcan.py
```diff
@@ -44,7 +38,7 @@
     LimitedDurationCyclicSendTaskABC,
 )
 from can.typechecking import CanFilters
-from can.interfaces.socketcan.constants import *  # CAN_RAW, CAN_*_FLAG
+from can.interfaces.socketcan.constants import *  # pylint: disable=unused-wildcard-import
```
felixdivo, line 41: Maybe, we should just inline that file?

can/io/generic.py
```diff
@@ -28,7 +28,7 @@ class BaseIOHandler(ContextManager, metaclass=ABCMeta):
 
     file: Optional[can.typechecking.FileLike]
 
-    def __init__(
+    def __init__(  # pylint: disable=unused-argument
```
felixdivo, line 31: Is there a point in keeping `*args` and `**kwargs` if we don't use them? Don't they just hide mistyped arguments?

felixdivo (approved):
Nice QA contribution!

can/io/trc.py
zariiii9003: The problem is: `file.newlines` returns either `None`, a string or a tuple of strings. So to make this robust, we'd have to convert None and the tuple to a string `str(file.newlines)`. But then there is no benefit in doing lazy interpolation.

can/io/generic.py
zariiii9003: I will try to remove all `*args`, i think they serve no purpose in this library. However, in most classes, CLI kwargs might get passed down, so i'd keep those even if they are unused.

can/interfaces/socketcan/socketcan.py
zariiii9003: We could, but i think that might break downstream code. How about using the constants namespace?
`from can.interfaces.socketcan import constants`

.pylintrc
zariiii9003: I agree, will do that.

can/io/generic.py
```diff
@@ -49,7 +49,9 @@ def __init__(
             # file is some path-like object
             self.file = cast(
                 can.typechecking.FileLike,
-                open(cast(can.typechecking.StringPathLike, file), mode),
+                open(  # pylint: disable=unspecified-encoding
```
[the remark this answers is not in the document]
zariiii9003: `encoding=None` did not help.

I added this instead:
```python3
            encoding: Optional[str] = (
                None
                if "b" in mode
                else kwargs.get("encoding", locale.getpreferredencoding(False))
            )
            # pylint: disable=consider-using-with
            # file is some path-like object
            self.file = cast(
                can.typechecking.FileLike,
                open(
                    cast(can.typechecking.StringPathLike, file), mode, encoding=encoding
                ),
            )
```

can/io/generic.py
j-c-cook: @zariiii9003 Sounds reasonable to me.

can/io/trc.py
zariiii9003: To be honest, the logic does not make sense at all. file.newlines will always return None here. If the author wants to set specific newline characters, it can be done with the [reconfigure](https://docs.python.org/3/library/io.html#io.TextIOWrapper.reconfigure) method.

A bit further down there is this, which is also incorrect:
```
    def _write_line(self, line: str) -> None:
        self.file.write(line + "\r\n")
```
I will try to improve this

can/io/generic.py
felixdivo: Yeah, `kwargs` are much safer 👍

.pylintrc
```diff
@@ -81,7 +81,8 @@ disable=invalid-name,
 # either give multiple identifier separated by comma (,) or put this option
 # multiple time (only on the command line, not in the configuration file where
 # it should appear only once). See also the "--disable" option for examples.
-enable=c-extension-no-member
+enable=c-extension-no-member,
+       useless-suppression,
```
felixdivo, line 85: Good idea 😁👍

felixdivo (approved):
Great @zariiii9003 !

PR #1474, opened 2023-01-01
Add `ignore_config` parameter to can.Bus

zariiii9003:


can/interface.py
```diff
@@ -71,8 +78,13 @@ class Bus(BusABC):  # pylint: disable=abstract-method
         Set to ``None`` to let it be resolved automatically from the default
         :ref:`configuration`.
 
-    :param args:
-        ``interface`` specific positional arguments.
+    :param context:
```
felixdivo, line 81: Maybe call it `config_context`?

felixdivo (approved):

can/interface.py
zariiii9003: That's exactly what i called it at first 😄 
But the existing implementation gets `context` from kwargs, so that might break user code.

zariiii9003: Done, i also added it to the deprecation decorator for whoever might need it.

PR #1475, opened 2023-01-03
make socketcan if detection more generic

1atabey1:
Automatic detection would previously fail to detect socketcan interfaces if they had been renamed. This pr changes the detection such that it checks the identifier `link/can` in the output.

This is how outputs with renamed CANs look like on my machine: 
```
4: ptcan: <NOARP,UP,LOWER_UP,ECHO> mtu 16 qdisc pfifo_fast state UP mode DEFAULT group default qlen 10\    link/can
5: pcan: <NOARP,UP,LOWER_UP,ECHO> mtu 16 qdisc pfifo_fast state UP mode DEFAULT group default qlen 10\    link/can
6: sensorcan: <NOARP,UP,LOWER_UP,ECHO> mtu 16 qdisc pfifo_fast state UP mode DEFAULT group default qlen 10\    link/can
7: hilcan: <NOARP,UP,LOWER_UP,ECHO> mtu 16 qdisc pfifo_fast state UP mode DEFAULT group default qlen 10\    link/can
```

I checked on different debian based systems and the identifier was always there so hope that's a generic thing, if someone has the capabilities to verify this please feel free to do so :)

### DIFF
--- a/can/interfaces/socketcan/utils.py
+++ b/can/interfaces/socketcan/utils.py
@@ -5,7 +5,6 @@ Defines common socketcan functions.
 import errno
 import logging
 import os
-import re
 import struct
 import subprocess
 from typing import cast, Iterable, Optional
@@ -38,14 +37,14 @@ def pack_filters(can_filters: Optional[typechecking.CanFilters] = None) -> bytes
     return struct.pack(can_filter_fmt, *filter_data)
 
 
-_PATTERN_CAN_INTERFACE = re.compile(r"(sl|v|vx)?can\d+")
-
-
 def find_available_interfaces() -> Iterable[str]:
     """Returns the names of all open can/vcan interfaces using
     the ``ip link list`` command. If the lookup fails, an error
     is logged to the console and an empty list is returned.
     """
+
+    def clean_split(string_in: str, separator: str = None):
+        return [x for x in string_in.split(separator) if x != ""]
 
     try:
         # adding "type vcan" would exclude physical can devices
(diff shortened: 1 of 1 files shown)


lumagi:
Sorry for the intrusion, I just saw your pull request. I checked the iproute2 and the kernel code. The type should be set to `link/can` for every CAN interface, the value is exported by the kernel. As an example, when the kernel creates and ISO-TP socket, it also checks if the [interface is of type CAN](https://elixir.bootlin.com/linux/latest/source/net/can/isotp.c#L1271). So this should be fairly save to use.
I noticed that `ip link` also offers a JSON output that is easier to parse and would do away with the string operations. I'm sorry to hijack this, I just thought that it could be made more robust with the JSON output. I have created a commit in my fork: https://github.com/lumagi/python-can/commit/a5a7e10c86a90dae69abe807220111d9773f4629
Which version of Debian are you running? I checked the package index and all Debian / Ubuntu versions that have the minimum required Python version 3.7 also ship an iproute2 release > 4.14.0 ([Changelog](https://lwn.net/Articles/738897/)).

zariiii9003:
I think using json is indeed more robust.

1atabey1:
Neat! Didn't know it provided that option, I agree that it's certainly the much nicer solution. 
I am mainly running ubuntu 22.04 on our systems so the iproute2 json output so it's for sure available in my usecase as well.

@lumagi how do we do this process wise? Do you just create a new pull request and I close this one or should I pull your changes into my current branch?

lumagi:
I wouldn't have write permission on your pull request. That would make it a little complicated if there's review feedback. I opened another one in #1478.

PR #1477, opened 2023-01-05
Add deprecation period to deprecated_args_alias

zariiii9003:
I added two parameters to `deprecated_args_alias` so the warning looks like this:

> The 'old_a' argument is deprecated since python-can v1.6.0, and scheduled for removal in python-can v3.4.0. Use 'a' instead.

@pierreluctg @felixdivo What do you think?

### DIFF
--- a/can/interface.py
+++ b/can/interface.py
@@ -100,8 +100,11 @@ class Bus(BusABC):  # pylint: disable=abstract-method
     """
 
     @staticmethod
-    @util.deprecated_args_alias(  # Deprecated since python-can 4.2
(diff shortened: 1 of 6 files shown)


felixdivo:
I didn't check the code, but your proposal looks very good! 😃 🚀

I generally tried to waddle out of the responsibilities of this project though, so please understand that I don't have time to go through all changes each time. 🙃

pierreluctg (approved):

hardbyte (approved):
Awesome, much nicer messages for users. Impressive testing too!

PR #1481, opened 2023-01-08
Update PCAN Basic to 4.6.2.753

zariiii9003:
@lumagi Since you have a PEAK device, could you check whether `"channel": channel.device_name + str(channel.controller_number)` is correct? 

It is used in `can.detect_available_configs("pcan")`

### DIFF
--- a/can/interfaces/pcan/basic.py
+++ b/can/interfaces/pcan/basic.py
@@ -8,15 +8,15 @@
 #
 #  ------------------------------------------------------------------
 #  Author : Keneth Wagner
+#  Last change: 2022-07-06
 #  ------------------------------------------------------------------
 #
-#  Copyright (C) 1999-2021  PEAK-System Technik GmbH, Darmstadt
+#  Copyright (C) 1999-2022  PEAK-System Technik GmbH, Darmstadt
 #  more Info at http://www.peak-system.com
 
 # Module Imports
 from ctypes import *
 from ctypes.util import find_library
-from string import *
 import platform
 
 import logging
@@ -282,6 +282,12 @@ PCAN_ATTACHED_CHANNELS_COUNT = TPCANParameter(
 PCAN_ATTACHED_CHANNELS = TPCANParameter(
     0x2B
 )  # Get information about PCAN channels attached to a system
+PCAN_ALLOW_ECHO_FRAMES = TPCANParameter(
+    0x2C
+)  # Echo messages reception status within a PCAN-Channel
+PCAN_DEVICE_PART_NUMBER = TPCANParameter(
+    0x2D
+)  # Get the part number associated to a device
 
 # DEPRECATED parameters
 #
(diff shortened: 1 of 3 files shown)


lumagi:
Sure, I can test it. But it's gonna have to wait until Tuesday because I won't have access to the hardware until then.

lumagi:
I'm sorry to keep you waiting on this. I was planning to test this on my Windows machine (at work), but I didn't get a chance to. I have reserved some time for Monday to test this.

zariiii9003:
> I'm sorry to keep you waiting on this. I was planning to test this on my Windows machine (at work), but I didn't get a chance to. I have reserved some time for Monday to test this.

It's not urgent, don't worry

lumagi:
I ran `can.detect_available_configs("pcan")` with a single PCAN USB FD attached. One problem I ran into is that `channel.device_name` is of type `bytes` instead of `str`. When I fixed that, it returned the following result:
```
[{'interface': 'pcan', 'channel': 'PCAN-USB FD0', 'supports_fd': True}]
```
It looks like `channel.device_name` does not return the required `PCAN_USBBUS` and `channel.controller_number` is zero-based.

zariiii9003:
Thank you for testing this! I guess we would need to search for TPCANChannelInformation.channel_handle in PCAN_CHANNEL_NAMES, right?

lumagi:
Yes, I agree. The `channel_handle` had a value of 0x51 in my case, which is the numeric equivalent of `PCAN_USBBUS1`. That would allow a reverse lookup. Here is the binary output of `PCAN_ATTACHED_CHANNELS`, i.e. `TPCANChannelInformation`, to allow you to write a unit test:
```
b'Q\x00\x05\x00\x01\x00\x00\x00PCAN-USB FD\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x003"\x11\x00\x01\x00\x00\x00'
```

zariiii9003:
That's great, thank you!

lumagi:
I tested again with:
```python
configs = can.detect_available_configs("pcan")
print(configs)
```
and two interfaces attached. This is what I get:
```
[{'interface': 'pcan', 'channel': 'PCAN_USBBUS1', 'supports_fd': True}, {'interface': 'pcan', 'channel': 'PCAN_USBBUS2', 'supports_fd': True}]
```
So looks nice!

One thing that came to my mind is the `device_id` in `TPCANChannelInformation`. Could that be exported as well or would that break compatibility or some constraints for the dictionary entries in the result? Since the channel name can change between reboots or between plugging in / out some adapters, the device_id provides a means to detect a specific adapter more reliably.

zariiii9003:
I think it should be okay as long as there are no conflicts with the arguments of the `__init__` method.
I also added more info for vector:
https://github.com/hardbyte/python-can/blob/69a5209edb893413d932d8881ad93b5cd599af95/can/interfaces/vector/canlib.py#L890
Do you need anything else besides device_id?

lumagi:
Good question. Off the top of my head, I would say maybe `device_type` and `channel_condition` for additional inflection needs that other people might have.

Issue #1479, opened 2023-01-06
CI skips some of the SocketCAN unit tests


lumagi:
### Describe the bug
The SocketCAN tests are only executed when the environment variable `TEST_SOCKETCAN` is present. The variable was likely introduced in the Travis CI environment and passed into test execution via tox. I noticed that it is no longer present in the current GitHub Workflow CI environment. This causes several of the SocketCAN tests to be skipped.

Updating the pipeline would include two things:
* Executing `test/open_vcan.sh` on Linux machines
* Adding the environment variable `TEST_SOCKETCAN` or removing it and simply executing these tests for all Linux systems

I know that workflow scripts have root permission on the GitHub runners. But I wasn't able to check if they also have capability `NET_ADMIN` to open vcan interfaces. Additionally, I wasn't sure if I could and didn't want to open a PR and mess with the workflow pipeline before opening an issue first. @zariiii9003  was the variable removed on purpose? If not, I can volunteer a patch, but I wanted to ask for some feedback first.

zariiii9003:
I'm no linux expert, but i think the problem is, that the github actions runner images do not contain the vcan module

lumagi:
You are correct. I did some testing, and Microsoft does not ship the vcan module with their Azure kernels. And it doesn't look like this is something they would like to change.
I did some tests and it's possible to compile the module manually. But this takes approx. 2 minutes. I guess the only option that would make sense would be to have a job that runs nightly and recompiles the module and pushes it into the repository cache. Then, the runners can pull the module from the cache and load it to open the `vcan` interfaces.
I set this up in a test repo: https://github.com/lumagi/gh_workflow_tests/tree/main/.github/workflows
But tbh, it's not entirely straight forward :-/ It relies on the fact that the kernel source code contained in the `linux-source` package is synced to the kernel version of the runner. For `ubuntu-latest` this seems to be the case because Azure uses version 5.15.0 which is the same as the version used by Ubuntu 22.04. If we were to use something else like Ubuntu 20.04 for the runner, it seems like Microsoft still runs kernel version 5.15.0 although Ubuntu 20.04 normally runs with 5.4.0.

zariiii9003:
Is the repository cache reliable? Can we assume, that it will not be cleaned randomly?

The windows pypy tests currently take ~5min.
If it's only 2 minutes we could probably integrate it into our main CI workflow. Maybe via shell script? We already have one in `test/open_vcan.sh`.

CC @hartkopp

[...]

PR #1484, opened 2023-01-10
Enable SocketCAN interface tests in GitHub CI

lumagi:
See #1479
I still need to investigate why the CPython tests fail.

### DIFF
--- a/test/test_socketcan.py
+++ b/test/test_socketcan.py
@@ -7,7 +7,15 @@ import ctypes
 import struct
 import unittest
 from unittest.mock import patch
+import can
 
+from can.interfaces.socketcan.constants import (
+    CAN_BCM_TX_DELETE,
+    CAN_BCM_TX_SETUP,
+    SETTIMER,
+    STARTTIMER,
+    TX_COUNTEVT,
+)
 from can.interfaces.socketcan.socketcan import (
     bcm_header_factory,
     build_bcm_header,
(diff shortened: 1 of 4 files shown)


lumagi:
For future reference: As of PR #1440 the default behavior for restarting Cyclic Send Tasks for SocketCAN was changed from raising an exception to not raising an exception. The test was still testing for an exception to be raised, which is no longer the case.

lumagi:
@zariiii9003 I think this should be fine now. Please check if the CI integration and the tests suit you.

test/test_socketcan.py
```diff
@@ -353,6 +355,21 @@ def test_build_bcm_update_header(self):
...
+        This test shall document raw CAN socket support under PyPy. Once this test fails, it is likely that PyPy
+        either implemented raw CAN socket support or at least changed the error that is thrown.
+        https://foss.heptapod.net/pypy/pypy/-/issues/3809
+        https://github.com/hardbyte/python-can/issues/1479
+        """
+        try:
+            can.Bus(interface="socketcan", channel="vcan0", bitrate=500000)
+        except OSError as e:
+            if "unknown address family" not in str(e):
+                raise
```
zariiii9003, line 371: This test is a good idea, but let's change this to a warning message instead of a failing CI.

zariiii9003 (approved):
Really nice addition, thank you

PR #1489, opened 2023-01-17
Add `__del__` method to `BusABC`

### LINKED ISSUE #1448: Consider using __del__
This was brought up in  #1435. We could implement `__del__` in the `BusABC` to do a best-effort cleanup and call `shutdown()` in case the user forgot to. We could also print a warning in such a case to notify the user of the improper use of the library. It is not guaranteed that the method is called, but if it is in some cases we might as well use it.

Tbruno25:
Initial PR to address #1448

Closes #1448


can/bus.py
```diff
@@ -420,14 +421,27 @@ def shutdown(self) -> None:
         Called to carry out any interface specific cleanup required
         in shutting down a bus.
         """
+        if self._is_shutdown:
+            LOG.debug(f"{self.__class__} is already shut down")
```
Tbruno25, line 425: initially I wanted to raise an exception here but there are existing unit tests that seem to want to preserve the ability to call `shutdown` multiple times without error

felixdivo: Yes, I think this needs to be possible. :) Do all subclasses of `BusABC` call `super().shutdown()`? I think this is a good point in time to make sure they do, such that they properly handly multiple calls to `shutdown()`.

can/bus.py
```diff
@@ -420,14 +421,26 @@ def shutdown(self) -> None:
         Called to carry out any interface specific cleanup required
         in shutting down a bus.
         """
```
felixdivo, line 423:
```suggestion
        Called to carry out any interface specific cleanup required in shutting down a bus.

        This method can be safely called multiple times.
        """
```

can/bus.py
```diff
@@ -420,14 +421,26 @@ def shutdown(self) -> None:
...
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.shutdown()
 
+    def __del__(self) -> None:
+        if self._is_shutdown:
+            return
+
+        self.shutdown()
+        LOG.warning("%s was not properly shut down", self.__class__)
```
felixdivo, line 442: I think this is easier:

```suggestion
        if not self._is_shutdown:
            # We do some best-effort cleanup if the user forgot to properly close the bus instance
            self.shutdown()
            LOG.warning("%s was not properly shut down", self.__class__)
```

Tbruno25: Works for me :slightly_smiling_face:

can/bus.py
Tbruno25: Added a unittest to check and preserve this behavior.

There were a couple missing such as `socketcand` and `ixxat` which I added commits for.

It's also missing from`etas` but I'm unsure whether it should call `super().shutdown()` as well? I ask because the implementation looks vastly different than the rest.

felixdivo: Regarding `etas`: I think it should not be an issue to add it here: https://github.com/hardbyte/python-can/blob/5c523ec9cc5ab3badbb6def6fb3750d228c7c7c0/can/interfaces/etas/__init__.py#L250?

felixdivo: The test is a neat idea 😄 So you also checked the code manually? Because due to imports etc. not all all interfaces might run through that test.

can/bus.py
```diff
@@ -417,17 +419,34 @@ def flush_tx_buffer(self) -> None:
...
+        try:
+            if not self._is_shutdown:
+                # We do some best-effort cleanup if the user
+                # forgot to properly close the bus instance
+                self.shutdown()
+                LOG.warning("%s was not properly shut down", self.__class__)
+        except Exception as e:  # pylint: disable=W0703
+            # Prevent unwanted output from being printed to stdout/stderr
+            LOG.debug(e)
+
```
felixdivo, line 449: Why would we want to supress exceptions there? Seems like an excpetion printed like [here](https://github.com/redis/redis-py/issues/1339#issue-614826236) would be quite useful, right?

Also a small suggestion as follows. That way, the user gets the information about the shutdown as context before the stack trace.
```suggestion
        if not self._is_shutdown:
            LOG.warning("%s was not properly shut down", self.__class__)
            # We do some best-effort cleanup if the user
            # forgot to properly close the bus instance
            try:
                self.shutdown()
            except Exception as e:  # pylint: disable=W0703
                # Prevent unwanted output from being printed to stdout/stderr
                LOG.debug(e)
```

felixdivo:
I added the `bug` label since now, for example, `etas` properly calls `stop_all_periodic_tasks()` on shutdown.

can/bus.py
Tbruno25: Thanks! I did verify within `etas` and `gs_usb` which appear to be the only 2 that are "missing coverage" so to speak. 

I will get that added to the `etas` interface :+1:

can/bus.py
Tbruno25: Hmm I think you're right. My thought was that because `__del__` _always_ gets called that it could end up adding confusion to the stack trace. 

But after testing I see that as long as `_is_shutdown` is true it won't result in double output. I need to swap lines 434 & 435 to ensure this always happens.

Tbruno25: Also without this try/except block the `tox` test report gets littered with captured warnings from existing ut's. I opted to set `pytest.filterwarnings` to ignore these.

felixdivo (approved):
This is great! I'll have to check if `pytest.mark.filterwarnings` is really a good idea, and not hiding any other problems. Else, this can be merged.

felixdivo:
> I'll have to check if `pytest.mark.filterwarnings` is really a good idea

Hey @Tbruno25, would you mind commenting on whether #1519 is a good idea?

zariiii9003:
@felixdivo i cherry-picked your changes and fixed the AttributeError. Would you like to take another look?

PR #1490, opened 2023-01-17
Improve slcan.py

mikisama:
1. Improve receiving performance
2. Fix an issue that the first read may blocking even if the `timeout` parameter is not `None`.


mikisama:
## test code
```py
bus = slcanBus(channel="COM11", bitrate=20000)

tx_msg = Message(
    arbitration_id=0x3C,
    is_extended_id=False,
    is_remote_frame=False,
    dlc=8,
    data=bytes(
        [
            0x01,
            0x02,
            0x10,
            0x01,
            0xFF,
            0xFF,
            0xFF,
            0xFF,
        ]
    ),
)

rx_msg = Message(
    arbitration_id=0x3D,
    is_extended_id=False,
    is_remote_frame=True,
    dlc=8,
)

for _ in range(16):
    bus.send(tx_msg)
    bus.send(rx_msg)
    msg = bus.recv()
    print(msg)

bus.shutdown()
```
## gap is 39ms
old
## gap is 1ms
new

```
| ID:0x3C | ID:0x3D | <-- gap(39ms) --> | ID:0x3C | ID:0x3D |

| ID:0x3C | ID:0x3D | <-- gap(1ms) ---> | ID:0x3C | ID:0x3D |
```

can/interfaces/slcan.py
```diff
@@ -161,7 +162,6 @@ def _read(self, timeout: Optional[float]) -> Optional[str]:
             while not (
                 ord(self._OK) in self._buffer or ord(self._ERROR) in self._buffer
             ):
-                self.serialPortOrig.timeout = time_left
```
zariiii9003, line 164: Why did you remove this?

mikisama: Setting the serial port timeout in the while loop slow down the RX performance.

zariiii9003: Is there a way to ensure, that the function respects the `timeout` parameter?

mikisama: > Is there a way to ensure, that the function respects the `timeout` parameter?

- This is the test code for testing the `timeout` parameter

```py
import time

from can.interfaces.slcan import slcanBus
from can.message import Message

bus = slcanBus(channel="COM11", bitrate=20000, sleep_after_open=0)

tx_msg = Message(
    arbitration_id=0x3C,
    is_extended_id=False,
    is_remote_frame=False,
    dlc=8,
    data=bytes(
        [
            0x01,
            0x02,
            0x10,
            0x01,
            0xFF,
            0xFF,
            0xFF,
            0xFF,
        ]
    ),
)

rx_msg = Message(
    arbitration_id=0x3D,
    is_extended_id=False,
    is_remote_frame=True,
    dlc=8,
)

count = 5
timeout = 1

start_time = time.time()
for _ in range(count):
    bus.send(tx_msg)
    bus.send(rx_msg)
    msg = bus.recv(timeout)
    assert msg is None
    print(msg)
end_time = time.time()
spend_time = end_time - start_time
assert abs((timeout * count) - spend_time) < 0.1
print(f"{spend_time:.3f}")

bus.shutdown()
```

- This is the result
image

Tbruno25: With this proposed change I think there's still a scenario where the timeout is not enforced.

zariiii9003: Would this work?

```python3
if timeout is not None:
    self.serialPortOrig.timeout = timeout
self._buffer += self.serialPortOrig.read_until(self._OK)
```

Tbruno25: It would work for expected messages but not errors I believe?

But I wrote a unittest that accurately catches the original error. Going to open a separate PR.

mikisama: > Would this work?
> 
> ```python
> if timeout is not None:
>     self.serialPortOrig.timeout = timeout
> self._buffer += self.serialPortOrig.read_until(self._OK)
> ```

hi, @zariiii9003 

This reminds me that the pyserial has an RX buffer inside. 

I think we can remove `self._buffer` if we use `read_until`.

This code works and it's simpler and more effective than the previous one

```py
def _read(self, timeout: Optional[float]) -> Optional[str]:

    with error_check("Could not read from serial device"):
        if timeout is not None:
            self.serialPortOrig.timeout = timeout
        buffer = self.serialPortOrig.read_until(self._OK)
        if buffer:
            return buffer.strip().decode()
```

zariiii9003: As @Tbruno25 said, we would still need to check whether there is a `self._ERROR` in the returned buffer. There even might be a `self._ERROR` and a valid message after the error.

Tbruno25: While reviewing the `pyserial` [docs](https://pyserial.readthedocs.io/en/latest/pyserial_api.html#serial.Serial.__init__) I noticed `timeout` has a non-blocking mode 

> timeout = 0: non-blocking mode, return immediately in any case, returning zero or more, up to the requested number of bytes

@mikisama It seems like setting the serial port to non-blocking before the while loop and adjusting accordingly may achieve the results you're after.

mikisama: @Tbruno25 

I try your suggestion and I found it causes higher CPU usage.

The I7-7700HQ has about 13-18% CPU usage, while the previous blocking mode had about 3% CPU usage.

```py
# pyserial non-blocking mode
self.serialPortOrig.timeout = 0
# pyserial blocking mode
# self.serialPortOrig.timeout = timeout
# first read what is already in receive buffer
while self.serialPortOrig.in_waiting:
    self._buffer += self.serialPortOrig.read()
# if we still don't have a complete message, do a blocking read
start = time.time()
time_left = timeout
while not (
    ord(self._OK) in self._buffer or ord(self._ERROR) in self._buffer
):
    byte = self.serialPortOrig.read()
    if byte:
        self._buffer += byte
    ...
```

mikisama: > As @Tbruno25 said, we would still need to check whether there is a `self._ERROR` in the returned buffer. There even might be a `self._ERROR` and a valid message after the error.

Thanks for the explanation, I now understand what @Tbruno25 means

mikisama: @Tbruno25 

Following your thoughts, I found maybe we can try `read_all`.

- The `read_all` method of `pyserial` is no-blocking too, but it also causes higher CPU usage. 
- The advantage of using the `read_all` method is that we don't have to deal with the `timeout` parameter of `serialPortOrig`

```py
with error_check("Could not read from serial device"):
    # with `read_all`, we don't have to deal with the `timeout` parameter of `serialPortOrig`
    # self.serialPortOrig.timeout = timeout
    if self.serialPortOrig.in_waiting:
        self._buffer += self.serialPortOrig.read_all()
    # if we still don't have a complete message, do a blocking read
    start = time.time()
    time_left = timeout
    while not (
        ord(self._OK) in self._buffer or ord(self._ERROR) in self._buffer
    ):
        byte = self.serialPortOrig.read_all()
        if byte:
            self._buffer += byte
        ...
```

Tbruno25: I'm not sure I agree. `read_all` appears to be omitted from the pyserial docs, but from looking at the source code it calls the underlying `read` function.

```
    def read_all(self):
        """\
        Read all bytes currently available in the buffer of the OS.
        """
        return self.read(self.in_waiting)
```

This would still be subject to violating the timeout.

mikisama: @Tbruno25 

If there is no data in the buffer, `in_waiting` is 0, and `read(0)` will return immediately.

```py
def read_all(self):
    return self.read(self.in_waiting)
```

You can try this code.

```py
import serial
import time

with serial.Serial("COM11") as ser:
    start = time.time()
    for _ in range(100):
        data = ser.read_all()
        # make sure read_all get nothing
        assert data == bytes()
    end = time.time()
    delta = end - start
    # check whether read_all is no-blocking 
    assert delta < 0.01
    print(f"{delta:.3f}")
```
And the result.
image

Tbruno25: But what about when there _is_ data in the buffer?

Please reference the test case in #1492 for an example.

zariiii9003: `read_all` sounds good to me.

```python3
    def _read(self, timeout: Optional[float]) -> Optional[str]:
        _timeout = serial.Timeout(timeout)

        with error_check("Could not read from serial device"):
            while not _timeout.expired():
                new_data = self.serialPortOrig.read_all()
                if new_data is not None:
                    self._buffer.extend(new_data)
                else:
                    time.sleep(0.001)
                    continue

                for terminator in (self._ERROR, self._OK):
                    if terminator in self._buffer:
                        i = self._buffer.index(terminator) + 1
                        string = self._buffer[:i].decode()
                        del self._buffer[:i]
                        return string

            return None
```

The serial port must be instantiated with `timeout=0`.

mikisama: > But what about when there _is_ data in the buffer? What if there were an infinite amount of 'junk'?

@Tbruno25 

```py
def read(self, __size: int = ...) -> bytes: ...
def read_all(self):
    return self.read(self.in_waiting)
```
1. If there is data in the buffer, the `read` method only copies it out. Only when the data bytes in the buffer are smaller than the parameter `__size`, the `read` method will blocking until the data bytes in the buffer is `__size`, then copy it out.
2. Since the `in_waiting` parameter shows the current number of bytes in the buffer, blocking will not occur. (The time to copy data out is negligibly small)
3. According to my test, the `MAX` receive buffer size in win10 is `16384`, and ubuntu 20.04 is `4095`. I'm not sure whether it can be changed. But I think there will not be `infinite` junk.

```py
import serial
import time

with serial.Serial("COM11") as ser:
    # send a start command to make
    # the STM32 dev board start sending data.
    ser.write(b"start")

    while True:
        time.sleep(0.5)
        print(f"in_waiting {ser.in_waiting}")
```
image

image

mikisama: @zariiii9003 

Looks good. Just a few minor changes requested.

According to my test, the `read_all` method returns `b''` if there is no data to be read out.

So the `if new_data is not None:` should be changed to `if new_data:`, otherwise it will never do `time.sleep`, which will cause high CPU usage.

```py
    def _read(self, timeout: Optional[float]) -> Optional[str]:
        _timeout = serial.Timeout(timeout)

        with error_check("Could not read from serial device"):
            while not _timeout.expired():
                new_data = self.serialPortOrig.read_all()
                if new_data:
                    self._buffer.extend(new_data)
                else:
                    time.sleep(0.001)
                    continue

                for terminator in (self._ERROR, self._OK):
                    if terminator in self._buffer:
                        i = self._buffer.index(terminator) + 1
                        string = self._buffer[:i].decode()
                        del self._buffer[:i]
                        return string

            return None
```

 >The serial port must be instantiated with timeout=0.

By the way, I did not initialize the `timeout` parameter of `serialPortOrig` at all and it defaults to `None`. It seems still works for me. 

I'm curious. If you don't set the `timeout` to 0, what will happen to you?

can/interfaces/slcan.py
```diff
@@ -150,40 +150,25 @@ def _write(self, string: str) -> None:
...
-            ):
-                byte = self.serialPortOrig.read()
-                if byte:
-                    self._buffer += byte
-                # if timeout is None, try indefinitely
-                if timeout is None:
-                    continue
-                # try next one only if there still is time, and with
-                # reduced timeout
+            while not _timeout.expired():
```
mikisama, line 156: When the `timeout` is 0, `_timeout.expired()` always `True`.

mikisama (commented):
The `timeout` parameter still needs to improve.

can/interfaces/slcan.py
zariiii9003: > I'm curious. If you don't set the timeout to 0, what will happen to you?

Nothing, i don't have an interface to test this 😄

mikisama:
@zariiii9003 

There is a slight performance difference between the `_read` method using `pyserial.read_all` and `pyserial.read`.
Any suggestions?

## Performance Comparison
```
# the original version
| ID:0x3C | ID:0x3D | <-- 39ms    ---> | ID:0x3C | ID:0x3D |

# the read_all version
| ID:0x3C | ID:0x3D | <-- 2~3.8ms ---> | ID:0x3C | ID:0x3D |

# the read version
| ID:0x3C | ID:0x3D | <-- 1~2ms   ---> | ID:0x3C | ID:0x3D |
```

## the performance test code
```py
from can.interfaces.slcan import slcanBus
from can.message import Message

bus = slcanBus(channel="COM11", bitrate=20000, sleep_after_open=0)

data_frm = Message(
    arbitration_id=0x3C,
    is_extended_id=False,
    is_remote_frame=False,
    dlc=8,
)

remote_frm = Message(
    arbitration_id=0x3D,
    is_extended_id=False,
    is_remote_frame=True,
    dlc=8,
)

for _ in range(16):
    # send a data frame
    bus.send(data_frm)
    # send a remote frame
    bus.send(remote_frm)
    # receive a data frame if a node respond the remote frame
    msg = bus.recv()
    print(msg)

# bus.shutdown()
```

## _read method using pyserial.read_all
```py
def _read(self, timeout: Optional[float]) -> Optional[str]:
    _timeout = serial.Timeout(timeout)

    with error_check("Could not read from serial device"):
        while True:
            new_data = self.serialPortOrig.read_all()
            if new_data:
                self._buffer.extend(new_data)
            else:
                if _timeout.expired():
                    break
                else:
                    time.sleep(0.001)
                    continue

            for terminator in (self._ERROR, self._OK):
                if terminator in self._buffer:
                    i = self._buffer.index(terminator) + 1
                    string = self._buffer[:i].decode()
                    del self._buffer[:i]
                    return string

        return None
```
## _read method using pyserial.read
```py
def _read(self, timeout: Optional[float]) -> Optional[str]:
    _timeout = serial.Timeout(timeout)
    self.serialPortOrig.timeout = timeout

    with error_check("Could not read from serial device"):
        while True:
            new_data = self.serialPortOrig.read()
            if new_data:
                self._buffer.extend(new_data)
            else:
                if _timeout.expired():
                    break
                else:
                    continue

            for terminator in (self._ERROR, self._OK):
                if terminator in self._buffer:
                    i = self._buffer.index(terminator) + 1
                    string = self._buffer[:i].decode()
                    del self._buffer[:i]
                    return string

        return None
```

zariiii9003:
Can you tell whether the delay is due to `time.sleep` or the `in_waiting` call of pyserial?

can/interfaces/slcan.py
Tbruno25: > According to my test, the MAX receive buffer size in win10 is 16384, and ubuntu 20.04 is 4095. I'm not sure whether it can be changed. But I think there will not be infinite junk.

Interesting! And I agree -- I wasn't considering the physical limitations of the hardware/kernel. Good catch :slightly_smiling_face:

mikisama:
> Can you tell whether the delay is due to `time.sleep` or the `in_waiting` call of pyserial?

@zariiii9003 

When I remove `time.sleep(0.001)`, their performance is almost the same. But the CPU usage is high.
Maybe `pyserial.read` is a better choice. What do you think? 🤔

zariiii9003:
Yes, i agree, your `pyserial.read` seems like a good compromise. Regarding the `sleep`, did you test this on Python 3.11? There have been improvements regards sleep precision.

mikisama:
Yes, you are right. python 3.11.1 have better performance than python 3.10.8.

```
# the read_all version with python 3.10.8
| ID:0x3C | ID:0x3D | <-- 2~3.8ms ---> | ID:0x3C | ID:0x3D |

# the read_all version with python 3.11.1
| ID:0x3C | ID:0x3D | <-- 1ms~2.4ms ---> | ID:0x3C | ID:0x3D |
```

In python 3.11.1, almost all gaps are 2ms, and some gaps are 1ms
In python 3.10.8, almost all gaps are 3ms, and some gaps are 2ms.


image

zariiii9003:
Since `self.serialPortOrig.read()` only reads a single byte, we can simplify the control flow a bit:

```python3
    def _read(self, timeout: Optional[float]) -> Optional[str]:
        _timeout = serial.Timeout(timeout)
        self.serialPortOrig.timeout = timeout

        with error_check("Could not read from serial device"):
            while True:
                new_byte = self.serialPortOrig.read(size=1)
                if new_byte:
                    self._buffer.extend(new_byte)

                if new_byte in (self._ERROR, self._OK):
                    string = self._buffer.decode()
                    del self._buffer[:]
                    return string

                if _timeout.expired():
                    break

            return None
```

I added `size=1` and renamed `new_data` so it is more obvious to the reader.

mikisama:
> Since `self.serialPortOrig.read()` only reads a single byte, we can simplify the control flow a bit:
> 
> ```python
>     def _read(self, timeout: Optional[float]) -> Optional[str]:
>         _timeout = serial.Timeout(timeout)
>         self.serialPortOrig.timeout = timeout
> 
>         with error_check("Could not read from serial device"):
>             while True:
>                 new_byte = self.serialPortOrig.read(size=1)
>                 if new_byte:
>                     self._buffer.extend(new_byte)
> 
>                 if new_byte in (self._ERROR, self._OK):
>                     string = self._buffer.decode()
>                     del self._buffer[:]
>                     return string
> 
>                 if _timeout.expired():
>                     break
> 
>             return None
> ```
> 
> I added `size=1` and renamed `new_data` so it is more obvious to the reader.

Yes, Looks better. 👍

mikisama (commented):
Still need to improve.

After the test, I found a very strange thing.

When the `timeout=0`, the previous version received fewer `None`.

 Before.
image

After change.
image

can/interfaces/slcan.py
```diff
@@ -155,21 +155,17 @@ def _read(self, timeout: Optional[float]) -> Optional[str]:
...
-                    if terminator in self._buffer:
-                        i = self._buffer.index(terminator) + 1
-                        string = self._buffer[:i].decode()
-                        del self._buffer[:i]
-                        return string
+                new_byte = self.serialPortOrig.read(size=1)
+                if new_byte:
+                    self._buffer.extend(new_byte)
+
+                if new_byte in (self._ERROR, self._OK):
```
mikisama, line 162: Since `serialPortOrig.read` only reads one byte, the SLCAN message contains multiple bytes.

It may return `None` before receiving a full SLCAN message.

mikisama (commented):
Still need to improve.

can/interfaces/slcan.py
```diff
@@ -158,18 +158,14 @@ def _read(self, timeout: Optional[float]) -> Optional[str]:
...
-                for terminator in (self._ERROR, self._OK):
-                    if terminator in self._buffer:
-                        i = self._buffer.index(terminator) + 1
+                    if new_byte in (self._ERROR, self._OK):
+                        i = self._buffer.index(new_byte) + 1
                         string = self._buffer[:i].decode()
                         del self._buffer[:i]
                         return string
+                else:
+                    if _timeout.expired():
```
mikisama, line 167: Do we really need `_timeout` to handle timeout?

The timeout is determined when `pyserial.read` returns `b''`.

zariiii9003:
You call it with timeout=0, it times out immediately. Call it with timeout=None for a blocking call. That behaviour seems correct to me.

can/interfaces/slcan.py
zariiii9003: The timeout of `pyserial.read` is reset on every `read` call, so it might add up. The timeout argument of the user should be respected, `_timeout` takes care of that.

mikisama:
> You call it with timeout=0, it times out immediately. Call it with timeout=None for a blocking call. That behaviour seems correct to me.

Yes, calling it with `timeout=None` is ok, and I rarely set timeout to 0 when I actually use it. 
But some unit tests contain `recv(0)` and all of them failed.

```py
def _read(self, timeout: Optional[float]) -> Optional[str]:
    _timeout = serial.Timeout(timeout)
    self.serialPortOrig.timeout = timeout

    with error_check("Could not read from serial device"):
        while True:
            new_byte = self.serialPortOrig.read(1)
            if new_byte:
                self._buffer.extend(new_byte)

            if new_byte in (self._ERROR, self._OK):
                string = self._buffer.decode()
                del self._buffer[:]
                return string

            if _timeout.expired():
                break

        return None
```
image

can/interfaces/slcan.py
mikisama: I think the `timeout's add up` should be ignored. 
For python on windows, the timer's resolution is about 16ms, obviously it is not very accurate.
And I think the `timeout add up` is not as big as the impact of timer jitter.

zariiii9003:
How about this? I added the `in_waiting` check before the timeout check.

```python3
    def _read(self, timeout: Optional[float]) -> Optional[str]:
        _timeout = serial.Timeout(timeout)
        self.serialPortOrig.timeout = timeout

        with error_check("Could not read from serial device"):
            while True:
                new_byte = self.serialPortOrig.read(size=1)
                if new_byte:
                    self._buffer.extend(new_byte)

                if new_byte in (self._ERROR, self._OK):
                    string = self._buffer.decode()
                    del self._buffer[:]
                    return string

                if self.serialPortOrig.in_waiting:
                    continue

                if _timeout.expired():
                    break

            return None
```

can/interfaces/slcan.py
zariiii9003: That depends on the `timeout` value, It might be seconds, or minutes, or hours, i don't know what users might be doing.

mikisama:
> How about this? I added the `in_waiting` check before the timeout check.
> 
> ```python
>     def _read(self, timeout: Optional[float]) -> Optional[str]:
>         _timeout = serial.Timeout(timeout)
>         self.serialPortOrig.timeout = timeout
> 
>         with error_check("Could not read from serial device"):
>             while True:
>                 new_byte = self.serialPortOrig.read(size=1)
>                 if new_byte:
>                     self._buffer.extend(new_byte)
> 
>                 if new_byte in (self._ERROR, self._OK):
>                     string = self._buffer.decode()
>                     del self._buffer[:]
>                     return string
> 
>                 if self.serialPortOrig.in_waiting:
>                     continue
> 
>                 if _timeout.expired():
>                     break
> 
>             return None
> ```

Unit test passed. 
Currently, I don't have a SLCAN device around. I'll have to do a test with a SLCAN device tomorrow and tell you the results.

image

mikisama:
> How about this? I added the `in_waiting` check before the timeout check.
> 
> ```python
>     def _read(self, timeout: Optional[float]) -> Optional[str]:
>         _timeout = serial.Timeout(timeout)
>         self.serialPortOrig.timeout = timeout
> 
>         with error_check("Could not read from serial device"):
>             while True:
>                 new_byte = self.serialPortOrig.read(size=1)
>                 if new_byte:
>                     self._buffer.extend(new_byte)
> 
>                 if new_byte in (self._ERROR, self._OK):
>                     string = self._buffer.decode()
>                     del self._buffer[:]
>                     return string
> 
>                 if self.serialPortOrig.in_waiting:
>                     continue
> 
>                 if _timeout.expired():
>                     break
> 
>             return None
> ```

It‘s performance is almost the same as using `pyserial.read_all()`

I think as long as we access `pyserial` `property` too often (e.g in the while loop read write `pyserial.in_waiting`, `pyserial.timeout`...), it will reduce the performance.

zariiii9003:
What if we call `in_waiting` only once?

```python3
    def _read(self, timeout: Optional[float]) -> Optional[str]:
        _timeout = serial.Timeout(timeout)
        self.serialPortOrig.timeout = timeout

        with error_check("Could not read from serial device"):
            while True:
                for _ in range(self.serialPortOrig.in_waiting):
                    new_byte = self.serialPortOrig.read(size=1)
                    if new_byte:
                        self._buffer.extend(new_byte)
                    else:
                        break
				    
                    if new_byte in (self._ERROR, self._OK):
                        string = self._buffer.decode()
                        del self._buffer[:]
                        return string

                if _timeout.expired():
                    break

            return None
```

mikisama:
> What if we call `in_waiting` only once?
> 
> ```python
>     def _read(self, timeout: Optional[float]) -> Optional[str]:
>         _timeout = serial.Timeout(timeout)
>         self.serialPortOrig.timeout = timeout
> 
>         with error_check("Could not read from serial device"):
>             while True:
>                 for _ in range(max(1, self.serialPortOrig.in_waiting)):
>                     new_byte = self.serialPortOrig.read(size=1)
>                     if new_byte:
>                         self._buffer.extend(new_byte)
> 					else:
> 					    break
> 				    
>                     if new_byte in (self._ERROR, self._OK):
>                         string = self._buffer.decode()
>                         del self._buffer[:]
>                         return string
> 
>                 if _timeout.expired():
>                     break
> 
>             return None
> ```
> 
> Edit: i added `max`, otherwise we'll cause a hot loop and high CPU usage again.


Better than `pyserial.read_all` version but not as good as **just**`pyserial.read`.
CPU usage is almost unchanged

mikisama:
I try remove the `_timeout`
```py
def _read(self, timeout: Optional[float]) -> Optional[str]:
    self.serialPortOrig.timeout = timeout

    with error_check("Could not read from serial device"):
        while True:
            new_byte = self.serialPortOrig.read(1)
            if new_byte:
                self._buffer.extend(new_byte)
                if new_byte in (self._ERROR, self._OK):
                    i = self._buffer.index(new_byte) + 1
                    string = self._buffer[:i].decode()
                    del self._buffer[:i]
                    return string
            else:
                break

        return None
```

Then run this code
```py
from can.interfaces.slcan import slcanBus
from can.message import Message

bus = slcanBus(channel="COM11", bitrate=20000, sleep_after_open=0)

data_frm = Message(
    arbitration_id=0x3C,
    is_extended_id=False,
    is_remote_frame=False,
    dlc=8,
    data=bytes(
        [
            0x01,
            0x02,
            0x10,
            0x01,
            0xFF,
            0xFF,
            0xFF,
            0xFF,
        ]
    ),
)

remote_frm = Message(
    arbitration_id=0x3D,
    is_extended_id=False,
    is_remote_frame=True,
    dlc=8,
    data=None,
)

TIMEOUT = 1
# TIMEOUT = 10
# TIMEOUT = 100

for _ in range(64):
    # send a data frame
    bus.send(data_frm)
    # send a remote frame
    bus.send(remote_frm)
    # receive a data frame if a node respond the remote frame
    # I make the stm32 device not ack the remote frame, so `recv` will always bloking and return `None`
    msg = bus.recv(TIMEOUT)
    assert msg is None
    print(msg)

# bus.shutdown()
```

`TIMEOUT=1` and we got `1006.03ms`
image

`TIMEOUT=10` and we got `10.00914s` =>`10009.14ms`
image

`TIMEOUT=100` and we got `1:40.01856` => `100018.56ms`
image

The `pyserial.timeout` was more accurate than we expected.
The timeout add-up effect is not obvious.

zariiii9003:
But how can you ensure the timeout? Try this example:

```python3
import time
from threading import Thread

from can.interfaces.slcan import slcanBus


def send_thread(_bus: slcanBus):
    while True:
        bus.serialPortOrig.write(b"f")
        time.sleep(2.0)


if __name__ == '__main__':
    bus = slcanBus(channel="loop://")
    thread = Thread(target=send_thread, args=(bus,))
    thread.start()
    while True:
        t0 = time.perf_counter()
        bus.recv(3.0)
        print(time.perf_counter() - t0)
```

It should time out every 3 seconds, but it blocks forever.

mikisama:
> But how can you ensure the timeout? Try this example:
> 
> ```python
> import time
> from threading import Thread
> 
> from can.interfaces.slcan import slcanBus
> 
> 
> def send_thread(_bus: slcanBus):
>     while True:
>         bus.serialPortOrig.write(b"f")
>         time.sleep(2.0)
> 
> 
> if __name__ == '__main__':
>     bus = slcanBus(channel="loop://")
>     thread = Thread(target=send_thread, args=(bus,))
>     thread.start()
>     while True:
>         t0 = time.perf_counter()
>         bus.recv(3.0)
>         print(time.perf_counter() - t0)
> ```
> 
> It should time out every 3 seconds, but it blocks forever.

Thanks for the example code, I finally get your point.

I noticed that there are two types of SLCAN devices categorized by how data is transmitted to the PC.
1. **Serial based**. The MCU sends the SLCAN data `byte by byte` to FT232 through the UART, then the FT232 sends the SLCAN data to PC through the USB.
```
CAN <--> MCU <--> FT232 <--> PC
```
2. **USB CDC VCOM based**. The MCU sends the `whole` SLCAN data to the PC via `one` USB transfer.
```
CAN <--> MCU <--> PC
```

Since my SLCAN device is **CDC VCOM** based, I don't understand why there is an add-up effect.😂

mikisama:
> What if we call `in_waiting` only once?
> 
> ```python
>     def _read(self, timeout: Optional[float]) -> Optional[str]:
>         _timeout = serial.Timeout(timeout)
>         self.serialPortOrig.timeout = timeout
> 
>         with error_check("Could not read from serial device"):
>             while True:
>                 for _ in range(max(1, self.serialPortOrig.in_waiting)):
>                     new_byte = self.serialPortOrig.read(size=1)
>                     if new_byte:
>                         self._buffer.extend(new_byte)
> 					else:
> 					    break
> 				    
>                     if new_byte in (self._ERROR, self._OK):
>                         string = self._buffer.decode()
>                         del self._buffer[:]
>                         return string
> 
>                 if _timeout.expired():
>                     break
> 
>             return None
> ```
> 
> Edit: i added `max`, otherwise we'll cause a hot loop and high CPU usage again.

@zariiii9003 
I'm sorry. This code works fine. I don't know why the test result was not very good yesterday.
This code is currently the best solution.

mikisama:
I change `self.serialPortOrig.timeout=timeout` to `self.serialPortOrig.timeout=0.001`
It can now handle timeout with relative accuracy.

image

mikisama:
When we call `send` to send a message, it will clear the receive buffer. Is this OK?

image

can/interfaces/slcan.py
```diff
@@ -287,7 +287,9 @@ def get_version(
 
         string = self._read(timeout)
 
-        if len(string) == 6 and string[0] == cmd:
+        if not string:
```
mikisama, line 290: Added back the `None` checking.
Due to the unit test using the `loop://` parameter. It can not catch the `None` error.

mikisama:
@zariiii9003 
Do you know what's wrong with **pypy**?
Some github CI tests failed, and they were all **pypy** based.

mikisama:
> LGTM. I will merge if you are happy with the performance.

Thanks for your review. The performance seems to be good so far.

PR #1492, opened 2023-01-20
Add unittest to verify interface `Bus.recv` timeouts

Tbruno25:
Created in response to #1490


Tbruno25:
wip -- going to implement this test for all easily loadable/instantiable bus interfaces.

PR #1496, opened 2023-01-23
Extented slcan.py to use "L" command optional to "O" command. 

wolfraven:
Added additional parameter "listen_only" to open interface/channel with "L" command (in opposite to "O" command).


felixdivo (changes requested):
Nice change, but I'd suggest using `BusState` to be more consitent. See this example on how to parse it:

https://github.com/hardbyte/python-can/blob/5c523ec9cc5ab3badbb6def6fb3750d228c7c7c0/can/interfaces/pcan/pcan.py#L248-L251

j-c-cook:
ping @wolfraven -- unaddressed requested changes.

PR #1503, opened 2023-01-26
Handle extended arbitration id properly

faisal-shah:
socketcand uses the length of the arbitration id field to indicate whether a frame is using extended id or not. 3 characters for standard, 8 for extended.

Prior to this fix, the arbitration id would be truncated to the lower 11 bits, or at times even garbage.

### DIFF
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -27,11 +27,12 @@ def convert_ascii_message_to_can_message(ascii_msg: str) -> can.Message:
         frame_string = ascii_msg[8:-2]
         parts = frame_string.split(" ", 3)
         can_id, timestamp = int(parts[0], 16), float(parts[1])
+        is_ext = (len(parts[0]) != 3)
 
         data = bytearray.fromhex(parts[2])
         can_dlc = len(data)
         can_message = can.Message(
-            timestamp=timestamp, arbitration_id=can_id, data=data, dlc=can_dlc
+            timestamp=timestamp, arbitration_id=can_id, data=data, dlc=can_dlc, is_extended_id=is_ext
         )
         return can_message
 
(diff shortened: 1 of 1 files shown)


can/interfaces/socketcand/socketcand.py
```diff
@@ -27,11 +27,12 @@ def convert_ascii_message_to_can_message(ascii_msg: str) -> can.Message:
         frame_string = ascii_msg[8:-2]
         parts = frame_string.split(" ", 3)
         can_id, timestamp = int(parts[0], 16), float(parts[1])
+        is_ext = (len(parts[0]) > 3) or (can_id > 0x7FF)
```
hartkopp, line 30: The check for `can_id > 0x7FF` is a very common wrong assumption.

Shouldn't it be: `is_ext = (len(parts[0]) == 8)` ??

faisal-shah: I tried to observe the robustness principle. This part of the code is parsing the ASCII message from socketcand. Sure, from what I have observed, extended messages should be 8 chars. However, what if due to a bug or a slight change in behavior, a message with an arbitration id > 0x7FF but less than 8 chars is received? I would say instead of dropping the frame altogether, interpret it as an extended frame.

hartkopp: As there can be EXT frames with CAN IDs 0x000 - 0x7FF too, looking at the CAN ID is simply wrong.

If you want to be more robust the check could be ` is_ext = (len(parts[0]) != 3)` .

Additionally there is no way for a _slight change_ in behavior about the length being either 3 or 8.
Believe me! I wrote that code :-D

faisal-shah: Sounds good - that's exactly the type of feedback I was looking for. Force pushed and adopted the != 3 method.

Also, please take a look at #1505 while you're at it.

faisal-shah: What do you think about using len(parts[0]) >3 instead of != 3?

hartkopp: > What do you think about using len(parts[0]) >3 instead of != 3?

In fact I would suggest != 3 as there are only 3 or 8. And when it is not 3 , it is definitely safe to store the other stuff as EXT CAN ID - even when it would be an error (that should be impossible).

faisal-shah: Ok, in that case this is ready to go. Thanks

faisal-shah:
> I'm not sure about the patch process. Are these now one patch or two? Would it make sense to squeeze them into one?

Are you referring to the other pull request #1505 ? Just in case, I combined this and #1505 into another branch and submitted #1508 (encapsulates all changes I'm making to socketcand support). You can close this (#1503) and #1505 .

On a related note @hartkopp, I think your expertise is needed over at https://github.com/linux-can/socketcand/pull/26  :) .

Thanks.

hartkopp:
See https://github.com/hardbyte/python-can/pull/1508

PR #1505, opened 2023-01-26
Populate socketcand channel and rx message attributes

faisal-shah:


hartkopp:
See https://github.com/hardbyte/python-can/pull/1508

Issue #1504, opened 2023-01-26
Arbitration ID corruption and extended id mishandling with socketcand


faisal-shah:
### Describe the bug
socketcand uses the length of the arbitration id field to indicate whether a frame is using extended id or not. 3 characters for standard, 8 for extended.

Prior to this fix, the arbitration id would be truncated to the lower 11 bits, or at times even garbage. This is because the length of the arbitration id string was not controlled. The is_extended_id attribute of the Message was also not set properly upon receiving a frame from socketcand.

### To Reproduce
Create a bridge between socketcan (vcan1) and socketcand (vcan0). Use cangen (standard) and cangen -e (extended) to generate can frames on each interface, and see the resulting mismatch in can frames. See code example below for such a bridge.

### Expected behavior
The correct arbitration id and whether or not the frame is standard or extended should be reflected on the bus.

### Additional context

<details><summary>Traceback and logs</summary>

```python
#!/usr/bin/env -S python
# test.py

import asyncio
import can

async def rx_loop_gen (bus1, bus2):
    async def loop():
        reader = can.AsyncBufferedReader()
        loop = asyncio.get_running_loop()
        notifier = can.Notifier(bus1, [reader], loop=loop)
        async for msg in reader:
            print(f"id:{msg.arbitration_id:x}, ext={msg.is_extended_id}")
            bus2.send(msg)

    return loop


async def main():
    import argparse
    parser = argparse.ArgumentParser(description='python-can to socketcand bridge')
    parser.add_argument('-i', '--interface', required=False, default='socketcan')
    parser.add_argument('-c', '--channel', required=False, default='vcan1')
    parser.add_argument('-a', '--addr', required=False, default = 'localhost')
    parser.add_argument('-p', '--port', required=False, default = 29536)
    args = parser.parse_args()

    bus_pycan = can.Bus(interface=args.interface, channel=args.channel)
    bus_socketcand = can.Bus(interface='socketcand', host=args.addr, port = args.port, channel='vcan0')

    l1 = await rx_loop_gen(bus_pycan, bus_socketcand)
    l2 = await rx_loop_gen(bus_socketcand, bus_pycan)

    await asyncio.gather(l1(), l2())

if __name__ == '__main__':
    try:
        asyncio.run(main())
    except KeyboardInterrupt:
        pass

```

vcan0: socketcand
vcan1: socketcan
test.py above bridges those two interfaces
```
$ cangen -e vcan1 # This is generating extended frames on the socketcan device
```

[...]

PR #1508, opened 2023-01-27
Socketcand ext ID bug fixes, and implement channel{_info} and Message attributes

faisal-shah:
See #1504 for bug description

This pull request combines:
#1505
#1503

### DIFF
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -27,11 +27,16 @@ def convert_ascii_message_to_can_message(ascii_msg: str) -> can.Message:
         frame_string = ascii_msg[8:-2]
         parts = frame_string.split(" ", 3)
         can_id, timestamp = int(parts[0], 16), float(parts[1])
+        is_ext = len(parts[0]) != 3
 
         data = bytearray.fromhex(parts[2])
         can_dlc = len(data)
         can_message = can.Message(
-            timestamp=timestamp, arbitration_id=can_id, data=data, dlc=can_dlc
+            timestamp=timestamp,
+            arbitration_id=can_id,
+            data=data,
+            dlc=can_dlc,
+            is_extended_id=is_ext,
         )
         return can_message
 
@@ -40,11 +45,15 @@ def convert_can_message_to_ascii_message(can_message: can.Message) -> str:
     # Note: socketcan bus adds extended flag, remote_frame_flag & error_flag to id
     # not sure if that is necessary here
     can_id = can_message.arbitration_id
+    if can_message.is_extended_id:
+        can_id_string = f"{(can_id&0x1FFFFFFF):08X}"
+    else:
+        can_id_string = f"{(can_id&0x7FF):03X}"
     # Note: seems like we cannot add CANFD_BRS (bitrate_switch) and CANFD_ESI (error_state_indicator) flags
     data = can_message.data
     length = can_message.dlc
     bytes_string = " ".join(f"{x:x}" for x in data[0:length])
-    return f"< send {can_id:X} {length:X} {bytes_string} >"
+    return f"< send {can_id_string} {length:X} {bytes_string} >"
 
 
 def connect_to_server(s, host, port):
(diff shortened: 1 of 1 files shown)


can/interfaces/socketcand/socketcand.py
```diff
@@ -40,11 +45,15 @@ def convert_can_message_to_ascii_message(can_message: can.Message) -> str:
     # Note: socketcan bus adds extended flag, remote_frame_flag & error_flag to id
     # not sure if that is necessary here
     can_id = can_message.arbitration_id
+    if can_message.is_extended_id or (can_id > 0x7FF):
```
hartkopp, line 48: I'm sorry but this check for `(can_id > 0x7FF)` is still wrong.

If you want to create robust code you might mask the can_id after the check for `is_extended_id`.
(masking ext ID with & 1FFFFFFF and a 11 bit ID with & 0x7FF)

faisal-shah: Absolutely correct .. fixed and force pushed.

faisal-shah: I added Message.is_rx attribute in the convert_ascii_message_to_can_message function, instead of after it is called.

hartkopp:
@zariiii9003 I'm definitely fine with https://github.com/hardbyte/python-can/pull/1508/commits/6ddbfa9561b74f9e6aabde4c0b02a5aa29dcf9cf now. Can you please check (and commit?) the two more Python specific patches? Thx!

can/interfaces/socketcand/socketcand.py
```diff
@@ -40,11 +46,15 @@ def convert_can_message_to_ascii_message(can_message: can.Message) -> str:
...
     # not sure if that is necessary here
     can_id = can_message.arbitration_id
+    if can_message.is_extended_id:
+        can_id_string = f"{(can_id&0x1FFFFFFF):08X}"
+    else:
+        can_id_string = f"{(can_id&0x7FF):03X}"
     # Note: seems like we cannot add CANFD_BRS (bitrate_switch) and CANFD_ESI (error_state_indicator) flags
     data = can_message.data
     length = can_message.dlc
     bytes_string = " ".join(f"{x:x}" for x in data[0:length])
```
zariiii9003, line 56: @hartkopp Shouldn't the hex bytes here be zero-padded, e.g. `0c` instead of `c`?
If so, the format string should be `f"{x:02x}"`

hartkopp: Doesn't `f"{(can_id&0x1FFFFFFF):08X}"` mean it is zero padded with a total length of 8 -> `00000123` ??
Because  this is what would be needed here.

`f"{x:02x}"` would be the right approach for data bytes (not CAN identifier), right?

zariiii9003: I'm talking about the data bytes here:
```python3
bytes_string = " ".join(f"{x:x}" for x in data[0:length])
```

For example, should it be `2 4 6 8 a` or `02 04 06 08 0a`? Does it make a difference for socketcand? Or is it okay as long as there is spacing between the bytes?

faisal-shah: It's supposed to be lowercase unpadded (%hhx):
https://github.com/linux-can/socketcand/blob/master/state_raw.c#L158-L171

faisal-shah: And on the receive side (< frame > message), the data is padded uppercase (%02X) with no spaces in between:
https://github.com/linux-can/socketcand/blob/master/state_raw.c#L125-L135

hartkopp: Ouch! I wasn't aware of this inconsistency.
But I assume changing this in socketcand is not the way to go as it would potentially break other applications.

zariiii9003: Alright, then i will merge this.

zariiii9003 (approved):

can/interfaces/socketcand/socketcand.py
faisal-shah: > Ouch! I wasn't aware of this inconsistency.
> But I assume changing this in socketcand is not the way to go as it would potentially break other applications.

Sent PR to correct the docs in socketcand:

https://github.com/linux-can/socketcand/pull/27

PR #1469, opened 2022-12-25
Add BitTiming/BitTimingFd support to PcanBus

zariiii9003:
Depends on #1468

PR #1514, opened 2023-01-28
Implement BitTiming/BitTimingFD handling for PCAN Bus

lumagi:
Continues work from #1469
It is marked as draft until I have time to test it next week.


lumagi:
I did some tests over the weekend, and it looks like the timing parameters are passed properly to the PCAN API.

zariiii9003 (approved):
Thank you, looks good 👍

PR #1528, opened 2023-02-27
Support CANdapter extended length arbitration ID

### LINKED ISSUE #1506: CANdapter Support for Extended Arbitration ID's (slcan ASCII Protocol)
Hello,

I am integrating the [Ewert Energy CANdapter](https://www.ewertenergy.com/products/candapter/downloads/candapter_manual.pdf). The protocol ASCII protocol seems to be extremely similar to what is device under the `slcan` interface. The `slcan` interface works for standard length arbitration ID's. 

The extended length ID's however, were not working at first. The CANdapter reports the first ASCII character as `x`. I was able to integrate without any issues by changing the `T` to the `x` in the linked line. 
https://github.com/hardbyte/python-can/blob/dcf15c962786b969fa45232e176ed02eabc889b4/can/interfaces/slcan.py#L213

I would be interested in determining a way for these characters to be a part of the configuration, and if this specific adapter is unique in how it implements this "protocol".

Cheers,
Jack

BroderickJack:
Add support for [Ewert Energy CANdapter](https://www.ewertenergy.com/products/candapter/downloads/candapter_manual.pdf) extended arbitration ID ASCII identifier of `x` for the slcan interface implementation.

Fixes #1506

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -205,7 +205,7 @@ class slcanBus(BusABC):
 
         if not string:
             pass
-        elif string[0] == "T" or string[0] == "x":
+        elif string[0] in ("T", "x"):
             # extended frame
             canId = int(string[1:9], 16)
             dlc = int(string[9])
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -205,7 +205,7 @@ class slcanBus(BusABC):
 
         if not string:
             pass
-        elif string[0] in ("T", "x"):
+        elif string[0] in ("T", "x"): # x is an alternative extended message identifier for CANDapter
             # extended frame
             canId = int(string[1:9], 16)
             dlc = int(string[9])
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -205,7 +205,7 @@ class slcanBus(BusABC):
 
         if not string:
             pass
-        elif string[0] == "T":
+        elif string[0] == "T" or string[0] == "x":
             # extended frame
             canId = int(string[1:9], 16)
             dlc = int(string[9])
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -205,7 +205,10 @@ class slcanBus(BusABC):
 
         if not string:
             pass
-        elif string[0] in ("T", "x"): # x is an alternative extended message identifier for CANDapter
+        elif string[0] in (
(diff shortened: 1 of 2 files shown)


can/interfaces/slcan.py
```diff
@@ -203,7 +203,7 @@ def _recv_internal(
 
         if not string:
             pass
-        elif string[0] == "T":
+        elif string[0] == "T" or string[0] == "x":
```
zariiii9003, line 206:
```suggestion
        elif string[0] in ("T", "x"):
```
You should also add a comment that explains, where the x comes from.
A test would be great too.

can/interfaces/slcan.py
NoahMaceri, line 1: @BroderickJack example comment
`elif string[0] == "T" or string[0] == "x":  # x is an alternative extended message identifier from the CANDapter`

NoahMaceri: Also in test_slcan.py you can insert this test after line 40 for a test
```
# Ewert Energy Systems CANDapter specific
elf.serial.write(b"x12ABCDEF2AA55\r")
msg = self.bus.recv(TIMEOUT)
self.assertIsNotNone(msg)
self.assertEqual(msg.arbitration_id, 0x12ABCDEF)
self.assertEqual(msg.is_extended_id, True)
self.assertEqual(msg.is_remote_frame, False)
self.assertEqual(msg.dlc, 2)
self.assertSequenceEqual(msg.data, [0xAA, 0x55])
```

BroderickJack: Got it, thank you!

test/test_slcan.py
```diff
@@ -38,6 +38,16 @@ def test_recv_extended(self):
         self.assertEqual(msg.dlc, 2)
         self.assertSequenceEqual(msg.data, [0xAA, 0x55])
 
+        # Ewert Energy Systems CANDapter specific
+        elf.serial.write(b"x12ABCDEF2AA55\r")
```
zariiii9003, line 42:
```suggestion
        self.serial.write(b"x12ABCDEF2AA55\r")
```

NoahMaceri: That's my bad, sorry!

BroderickJack: no worries, I definitely should've caught that too   ¯\_(ツ)_/¯

zariiii9003:
@BroderickJack can you fix the failing test?

PR #1532, opened 2023-03-05
Add `protocol` property to BusABC to determine active CAN Protocol

lumagi:
Something that I have been missing in the past about the generic `BusABC` is to introspect CAN-FD support independent of the bus implementation. The way I would imagine this is to simply have an attribute called `is_fd` or something similar that will return `True` if the bus supports the CAN-FD protocol.

As an example, I have implemented this for the PCAN bus to gather some feedback. If this gets support, I will implement it for all interfaces.


zariiii9003:
Just to clarify: you say

> return True if the bus supports the CAN-FD protocol

but you actually just want to know whether CAN FD is currently active, right?

Generally i don't mind standardizing this for all interfaces, but with CAN XL coming soon it might make sense to use an enum instead of a boolean flag. Otherwise we would need to add `is_xl` soon.

zariiii9003:
CAN FD is already part of ISO 11898 and CAN XL will also be part of 11898. So instead of `CAN_ISO_11898` i'd suggest `CAN_20` or something like that. Do you have any ideas for the class name? `CanSpec`? `Protocol`? `PhysicalLayer`? 

@hardbyte not sure how to contact you but i recommend @lumagi as maintainer ;)

hardbyte:
@lumagi - Lukas thanks so much for your contributions so far. As suggested I've invited you to the repository as a collaborator. Use your new powers for good! In essence, please be sensible and request reviews for non-trivial changes 👍

Great idea @zariiii9003!

lumagi:
@hardbyte Thank you for the invitation, I appreciate it! And thanks for the recommendation @zariiii9003  ;)

lumagi:
Regarding the naming of the enum, I think I like `Protocol`. But I would like make it a bit more specific like `CANProtocol`, otherwise this might clash with a different symbol when importing it into the module scope.

zariiii9003:
[Here](https://www.can-cia.org/can-knowledge/can/can-xl/) is an interesting sentence

> The third generation of the CAN data link layer supports all three protocol types (Classical CAN, CAN FD, and CAN XL)

So it is either "DataLinkLayer" or "Protocol". I'm not sure which one is technically correct in our case. I prefer `CanProtocol` over `BusProtocol` btw if we take the protocol option.

lumagi:
Ok, to me either one would be fine: `CANLinkLayer`, `CANProtocol`. But I wouldn't make the names any longer.

can/interfaces/udp_multicast/bus.py
```diff
@@ -103,7 +102,9 @@ def __init__(
                 "receiving own messages is not yet implemented"
             )
 
-        super().__init__(channel, **kwargs)
+        super().__init__(
+            channel, **kwargs, protocol=CANProtocol.CAN_FD if fd else CANProtocol.CAN_20
+        )
 
         self.is_fd = fd
```
lumagi, line 109: @zariiii9003 Would you consider this `is_fd` flag as a public attribute that would lead to a breaking API change if it were removed? With the `protocol` field it's technically redundant and could be removed.

zariiii9003: I guess we could replace it with a property and raise a DeprecationWarning with a removal scheduled for version 5.

zariiii9003:
> Ok, to me either one would be fine: `CANLinkLayer`, `CANProtocol`. But I wouldn't want to make the names too long.

I just looked through the rest of the code, can you rename it to `CanProtocol`? That's more in line with the rest of our code:
image

lumagi:
Done

zariiii9003:
That's a big one. We'll probably have to release 4.2 and then leave this one in develop for some time 😄

lumagi:
Yes sure, I don't mind. I only have a couple Bus implementations left, so this can probably change state away from draft soon. But I imagine there's gonna be some feedback :D

zariiii9003:
I resolved all the merge conflicts from #1551. I hope i didn't break anything.

lumagi:
Thanks for resolving the conflicts! And also the rogue black test - It was not able to come to terms with my local black version.

zariiii9003:
> Thanks for resolving the conflicts! And also the rogue black test - It was not able to come to terms with my local black version.

You could install the correct versions with `pip install -r requirements-lint.txt`

lumagi:
@zariiii9003 I think I'm through with all the implementations, if you would like to take a look.

can/bus.py
```diff
@@ -49,6 +57,7 @@ class BusABC(metaclass=ABCMeta):
     def __init__(
         self,
         channel: Any,
+        protocol: CanProtocol = CanProtocol.CAN_20,
```
zariiii9003, line 60: I'm not sure about this parameter here. I think subclasses should keep the parameters of its superclasses, which wouldn't make sense in this case.

What do you think about declaring `protocol` as a class attribute?

lumagi: Maybe as a protected instance attribute, i.e. (self._protocol)? If it was a class attribute, then it would have the same value for all instances, right?

zariiii9003: We could set the default value of `_protocol` as a class attribute. In the `__init__` method, the new value is assigned to the instance, not the class. So multiple instances will not share the same value.

lumagi: Not passing it in the super constructor but setting it directly sounds like a good plan. I assume you want to set it as a class attribute as a fallback in case it is not set in the subclass constructor? How would you be able to set it as an instance attribute if it is already present as a class attribute?

zariiii9003:
```python3
class A:
    protocol = "Class Attribute"


class B(A):
    def __init__(self):
        self.protocol = "Instance Attribute"
        super().__init__()


print(A().protocol)
print(B().protocol)
```

This prints 
```text
Class Attribute
Instance Attribute
```

lumagi: Great, I wasn't aware that the way the attribute was accessed for assignment changed the assignment scope. I will change the implementation to follow this approach.

zariiii9003: @lumagi tell me if you need help with the changes

lumagi: Thanks for the offer. I have been a bit preoccupied with my time the last weeks, but I should be able to make some progress this week.

lumagi: The attribute should now be set inside the subclass constructor for all classes.

can/interfaces/pcan/pcan.py
```diff
@@ -482,7 +490,7 @@ def _recv_internal(
         end_time = time.time() + timeout if timeout is not None else None
 
         while True:
-            if self.fd:
+            if self.protocol == CanProtocol.CAN_FD:
```
zariiii9003, line 493:
```suggestion
            if self._can_protocol is CanProtocol.CAN_FD:
```
properties are rather slow, and enums are compared with `is`

can/interfaces/pcan/pcan.py
```diff
@@ -590,7 +598,7 @@ def send(self, msg, timeout=None):
         if msg.error_state_indicator:
             msgType |= PCAN_MESSAGE_ESI.value
 
-        if self.fd:
+        if self.protocol == CanProtocol.CAN_FD:
```
zariiii9003, line 601:
```suggestion
        if self._can_protocol is CanProtocol.CAN_FD:
```

can/interfaces/udp_multicast/bus.py
```diff
@@ -122,13 +136,13 @@ def _recv_internal(self, timeout: Optional[float]):
                 "could not unpack received message"
             ) from exception
 
-        if not self.is_fd and can_message.is_fd:
+        if self.protocol != CanProtocol.CAN_FD and can_message.is_fd:
```
zariiii9003, line 139:
```suggestion
        if self._can_protocol is not CanProtocol.CAN_FD and can_message.is_fd:
```

can/interfaces/udp_multicast/bus.py
```diff
@@ -122,13 +136,13 @@ def _recv_internal(self, timeout: Optional[float]):
...
             return None, False
 
         return can_message, False
 
     def send(self, msg: can.Message, timeout: Optional[float] = None) -> None:
-        if not self.is_fd and msg.is_fd:
+        if self.protocol != CanProtocol.CAN_FD and msg.is_fd:
```
zariiii9003, line 145:
```suggestion
        if self._can_protocol is not CanProtocol.CAN_FD: and msg.is_fd:
```

can/interfaces/vector/canlib.py
```diff
@@ -647,7 +663,7 @@ def _recv_internal(
 
         while True:
             try:
-                if self.fd:
+                if self.protocol == CanProtocol.CAN_FD:
```
zariiii9003, line 666:
```suggestion
                if self._can_protocol is CanProtocol.CAN_FD:
```

can/interfaces/vector/canlib.py
```diff
@@ -781,7 +797,7 @@ def send(self, msg: Message, timeout: Optional[float] = None) -> None:
 
     def _send_sequence(self, msgs: Sequence[Message]) -> int:
         """Send messages and return number of successful transmissions."""
-        if self.fd:
+        if self.protocol == CanProtocol.CAN_FD:
```
zariiii9003, line 800:
```suggestion
        if self._can_protocol is CanProtocol.CAN_FD:
```

zariiii9003 (approved):
Looks great! The enum comparisons should be fixed though 😄

lumagi:
Great, thanks for the review! How many reviewers do you normally have for greater changes like this one?

zariiii9003:
> Great, thanks for the review! How many reviewers do you normally have for greater changes like this one?

At some point 2 reviews were required. But we had more active maintainers at that point. I will merge this after a 4.2.1 bugfix release to avoid backporting headaches.

lumagi:
@zariiii9003 When do you think this will make it into a release? Will this have to wait until a major release or will this go into the next minor? To be honest, I am quite looking forward to having this feature available :D

zariiii9003:
@lumagi I was hoping to include #1611. But i could release 4.3 anytime. I'll try to get it ready by the end of the week.
You could accelerate it by updating the changelog to include all relevant changes since 4.2.2 😄

lumagi:
Sure, I can take care of that. I'll probably get around to doing that over the weekend.

PR #1539, opened 2023-03-13
Improve ThreadBasedCyclicSendTask timing

zariiii9003:


can/broadcastmanager.py
```diff
@@ -5,30 +5,39 @@
...
 
 # try to import win32event for event-based cyclic send task (needs the pywin32 package)
+USE_WINDOWS_EVENTS = False
 try:
     import win32event
 
-    HAS_EVENTS = True
+    # Python 3.11 provides a more precise sleep implementation on Windows, so this is not necessary.
+    # Put version check here, so mypy does not complain about `win32event` not being defined.
+    if sys.version_info < (3, 11):
```
pierreluctg, line 31: Why do we event try to import win32event on Python 3.11+?

zariiii9003: Because i couldn't get mypy to shut up 👀

pierreluctg (approved):

lumagi:
I'm sorry I couldn't support here, my week was too busy.

PR #1556, opened 2023-04-04
Optional dependency and docs for the CANine interface

yconst:


zariiii9003 (approved):
looks good, thank you :)

PR #1551, opened 2023-03-30
Export symbols to satisfy type checkers

zariiii9003:
Closes #1547 

This also introduces ruff with isort, pyupgrade and the unused-imports check.

@felixdivo The version with the `import x as x` reexport can be seen [here](https://github.com/zariiii9003/python-can/blob/81a610820fa34c71461e6027334193b55c50043e/can/__init__.py). I think that one is less ugly, but i don't care too much.

PR #1557, opened 2023-04-04
Improve can.Bus typing

zariiii9003:
Convert `can.Bus` into a function to clean up a lot of the messy typing. This should fix the [pyright abstract class complaints](https://github.com/hardbyte/python-can/pull/1551#issuecomment-1495974163).

I had to move the `can.BusABC` documentation from `internal-api.rst` to `bus.rst`, the private methods of `can.BusABC` are still documented in `internal-api.rst`.

### DIFF
--- a/can/interface.py
+++ b/can/interface.py
@@ -55,8 +55,20 @@ def _get_class_for_interface(interface: str) -> Type[BusABC]:
     return cast(Type[BusABC], bus_class)
 
 
-class Bus(BusABC):  # pylint: disable=abstract-method
-    """Bus wrapper with configuration loading.
+@util.deprecated_args_alias(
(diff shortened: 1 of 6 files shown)


lumagi (commented):
I took the liberty of reviewing the PR. I think this is a really nice change. I know that having both `Bus` and `BusABC` was something that confused me when I first started using python-can.

zariiii9003:
Thank you for taking a look. I will wait with merging until 4.2 is released.
I had the same experience with `Bus` and `BusABC`. 😄 

I'd like to hear more opinions on this though.

hardbyte (approved):
Great job. I remember really struggling with the Bus/BusABC design many many years ago!

PR #1566, opened 2023-04-13
Enable CanFD support in "serial" interface

jacky309:
* Remove (DLC <= 8) limitation

### DIFF
--- a/can/interfaces/serial/serial_can.py
+++ b/can/interfaces/serial/serial_can.py
@@ -179,8 +179,6 @@ class SerialBus(BusABC):
                 s = self._ser.read(4)
                 timestamp = struct.unpack("<I", s)[0]
                 dlc = ord(self._ser.read())
-                if dlc > 8:
-                    raise ValueError("received DLC may not exceed 8 bytes")
 
                 s = self._ser.read(4)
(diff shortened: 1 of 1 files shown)


jacky309:
Any feedback regarding this pull-request ?

lumagi:
Hey @jacky309,

your PR does not take into consideration the fact that a regular CAN frame shouldn't have a payload larger than 8 Byte. If you want to send CAN-FD frames over the interface, I recommend adding proper CAN-FD support to the interface.
This would entail the following points:
* Adding a flag field to the transmission frame
* sending up to 8 byte if frame.is_fd == False, sending and receiving up to 64 byte if frame.is_fd is True

jacky309:
Do you mean that the serial frame format described at https://python-can.readthedocs.io/en/master/interfaces/serial.html should be changed to introduce a new field to identify if a frame is CanFD or not ? This would be a breaking change which is likely to cause some trouble with existing software, right ? IMO, this is not acceptable.
This PR, as it is now, introduces the improvement that CanFD frames can be transmitted over a serial connection, without requiring any change to the serial frame format. IMO, this benefit outweights the fact that the length check is gone, so I would consider merging this PR for now.

PR #1572, opened 2023-04-21
Fix inefficiencies, causes write delays to drop

silverv:
Did some experiments and measurements to conclude that these changes introduce less write delay. It is also likely that CPU load will be decreased with these changes.

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -49,8 +49,8 @@ class slcanBus(BusABC):
 
     _SLEEP_AFTER_SERIAL_OPEN = 2  # in seconds
 
-    _OK = b"\r"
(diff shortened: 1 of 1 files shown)


silverv:
On Windows it is important to note that a REALTIME priority level can be assigned to the process.

silverv:
This is a continuation of #1490

silverv:
I am doing this to improve the performance of SLCAN in Yakut (https://github.com/OpenCyphal/yakut) and Yukon (https://github.com/OpenCyphal/yukon).

zariiii9003:
With your changes the `timeout` parameter will do nothing.
Also, your title mentions write delays, but you change the read function?

PR #1590, opened 2023-05-11
With Windows event the first two periodic frames are sent without delay

pierreluctg:
With Windows event the first two periodic frames are sent without delay

When the `ThreadBasedCyclicSendTask` thread starts, the timer event was already set before entering the first loop. This resulted in the first timer wait to not wait.

Sample code to show the issue:
```python
import time

import can
from can import Message, Notifier, BufferedReader, broadcastmanager


class MyListener(BufferedReader):
    def __init__(self, msg, period):
        super().__init__()
        self.msg = msg
        self.period = period

    def on_message_received(self, msg: Message) -> None:
        if msg.arbitration_id == self.msg.arbitration_id:
            super().on_message_received(msg)


def main():
    # Set False to test without Windows events
    # broadcastmanager.USE_WINDOWS_EVENTS = False

    msg = Message(
        arbitration_id=0x100,
        is_extended_id=False,
        data=(0, ) * 8
    )
    period = 5
    duration = 10

    with can.Bus() as bus:
        listener = MyListener(msg, period)
        notifier = Notifier(bus, (listener,))
        bus.send_periodic(msg, period=period, duration=duration)
        time.sleep(duration)
        notifier.stop()

        frames = []
        while not listener.buffer.empty():
            frames.append(listener.buffer.get())

        print(f"Found {len(frames)} frame(s)")
        for i in range(len(frames)):
            if i > 0:
                print(f"Delta: {frames[i].timestamp - frames[i-1].timestamp}")
            print(frames[i])

        assert abs(frames[1].timestamp - frames[0].timestamp) > period/2.0


if __name__ == "__main__":
    main()

```

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -293,6 +293,10 @@ class ThreadBasedCyclicSendTask(
         msg_index = 0
         msg_due_time_ns = time.perf_counter_ns()
 
+        if USE_WINDOWS_EVENTS:
+            # Make sure the timer is non-signaled before entering the loop
+            win32event.WaitForSingleObject(self.event.handle, 0)
+
         while not self.stopped:
             # Prevent calling bus.send from multiple threads
             with self.send_lock:


can/broadcastmanager.py
```diff
@@ -293,6 +293,10 @@ def _run(self) -> None:
         msg_index = 0
         msg_due_time_ns = time.perf_counter_ns()
 
+        if USE_WINDOWS_EVENTS:
+            # Make sure the timer is non-signaled before entering the loop
+            win32event.WaitForSingleObject(self.event.handle, 0)
```
zariiii9003, line 298: Maybe we could call SetWaitableTimer here instead of in `start()`?

pierreluctg: Sure that should be possible, in another PR maybe?

zariiii9003: I meant SetWaitableTimer instead of WaitForSingleObject. Wouldn't that solve this issue?

pierreluctg: This would give us essentially what the current code with the bug is...

zariiii9003: Apparently the handle is signaled immediately, if the duetime parameter is 0.

Can you try this?
```python3
    def start(self) -> None:
        self.stopped = False
        if self.thread is None or not self.thread.is_alive():
            name = f"Cyclic send task for 0x{self.messages[0].arbitration_id:X}"
            self.thread = threading.Thread(target=self._run, name=name)
            self.thread.daemon = True

            if USE_WINDOWS_EVENTS:
                win32event.SetWaitableTimer(
                    self.event.handle,
                    -1 * self.period_ns // 100,
                    self.period_ms,
                    None,
                    None,
                    False,
                )

            self.thread.start()
```

zariiii9003 (approved):

PR #1597, opened 2023-05-16
Raise Minimum Python Version to 3.8

### LINKED ISSUE #1234: Remove use of pkg_resources when not supporting Python 3.7 any more
Currently, a fallback to `pkg_resources.iter_entry_points` is used in `can/interfaces/__init__.py`. However, the [setuptools docs](https://setuptools.pypa.io/en/latest/pkg_resources.html) state:

> Use of pkg_resources is discouraged in favor of importlib.resources, importlib.metadata, and their backports (resources, metadata). Please consider using those libraries instead of pkg_resources.

On Python 3.8+, `importlib.metadata` is [part of the standard library](https://docs.python.org/3/library/importlib.metadata.html), so there should be no need to keep `pkg_resources.iter_entry_points` around, right?

The PR #1110 is related.

zariiii9003:
Also, remove pkg_resources and the setuptools runtime dependency. Closes #1234


hardbyte (approved):
Great job.

I didn't dig deep into the entrypoint changes but everything else looks good to me.

PR #1601, opened 2023-05-20
Fixed bug that ASCReader doesn't show messages with negative timestamp

### LINKED ISSUE #1595: ASCReader doesn't show messages with negative timestamp
### Describe the bug

ASCReader doesn't show messages with negative timestamp

### To Reproduce

open .asc-file which contains messages with negative timestamps
print all messages from the.asc-file

### Expected behavior

every messages gets printed

### Additional context

OS and version: Win
Python version: 3.7
python-can version: 4.2.0

```python
def func():
    return "hello, world!"
```
</details>

ghost:
Changed the regex in the ASCReader in line 276 so the timestamp doesn't get ignored.
Just added [-]? to the regex.

Closes #1595

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -273,7 +273,7 @@ class ASCReader(TextIOMessageReader):
                 continue
 
             if not re.match(
-                r"\d+\.\d+\s+(\d+\s+(\w+\s+(Tx|Rx)|ErrorFrame)|CANFD)",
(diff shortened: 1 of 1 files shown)


hartkopp:
I'm not sure about this fix: https://github.com/hardbyte/python-can/issues/1595#issuecomment-1555867082

PR #1126, opened 2021-09-10
feat: extend IXXAT backend to support CAN-FD

ausserlesh:
Introduce CAN-FD support for IXXAT using C-API for CAN-FD vcinpl2.dll.

PR #1607, opened 2023-05-28
Implement _detect_available_configs for the Ixxat bus.

MattWoodhead:
Tested on local hardware, although I only have single channel devices so have not been able to validate that the multichannel detection works correctly.

I only realised once writing the pull request that there is some overlap with #1180 in the changes I have made. I have submitted a seperate PR anyway as I believe that the solution in these commits is much more in line with the behaviour of the rest of the library, but please feel free to close if you wish to progress the other PR.

### DIFF
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -156,9 +156,7 @@ class IXXATBus(BusABC):
         duration: Optional[float] = None,
         modifier_callback: Optional[Callable[[Message], None]] = None,
     ) -> CyclicSendTaskABC:
-        return self.bus._send_periodic_internal(
-            msgs, period, duration, modifier_callback
-        )
+        return self.bus._send_periodic_internal(msgs, period, duration, modifier_callback)
 
     def shutdown(self) -> None:
         super().shutdown()
@@ -170,3 +168,7 @@ class IXXATBus(BusABC):
         Return the current state of the hardware
         """
         return self.bus.state
+
+    @staticmethod
+    def _detect_available_configs():
+        return vcinpl._detect_available_configs()
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -1,4 +1,4 @@
-from typing import Callable, Optional, Sequence, Union
+from typing import Callable, Optional, Sequence, Union, List
 
 import can.interfaces.ixxat.canlib_vcinpl as vcinpl
 import can.interfaces.ixxat.canlib_vcinpl2 as vcinpl2
@@ -8,6 +8,7 @@ from can import (
     CyclicSendTaskABC,
     Message,
 )
+from can.typechecking import AutoDetectedConfig
 
 
 class IXXATBus(BusABC):
@@ -170,5 +171,5 @@ class IXXATBus(BusABC):
         return self.bus.state
 
     @staticmethod
-    def _detect_available_configs():
+    def _detect_available_configs() -> List[AutoDetectedConfig]:
         return vcinpl._detect_available_configs()
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -157,7 +157,9 @@ class IXXATBus(BusABC):
         duration: Optional[float] = None,
         modifier_callback: Optional[Callable[[Message], None]] = None,
     ) -> CyclicSendTaskABC:
-        return self.bus._send_periodic_internal(msgs, period, duration, modifier_callback)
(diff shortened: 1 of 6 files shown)


can/interfaces/ixxat/canlib.py
```diff
@@ -170,3 +171,7 @@ def state(self) -> BusState:
         Return the current state of the hardware
         """
         return self.bus.state
+
+    @staticmethod
+    def _detect_available_configs() -> List[AutoDetectedConfig]:
+        return vcinpl._detect_available_configs()
```
zariiii9003, line 177: Without any IXXAT-knowledge: does this detect FD-devices, too? What about `vcinpl2`?

MattWoodhead: That is a good question. I believe that `vcinpl.dll` and `vcinpl2.dll` are just the CAN LS/HS driver and CAN FD driver respectively, and that these are both components of the VCI4 package (so I think that both APIs should work for the CAN FD capable devices, but I am not 100% certain).

Based on the [implementation](https://github.com/hardbyte/python-can/blob/2582fe975c189471b0c439c1488909d48b219bdf/can/interfaces/ixxat/__init__.py#L20) of `get_ixxat_hwids()` in the `__init__.py`, I would expect that `vcinpl.dll` does work for the FD devices, and also has the advantage that it is compatible with the older VCI3 drivers (pre-FD) - I think this is why #1119 was chosen in favour of #1126. Unfortunately I dont have the FD capable hardware to test the theory.

If you would prefer, I can implement the function in the vcinpl2 file too and default to that implementation, falling back to to the vcinpl definition if `vcinpl2.dll` is not available?

zariiii9003: @fjburgos can you take a look?

zariiii9003: @MattWoodhead Regarding #1119 and #1126, with hindsight i think that was the wrong choice. The current solution with two IXXATBus implementations and that proxy class has lead to many bug reports. 

Isn't `vcinpl2.dll` backwards compatible to older devices and classical CAN 2.0? The docs say
image

MattWoodhead: The bug [canopen #296](https://github.com/christiansandberg/canopen/issues/296) in the canopen library apears to be related, and is one that is currently causing me a headache :)

Yes, I am 90% sure that the `vcinpl2.dll` driver is backwards compatible with all of the older ixxat devices, it just wont be present on any systems that have the VCI3 or earlier drivers installed. I can modify my local branch and test with a non-FD USB-to-CAN Compact V2 to confirm tomorrow.

MattWoodhead: Good morning @zariiii9003.
I have tested [this modified branch](https://github.com/MattWoodhead/python-can/tree/ixxat_unified_class2) with the non-FD ixxat interface onto a bus with another non-ixxat interface and it all worked as expected.
Tests:
* receive
* can.viewer
* send
* send periodic (using bus.bus.send_periodic workaround for proxy class)

zariiii9003: That sounds promising. Would you like to reimplement the IXXATBus based on vcinpl2 without a proxy class? Maybe we could build upon #1126

MattWoodhead: Hi @zariiii9003. Yes I am happy to give that a go. Do you want me to do it in this PR?

I have compared `canlib.py` from #1119 and `canlib_vcinpl2.py` #1126 from the time when 1119 was merged, and the only significant differences between them are:
1. The way in which the predefined bitrates are implemented (1126 stores them as class variables, 1119 stores them in the constants.py file.
2. 1119 implements [a check](https://github.com/hardbyte/python-can/blob/b886203701397e12e6d43a3d4b4ee32f07817eb6/can/interfaces/ixxat/canlib_vcinpl2.py#L676) of the hardware interfaces capabilities before adding features such as CAN FD to the open method of the VCI driver, whereas 1126 just adds them if the parameter provided to the class demands it. In my opinion the check of the interfaces available functionality is valuable if we are converging on a single implementation.

What are your thoughts?

zariiii9003: I'd suggest to create a new PR. If doesn't go anywhere, we can still merge this.

1. I'd prefer the constants file. Regarding the bitrates, it would be nice if support for the BitTiming classes was added, too.
2. I agree. Checking the capabilities doesn't hurt.

prunith: Just for information Ixxat PCI_CAN interface is not compatible with VCI4. I know this is an old devicebut please dont forget user that use legacy board :)

PR #1618, opened 2023-06-25
Relax BitTiming & BitTimingFd Validation

zariiii9003:
See #1616


Yuvalsh178 (approved):

lumagi:
I don't have access to a device with a decent screen size at the moment, so I don't think I can support with a review. But do we want the strict mode only to be adjustable in `__init__` and not in the class methods? This is only me asking without totally knowing the context.

lumagi:
That makes sense, thanks for clarifying!

Issue #1624, opened 2023-06-29
tests fail on bigendian


umlaeute:
### Describe the bug

Running the test-suite on a BingEndian architecture (like `s390x`), fails in two places:
- test_bit_timing.py::test_btr_persistence FAILED                                                                                                                                                                                      
- test_pcan.py::TestPCANBus::test_detect_available_configs FAILED

```
90s =================================== FAILURES ===================================
 90s _____________________________ test_btr_persistence _____________________________
 90s 
 90s     def test_btr_persistence():
 90s         f_clock = 8_000_000
 90s         for btr0btr1 in PCAN_BITRATES.values():
 90s             btr1, btr0 = struct.unpack("BB", btr0btr1)
 90s     
 90s >           t = can.BitTiming.from_registers(f_clock, btr0, btr1)
 90s 
 90s test_bit_timing.py:180: 
 90s _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 90s /usr/lib/python3/dist-packages/can/bit_timing.py:183: in from_registers
 90s     return cls(
 90s /usr/lib/python3/dist-packages/can/bit_timing.py:70: in __init__
 90s     self._validate()
 90s _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
 90s 
 90s self = can.BitTiming(f_clock=8000000, brp=21, tseg1=1, tseg2=1, sjw=1, nof_samples=1)
 90s 
 90s     def _validate(self) -> None:
 90s         if not 8 <= self.nbt <= 25:
 90s >           raise ValueError(f"nominal bit time (={self.nbt}) must be in [8...25].")
 90s E           ValueError: nominal bit time (=3) must be in [8...25].
 90s 
 90s /usr/lib/python3/dist-packages/can/bit_timing.py:74: ValueError
 90s __________________ TestPCANBus.test_detect_available_configs ___________________
 90s 
 90s self = <test.test_pcan.TestPCANBus testMethod=test_detect_available_configs>
 90s 
 90s     def test_detect_available_configs(self) -> None:
 90s         if platform.system() == "Darwin":
 90s             self.mock_pcan.GetValue = Mock(
[...]

PR #1625, opened 2023-06-29
BigEndian test fixes

umlaeute:
- PCAN_BITRATES are stored in BIGendian, no use to byteswap on littleendian.
- Also the TPCANChannelInformation expects data in the native byte order

Closes: https://github.com/hardbyte/python-can/issues/1624


zariiii9003 (approved):
Looks good, thanks!

Issue #1605, opened 2023-05-26
Ixxat USB-to-CAN v2 | python-can: Close bus, ACK and send_periodic errors


AntoineBricard:
Hi,

I'm new with python-can (package version= 4.2.1, running on python3.9.6, under Windows 10) and I wanted to perform a test to send/read CAN message using USB-to-CAN v2 (and SimplyCAN as target).
But I faced few issues using python-can:

**1. Close bus error:**

```
import can
bus = can.interface.Bus(interface='ixxat', channel=0, bitrate=250000, extended=True)
heartbeat = can.Message(arbitration_id=0x11487020, is_extended_id=True, data=[0x01])
bus.send(heartbeat)
bus.shutdown()
```

Above is the code I used to send CAN message. The point is once the code has been performed, it seems to not close properly the CAN bus. 
Bellow the error I got:
```
IXXATBus was not properly shut down
Exception ignored in: <function BusABC.__del__ at 0x000001ACF1418CA0>
```

Is there a way to properly close the CAN bus ? As i can rerun this script even after the error, should I just not taking account this error ?

**2. send_periodic :**
I wanted to send CAN message every seconds using the method "send_periodic" but it raise the bellow exception:
`AttributeError: 'IXXATBus' object has no attribute '_periodic_tasks'`

As in the doc, it's written that for Ixxat devices, the method "send_periodic" is supported, what did I misunderstand for the use of send_periodic method using Ixxat devices ? Bellow the code for the periodic message:
```
import can
bus = can.interface.Bus(interface='ixxat', channel=1, bitrate=250000, extended=True)
heartbeat = can.Message(arbitration_id=0x11487020, is_extended_id=True, data=[0x01])
bus.send_periodic(heartbeat, period=1)
bus.shutdown()
```

**3. NACK error :**

If the target has a problem and couldn't ACK the CAN message, the python-can API print errors logs as bellow:
```
CAN acknowledgment error
CAN message flags bAddFlags/bFlags2 0x00 bflags 0x05
```

And even once the target is reconnected this error message still appears. Is there a way to avoid them ?

Thanks in advance for any help,

Best regards,
Antoine

zariiii9003:
Hi, thanks for reporting
1. Try #1606 
2. The traceback might be helpful
3. 🤷‍♀️

AntoineBricard:
Hi zariiii9003,

1. Thanks for the link, using it could avoid the incorrect shutdown issue.

2. Here is the log I got:
```
[...]

Issue #1285, opened 2022-03-16
Unable to send periodic messages with IXXAT interface as of 4.0.0.


thshan:
### Description:
While attempting to set up a periodic task to synchronously write/read PDOs using an IXXAT USB-to-CAN and  the [canopen](https://github.com/christiansandberg/canopen) package, an AttributeError is generated due to the `IXXATBus` object missing `_periodic_tasks` (see: traceback below).

### Expected behavior
On v3.3.4, it is possible to use `node.rpdo[pdo].start` or `network.sync.start` from the `canopen` package (on its latest 2.0.0 release) without any errors being generated. Calling the constructor does not resolve the issue on V4.0.0, however, manually setting `bus._periodic_tasks = []` functions as a workaround. 

### Additional context

OS and version: Windows 10 Professional (10.0.19044 Build 19044)
Python version: Python 3.9.10, also confirmed with Python 3.10.3
python-can version: 4.0.0
python-can interface/s (if applicable): IXXAT. 

<details><summary>Traceback and logs</summary>

```

Traceback (most recent call last):
 line 78, in <module>
    main()
  line 54, in main
    network.start_nodes(net, 0.5)
  start_nodes
    node.rpdo[pdo].start(sync_time)
 line 449, in start
    self._task = self.pdo_node.network.send_periodic(
  "...\canopen\network.py", line 220, in send_periodic
    return PeriodicMessageTask(can_id, data, period, self.bus, remote)
 "...\canopen\network.py", line 294, in __init__
    self._start()
 "...\canopen\network.py", line 297, in _start
    self._task = self.bus.send_periodic(self.msg, self.period)
  "...\python_can\can\bus.py", line 239, in send_periodic
    periodic_tasks = self._periodic_tasks
AttributeError: 'IXXATBus' object has no attribute '_periodic_tasks'
```

Bus attributes on 4.0.0:
[...]

PR #1611, opened 2023-06-05
Ixxat bus refactor to use VCI4 driver only

MattWoodhead:
Re-implement ixxat bus in a single class using only the VCI4 driver as [discussed here](https://github.com/hardbyte/python-can/pull/1607#discussion_r1213660540)

- [x] initial refactor
- [ ] hardware testing
- [ ] implement support for BitTiming classess (WIP)
- [ ] implement tests

PR #1628, opened 2023-07-04
Fix IXXAT periodic messages not working

ChargedByte:
Hey,

As reported in a few instances, the periodic messages don't work on IXXAT. The `IXXATBus` appears to be just missing the call to the super-classes constructor. This call is present in the `develop` branch but is missing in the current release v4.2.2

References: #1285 #1605

### DIFF
--- a/can/interfaces/ixxat/canlib.py
+++ b/can/interfaces/ixxat/canlib.py
@@ -131,6 +131,8 @@ class IXXATBus(BusABC):
                 **kwargs
             )
 
+        super().__init__(channel=channel, **kwargs)
+
     def flush_tx_buffer(self):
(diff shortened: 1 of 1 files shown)


zariiii9003:
This change makes IXXATbus use thread based periodic tasks instead the hardware based implementations. I think it better than nothing, but it's not really fixing the issue.

I hope #1611 will fix all issues.

ChargedByte:
Ah, that makes sense. Looks like I didn't do enough research into the library, sorry about that. The workaround works for us and we can keep applying it locally. Probably best to close this PR and just wait for the refactoring?

zariiii9003:
I'll merge this. Thread-based periodic tasks are better than exceptions.

Issue #1531, opened 2023-03-03
ASCReader does not support ASC format since V8.1


Tian-Jionglu:
### Describe the bug

ASCReader in python-can 4.1.0 doesn't support ASC CANFD format since V8.1, since \<symbolic_name\> name is missing in code.
According to the document from Vector <can_log_trigger_asc_format.pdf>, there should be \<SYMBOLIC_NAME\> between \<ID\> and \<BRS\>.

### To Reproduce

Use python-can to read CANFD message in asc log since format V8.1.

### Expected behavior

Reading process stopped with CANFD message.

### Additional context
I made a quick fix in a forked [branch](https://github.com/Tian-Jionglu/python-can/tree/asc_quick_fix).
I think an overall fix about ASC format IO is needed in this issue.

OS and version:
Python version:
python-can version:
python-can interface/s (if applicable):

<details><summary>Traceback and logs</summary>

</details>

zariiii9003:
Since you have the format description, could you provide a backwards compatible solution as a PR?

Tian-Jionglu:
see PR #1641 , cause of  my misunderstanding of ASC logging.

PR #1641, opened 2023-08-03
Apply asc format since format v8.1

Tian-Jionglu:
A solution to solve issue #1531.

Mainly include 2 changes in `ASCReader` class:
1. extract header when initial ASCReader, to get correct `start_time`, `version`, `date`, etc before iterate the messages. I think this is necessary;
2. new function `_process_fd_can_frame_2` to process CANFD message in ASCFormat since V8.1;

Remaining issue:
some tests can not pass, mainly because after the change, correct `absolute` timestamp is get when iterate the messages;


zariiii9003:
@Tian-Jionglu Are you still working on the tests? I cannot merge if the tests are failing.

Tian-Jionglu:
Commits to parse the ASC format version using `packaging`.
Jobs in working: change ASC format test to adjust absolute time.

Tian-Jionglu:
@hardbyte @zariiii9003 ,
In example data, 
test/data/logfile.asc
test/data/logfile_errorframes.asc
test/data/test_CanErrorFrames.asc
test/data/test_CanFdMessage.asc
test/data/test_CanFdMessage64.asc
test/data/test_CanMessage.asc
test/data/test_CanRemoteMessage.asc

the ASC format version `// version 9.0.0` is actually logged, or only written for example?

The format in these files may be before `8.1.0`.

hardbyte (approved):

Tian-Jionglu:
When diving into test example, i found my mistake at the start.
The symbolic name issue is not only related to ASC format version. In fact, SymblicName works for symbolic logging at first place.
I'm still looking for example about symblic logging.

I would close this PR in coming days, and start a new PR to inherit the change to fix absolute time in ASC format.
If there is already support about symbolic logging in python-can, please give me a hint.

can/io/asc.py
```diff
@@ -98,14 +127,26 @@ def _extract_header(self) -> None:
...
             if comment_match:
                 continue
 
             if events_match:
                 self.internal_events_logged = events_match.group("no_events") is None
-                break
+                continue
 
-            break
+            if trigger_match:
```
zariiii9003, line 142: I think the files can contain multiple triggers. Parsing it as part of the header is probably not sufficient.

Tian-Jionglu:
see PR #1788

PR #1645, opened 2023-08-17
Do not stop notifier if exception was handled

### LINKED ISSUE #1643: ixxat: can.Notifier/Listener stops on too many errors
### Describe the bug
I am currently porting my application from python-can 3.3.4 to 4.x and noticed that the IXXAT implementation changed quite a lot. The new version raises much more exceptions than before, which breaks my usecase and I am unsure if there is a non-hacky way to get back the old behaviour.

The issue happens when registering a can.Listener/Notifier as a background thread and the IXXAT reports errors (e.g. because another controller on the bus sends something in a different baudrate). In that case an exception is raised in the receiver thread and it stops, which is permanent even if the bus errors recover.
The old implementation would just continue to call recv() till it succeeds and that's the behaviour I rely on in my code. Would it be possible to add e.g. a config flag to ignore such non-fatal errors?

### To Reproduce

Nothing special, just a can.Listener/Notifier and another CAN controller with a different baudrate is sending data on the bus.

### Expected behavior

Either non-fatal errors should be:
- ignored
- Communicated via some side-channel (e.g. some get_controller_warning_flags() function)
- Configurable to be disabled

### Additional context

OS and version: Windows 10
Python version: 3.9.x
python-can version: 4.2.2
python-can interface/s (if applicable): IXXAT

zariiii9003:
Closes #1643


prunith:
Thanks @zariiii9003 
i approuve your fix , now  notifier will not stop on error and lets decide listener the behavior of exception.
regards

Thierry

PR #1655, opened 2023-09-04
Place speedstudio in right, alphabetical order

Redysz:
Hi, I've just seen that list of available/handled interfaces here is not [fully](https://python-can.readthedocs.io/en/stable/configuration.html#interface-names) in order (speedstudio is not) so created this simple PR to fix that.


Redysz:
Sorry, I've read "speedstudio instead of seeedstudio" 🙃

PR #1660, opened 2023-09-10
* For interface Kvaser, add parameter exclusive

luojiaaoo:
* For interface Kvaser, add parameter exclusive
Don't allow sharing of this CANlib channel.

*  For interface Kvaser, add parameter override_exclusive
Open the channel even if it is opened for exclusive access already.


luojiaaoo:
I am sorry, I made a simple mistake, I have fixed in [8e75cc0]

zariiii9003 (approved):
Looks good to me

PR #1666, opened 2023-09-27
We do not need to account for drift when we USE_WINDOWS_EVENTS

pierreluctg:
When using USE_WINDOWS_EVENTS the waitable timer will be fired at fixed rate, we do not need to account for the drift.
The change makes it wait for the timer to be in signaled state only.
Previous code was causing multiple frames to be send back to back without delay.

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -324,7 +324,7 @@ class ThreadBasedCyclicSendTask(
             if USE_WINDOWS_EVENTS:
                 win32event.WaitForSingleObject(
                     self.event.handle,
-                    -1,
+                    win32event.INFINITE,
                 )
             else:
                 # Compensate for the time it takes to send the message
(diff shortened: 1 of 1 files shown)


pierreluctg:
@zariiii9003 please review

can/broadcastmanager.py
```diff
@@ -290,11 +289,12 @@ def start(self) -> None:
 
     def _run(self) -> None:
         msg_index = 0
-        msg_due_time_ns = time.perf_counter_ns()
 
         if USE_WINDOWS_EVENTS:
             # Make sure the timer is non-signaled before entering the loop
             win32event.WaitForSingleObject(self.event.handle, 0)
+        else:
+            msg_due_time_ns = time.perf_counter_ns()
```
zariiii9003, line 297: Could you move this out of the else clause? Otherwise the IDE complains, that `msg_due_time_ns` might be referenced before assignment. I know that can't actually happen..

zariiii9003 (approved):

PR #1671, opened 2023-10-05
add the possibility to return all the possible solutions using the from_sample_point static methods

danielhrisca:
The CAN databases can contain restrictions for the timings:
* minimum and maximum sampling times
* minimum and maximum SJW values
* minimum and maximum number of time quanta

With this change the user can get all the possible solutions for a given sampling value and can further filter the solutions using the known constraints.


zariiii9003:
Both me and mypy do not like the different return types of the `from_sample_point` method. 😄 

What do you think about adding another method which generates BitTiming instances:
```
    @classmethod
    def iterate_from_sample_point(
        cls,
        f_clock: int,
        bitrate: int,
        sample_point: float = 69.0,
    ) -> Iterator["BitTiming"]:
        ...
        yield bt
        ...
```

The `from_sample_point` method could use the generator to create the `possible_solutions` list and return the "best"

danielhrisca:
@zariiii9003 thanks for the pointers, let me know if the new code is ok

zariiii9003:
Looks good, could you add some trivial test?

zariiii9003 (approved):
Looks good, thank you 👍

PR #1679, opened 2023-10-13
Allow RestartableCyclicTaskABC using Windows events to restart

pierreluctg:
Allow RestartableCyclicTaskABC using Windows events to restart

Also adding unit test to cover RestartableCyclicTaskABC and avoid this type of bug.

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -269,9 +269,10 @@ class ThreadBasedCyclicSendTask(
         self.start()
 
     def stop(self) -> None:
-        if USE_WINDOWS_EVENTS:
-            win32event.CancelWaitableTimer(self.event.handle)
         self.stopped = True
+        if USE_WINDOWS_EVENTS:
+            # Reset and signal any pending wait by setting the timer to 0
+            win32event.SetWaitableTimer(self.event.handle, 0, 0, None, None, False)
 
     def start(self) -> None:
         self.stopped = False
--- a/test/simplecyclic_test.py
+++ b/test/simplecyclic_test.py
@@ -152,6 +152,41 @@ class SimpleCyclicSendTaskTest(unittest.TestCase, ComparingMessagesTestCase):
 
         bus.shutdown()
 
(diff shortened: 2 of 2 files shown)


pierreluctg:
@zariiii9003 please review and consider for a bug fix release.

can/broadcastmanager.py
```diff
@@ -275,16 +275,14 @@ def stop(self) -> None:
...
+        if USE_WINDOWS_EVENTS:
+            win32event.SetWaitableTimer(
+                self.event.handle, 0, self.period_ms, None, None, False
+            )
         if self.thread is None or not self.thread.is_alive():
             name = f"Cyclic send task for 0x{self.messages[0].arbitration_id:X}"
             self.thread = threading.Thread(target=self._run, name=name)
             self.thread.daemon = True
-
-            if USE_WINDOWS_EVENTS:
```
zariiii9003, line 283: Wasn't `SetWaitableTimer` called anyway? `self.thread.is_alive()` should return False on a restart, right?

zariiii9003:
Semi-related: Could you check why the `SimpleCyclicSendTaskTest` test is so slow after #1666?

pierreluctg:
> `SimpleCyclicSendTaskTest`

Your a right, will fix it

pierreluctg:
@zariiii9003 please have another look. This should address the issues.

> The `CancelWaitableTimer` function does not change the signaled state of the timer. 

So the thread was waiting for ever (`win32event.INFINITE`). This change replace  `CancelWaitableTimer` by `SetWaitableTimer` with a period of `0`.

This is documented in the `SetWaitableTimer` API doc.
> A periodic timer automatically reactivates each time the period elapses, until the timer is canceled using the [CancelWaitableTimer](https://learn.microsoft.com/en-us/windows/desktop/api/synchapi/nf-synchapi-cancelwaitabletimer) function or reset using SetWaitableTimer.

zariiii9003 (approved):
Looks good. The reset with `SetWaitableTimer` is a nice solution.

PR #1687, opened 2023-10-29
Add feature to detect socketcand beacon

faisal-shah:
This implements the client side service discovery for socketcand:
https://github.com/dschanoeh/socketcand/blob/master/doc/protocol.md#service-discovery

Need some feedback on how to enable this option. In the PR, I've proposed a way that makes the happy path as easy as possible ... though it is a breaking change for the SocketCanDaemonBus class.

Happy path:
```
bus = SocketCanDaemonBus(None) # Auto detect socketcand server, and connect to first available CAN bus
```

1. If a host address (hostname,port) is provided, it works the same way as before.

2. If a host address (hostname,port) is *not* provided, it activates detect_beacon(). The user can optionally provide an address to listen for the beacon on a specific interface and port. If not provided, it listens on all interfaces at the default port 42000. If channel is None, it selects the first channel in the channel list provided in the beacon.

### DIFF
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -26,7 +26,7 @@ DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS = ""
 DEFAULT_SOCKETCAND_DISCOVERY_PORT = 42000
 
 
-def detect_beacon():
+def detect_beacon(timeout_ms):
     with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
         sock.bind(
             (DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS, DEFAULT_SOCKETCAND_DISCOVERY_PORT)
@@ -37,8 +37,6 @@ def detect_beacon():
             DEFAULT_SOCKETCAND_DISCOVERY_PORT,
         )
 
-        # Time between beacons no more than 3 seconds. Allow for at least 3
-        timeout_ms = 12000
         now = time.time() * 1000
         end_time = now + timeout_ms
         while (time.time() * 1000) < end_time:
@@ -328,7 +326,9 @@ class SocketCanDaemonBus(can.BusABC):
     @staticmethod
     def _detect_available_configs() -> List[can.typechecking.AutoDetectedConfig]:
         try:
-            return detect_beacon()
+            # Time between beacons no more than 3 seconds. Allow for a little
+            # more than 3s
+            return detect_beacon(3.1)
         except Exception as e:
             log.warning(f"Could not detect socketcand beacon: {e}")
             return []
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -321,4 +321,8 @@ class SocketCanDaemonBus(can.BusABC):
 
     @staticmethod
     def _detect_available_configs() -> List[can.typechecking.AutoDetectedConfig]:
-        return detect_beacon()
+        try:
+            return detect_beacon()
+        except Exception as e:
+            log.warning(f"Could not detect socketcand beacon: {e}")
+            return []
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -30,9 +30,9 @@ def detect_beacon(timeout_ms: int = 3100) -> List[can.typechecking.AutoDetectedC
     """
     Detects socketcand servers
 
-    This is what :meth:`can.detect_available_configs` ends up calling
-    search for available socketcand servers with fixed timeout of 3.1
-    seconds (socketcand sends out a beacon packet every 3 seconds).
+    This is what :meth:`can.detect_available_configs` ends up calling to search
+    for available socketcand servers with a default timeout of 3100ms
+    (socketcand sends a beacon packet every 3000ms).
 
     Using this method directly allows for adjusting the timeout. Extending
     the timeout beyond the default time period could be useful if UDP
@@ -341,8 +341,6 @@ class SocketCanDaemonBus(can.BusABC):
     @staticmethod
     def _detect_available_configs() -> List[can.typechecking.AutoDetectedConfig]:
         try:
-            # Time between beacons no more than 3 seconds. Allow for a little
-            # more than 3s
             return detect_beacon()
         except Exception as e:
             log.warning(f"Could not detect socketcand beacon: {e}")
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -30,19 +30,19 @@ def detect_beacon(timeout_ms: int = 3100) -> List[can.typechecking.AutoDetectedC
     """
     Detects socketcand servers
 
-        This is what :meth:`can.detect_available_configs` ends up calling
-        search for available socketcand servers with fixed timeout of 3.1
-        seconds (socketcand sends out a beacon packet every 3 seconds).
+    This is what :meth:`can.detect_available_configs` ends up calling
+    search for available socketcand servers with fixed timeout of 3.1
+    seconds (socketcand sends out a beacon packet every 3 seconds).
 
-        Using this method directly allows for adjusting the timeout. Extending
-        the timeout beyond the default time period could be useful if UDP
-        packet loss is a concern.
+    Using this method directly allows for adjusting the timeout. Extending
+    the timeout beyond the default time period could be useful if UDP
+    packet loss is a concern.
 
-        :param timeout_ms:
-            Timeout in milliseconds to wait for socketcand beacon packets
+    :param timeout_ms:
+        Timeout in milliseconds to wait for socketcand beacon packets
 
-        :return:
-            See :meth:`~can.detect_available_configs`
+    :return:
+        See :meth:`~can.detect_available_configs`
     """
     with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
         sock.bind(
--- a/can/interfaces/socketcand/socketcand.py
+++ b/can/interfaces/socketcand/socketcand.py
@@ -27,83 +27,87 @@ DEFAULT_SOCKETCAND_DISCOVERY_PORT = 42000
 
 
 def detect_beacon():
-    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-    sock.bind((DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS, DEFAULT_SOCKETCAND_DISCOVERY_PORT))
-    log.info(
-        "Listening on for socketcand UDP advertisement on %s:%s",
-        DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS,
-        DEFAULT_SOCKETCAND_DISCOVERY_PORT,
-    )
+    with socket.socket(socket.AF_INET, socket.SOCK_DGRAM) as sock:
+        sock.bind(
+            (DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS, DEFAULT_SOCKETCAND_DISCOVERY_PORT)
+        )
+        log.info(
+            "Listening on for socketcand UDP advertisement on %s:%s",
+            DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS,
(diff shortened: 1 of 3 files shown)


zariiii9003:
Regarding the API, i'd suggest to leave the actual `SocketCanDaemonBus` class untouched.
The `detect_beacon` function could work similar to [can.detect_available_configs()](https://python-can.readthedocs.io/en/develop/utils.html#can.detect_available_configs) and return a list of dictionaries like this:

```python3
cfg_list: list[AutoDetectedConfig] = detect_beacon()
bus = SocketCanDaemonBus(**cfg_list[0])
```
A dictionary should contain all necessary keywords to instantiate the `SocketCanDaemonBus`.

faisal-shah:
I'll be testing this in a couple different environments, and will report back.

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,96 @@
...
 import can
 
 log = logging.getLogger(__name__)
 
+DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS = ""
+DEFAULT_SOCKETCAND_DISCOVERY_PORT = 42000
+
+
+def detect_beacon():
+    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
```
zariiii9003, line 30: Use a context manager to automatically close the socket.

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,96 @@
...
+    sock.bind((DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS, DEFAULT_SOCKETCAND_DISCOVERY_PORT))
+    log.info(
+        f"Listening on for socketcand UDP advertisement on {DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS}:{DEFAULT_SOCKETCAND_DISCOVERY_PORT}"
```
zariiii9003, line 33:
```suggestion
        "Listening on for socketcand UDP advertisement on %s:%s",
        DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS,
        DEFAULT_SOCKETCAND_DISCOVERY_PORT
```

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,96 @@
...
+    )
+
+    # Time between beacons no more than 3 seconds. Allow for at least 3
+    timeout_ms = 12000
```
zariiii9003, line 37: This could be a function parameter.

can/interfaces/socketcand/socketcand.py
```diff
@@ -231,3 +318,11 @@ def shutdown(self):
         """Stops all active periodic tasks and closes the socket."""
         super().shutdown()
         self.__socket.close()
+
+    @staticmethod
+    def _detect_available_configs() -> List[can.typechecking.AutoDetectedConfig]:
+        try:
+            return detect_beacon()
```
zariiii9003, line 325: I'm not sure, whether it's a good idea to delay the `can.detect_available_configs()` by 12 seconds for everyone.
Is it possible to reduce that duration?

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,96 @@
...
+
+
+def detect_beacon():
+    sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    sock.bind((DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS, DEFAULT_SOCKETCAND_DISCOVERY_PORT))
+    log.info(
+        f"Listening on for socketcand UDP advertisement on {DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS}:{DEFAULT_SOCKETCAND_DISCOVERY_PORT}"
+    )
+
+    # Time between beacons no more than 3 seconds. Allow for at least 3
```
zariiii9003, line 36: One timeout should be sufficient for all beacons, why do you multiply the timeout?

can/interfaces/socketcand/socketcand.py
faisal-shah: Should this be defined as a global, and use it has a function parameter to detect_beacon()? Or a parameter to _detect_available_configs()? If the later - we would need to add a kwargs to the BusABC::_detect_available_configs(), correct?

can/interfaces/socketcand/socketcand.py
faisal-shah: Beacons are send via UDP broadcast and are not guaranteed delivery. The time between each beacon broadcast is 3 seconds. So - 12 sec would be required to guarantee a receiving window of at least 3 beacons.

can/interfaces/socketcand/socketcand.py
faisal-shah: Done in 8e4a6db

can/interfaces/socketcand/socketcand.py
faisal-shah: Ok, I understand what you're saying. I wasn't aware calling can.detect_available_configs() will call all the detect_available_configs of all the interfaces. hehe. Yeah - we could reduce it. Right now the time window is designed to be wide enough for 4 beacon broadcasts. If we reduce that to 2 beacon broadcasts, it would be 6 seconds. How about 6?

can/interfaces/socketcand/socketcand.py
zariiii9003: Aren't the beacons indepedent of each other? I thought the max. time between broadcasts of each individual beacon is 3s.

faisal-shah: You are correct. I think there is a misunderstanding. When I mentioned beacon, I meant beacon-packet. Assume there is a single socketcand server on the network. I proposed a timeout that would guarantee a receiving window of at least 2 beacon packets. I didn't want to set a timeout on which only a single beacon packet could be received - as UDP does not guarantee delivery.

(The above would be true if there were multiple socketcand servers as well, as they send their beacon packets independently, with 3s between beacon packet broadcasts).

can/interfaces/socketcand/socketcand.py
zariiii9003: The timeout should be a function parameter for `detect_beacon` in seconds. `_detect_available_configs` can use a short timeout like 3.1s, while users can wait for longer if they call this function directly.

can/interfaces/socketcand/socketcand.py
zariiii9003: But packet loss is probably an exception, i would not wait this long per default. A little over 3 seconds should be enough unless the the user chooses to wait longer.

faisal-shah: Agreed.

can/interfaces/socketcand/socketcand.py
faisal-shah: Ok. I made it 3.1s in _detect_available_configs, and added the parameter to detect_beacon.

faisal-shah:
All review comments addressed.

zariiii9003:
You should mention it in the docs, or nobody will know about it 😄 
https://python-can--1687.org.readthedocs.build/en/1687/interfaces/socketcand.html

faisal-shah:
> You should mention it in the docs, or nobody will know about it 😄 https://python-can--1687.org.readthedocs.build/en/1687/interfaces/socketcand.html

Done.

zariiii9003:
You need to add `.. autofunction:: can.interfaces.socketcand.detect_beacon` somewhere before you can reference it.

doc/interfaces/socketcand.rst
```diff
@@ -37,6 +37,33 @@ The output may look like this::
     Timestamp: 1637791111.609763    ID: 0000031d    X Rx                DLC:  8    16 27 d8 3d fe d8 31 24
     Timestamp: 1637791111.634630    ID: 00000587    X Rx                DLC:  8    4e 06 85 23 6f 81 2b 65
 
+
+This interface also supports :meth:`can.BusABC._detect_available_configs`.
```
zariiii9003, line 41: Refer to `` :meth:`~can.detect_available_configs` `` instead of `BusABC._detect_available_configs`.
Avoid using the private method here.

faisal-shah:
> You need to add `.. autofunction:: can.interfaces.socketcand.detect_beacon` somewhere before you can reference it.

Done. Docs are passing now.

faisal-shah:
@zariiii9003 I've noticed sometimes tests will spuriously fail. I made an empty commit to trigger the tests and all are passing now. Let me know if any more changes are required to merge this. Appreciate the pointers in working with the doc system.

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,115 @@
...
 
 import can
 
 log = logging.getLogger(__name__)
 
+DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS = ""
+DEFAULT_SOCKETCAND_DISCOVERY_PORT = 42000
+
+
+def detect_beacon(timeout_ms) -> List[can.typechecking.AutoDetectedConfig]:
```
zariiii9003, line 29:
```suggestion
def detect_beacon(timeout_ms: int = 3100) -> List[can.typechecking.AutoDetectedConfig]:
```

I would set the default timeout here.

can/interfaces/socketcand/socketcand.py
```diff
@@ -231,3 +337,13 @@ def shutdown(self):
         """Stops all active periodic tasks and closes the socket."""
         super().shutdown()
         self.__socket.close()
+
+    @staticmethod
+    def _detect_available_configs() -> List[can.typechecking.AutoDetectedConfig]:
+        try:
+            # Time between beacons no more than 3 seconds. Allow for a little
+            # more than 3s
+            return detect_beacon(3.1)
```
zariiii9003, line 346:
```suggestion
            return detect_beacon()
```

Remove the timeout here. This would be a bug anyway, since you expect milliseconds in `detect_beacon`

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,115 @@
...
 
+DEFAULT_SOCKETCAND_DISCOVERY_ADDRESS = ""
+DEFAULT_SOCKETCAND_DISCOVERY_PORT = 42000
+
+
+def detect_beacon(timeout_ms) -> List[can.typechecking.AutoDetectedConfig]:
+    """
+    Detects socketcand servers
+
+        This is what :meth:`can.detect_available_configs` ends up calling
```
zariiii9003, line 33: This indetention should be removed. The docstring is formatted as a quote because of it:
image

can/interfaces/socketcand/socketcand.py
```diff
@@ -13,12 +13,115 @@
...
+        search for available socketcand servers with fixed timeout of 3.1
```
zariiii9003, line 34:
```suggestion
        to search for available socketcand servers with fixed timeout of 3.1
```

zariiii9003 (commented):
There are still a few minor issues. I'll merge after you fix these. Thank you for your patience 👍

can/interfaces/socketcand/socketcand.py
grant-allan-ctct: Just my opinion, but the mix of milliseconds in the code and seconds in the commentary could cause headaches and bugs for people in the future (and in the present 😎).  

I would suggest seriously think about whether it might be a good plan to work through top-to-bottom in this PR to achieve consistency.

can/interfaces/socketcand/socketcand.py
faisal-shah: Done.

faisal-shah:
> There are still a few minor issues. I'll merge after you fix these. Thank you for your patience 👍

Hope I got them all. Thanks for your maintainership.

zariiii9003 (approved):
Thank you!

FedericoSpada:
Hi All,
I've just updated from version 4.3.1 to 4.4.2 and noticed that dongle autodetection went from lasting 500ms to 4,5s. After some tests, I've found that this change is the main reason for the big increase in duration. Isn't there a better way to implement this detection? Perhaps avoiding it altogether if some drivers aren't installed? Or maybe it could be disabled by default and you need to force the search for this particular interface so that old codes will still run fast without having to update the instruction with something like this:

image

Thanks in advance, greetings.

PR #1690, opened 2023-11-14
Implement _detect_available_configs for Seeedbus

RyanHope:
Implements a _detect_available_configs method for Seeedbus

### DIFF
--- a/can/interfaces/seeedstudio/seeedstudio.py
+++ b/can/interfaces/seeedstudio/seeedstudio.py
@@ -325,7 +325,5 @@ class SeeedBus(BusABC):
         configs = []
         for port in list_comports():
             if port.vid == 0x1A86 and port.pid == 0x7523:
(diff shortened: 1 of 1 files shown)


can/interfaces/seeedstudio/seeedstudio.py
```diff
@@ -310,3 +319,11 @@ def fileno(self):
...
             )
             return -1
+
+    @staticmethod
+    def _detect_available_configs() -> List[AutoDetectedConfig]:
+        configs = []
+        for port in list_comports():
+            if port.vid == 0x1A86 and port.pid == 0x7523:
+                configs.append({"interface": "seeedstudio", "channel": port.device})
+        return configs
```
zariiii9003, line 329:
```suggestion
        return configs  # type: ignore[return-value]
```

PR #1709, opened 2023-12-22
improve the "ASCReader"  read performance.

XXIN0:
Improve the `ASCReader`  read performance.
Use `re.compile` to compile the regular expression,then `match` the message will faster than direct `match` the message .

The read speed is about 30% faster than before.

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -9,7 +9,7 @@ import logging
 import re
 import time
 from datetime import datetime
-from typing import Any, Dict, Generator, List, Optional, TextIO, Union
+from typing import Any, Dict, Final, Generator, List, Optional, TextIO, Union
 
 from ..message import Message
 from ..typechecking import StringPathLike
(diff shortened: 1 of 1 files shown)


can/io/asc.py
```diff
@@ -254,15 +254,17 @@ def _process_fd_can_frame(self, line: str, msg_kwargs: Dict[str, Any]) -> Messag
 
     def __iter__(self) -> Generator[Message, None, None]:
         self._extract_header()
-
+        trigger_match_re = re.compile(
```
zariiii9003, line 257: You can move these outside of the class as module constants, so they are only compiled once.

```python3
ASC_TRIGGER_REGEX: Final = re.compile(
    ...
)
ASC_MESSAGE_REGEX: Final = re.compile(
    ...
)
```

can/io/asc.py
```diff
@@ -254,15 +254,17 @@ def _process_fd_can_frame(self, line: str, msg_kwargs: Dict[str, Any]) -> Messag
...
         for _line in self.file:
             line = _line.strip()
 
-            trigger_match = re.match(
-                r"begin\s+triggerblock\s+\w+\s+(?P<datetime_string>.+)",
-                line,
-                re.IGNORECASE,
-            )
+            trigger_match = trigger_match_re.match(line)
             if trigger_match:
```
zariiii9003, line 268: You can use the walrus operator `:=` here:

```python3
if trigger_match := trigger_match_re.match(line):
    ...
```

XXIN0:
this pull request is a little confusing, so I close it and make a new pull request.

PR #1713, opened 2023-12-26
Fix send periodic

### LINKED ISSUE #1710: send_periodic with non-None duration sends one extra message before stopping
### Describe the bug

If we pass a `duration` argument to `send_periodic`, the message transmissions don't stop at the time specified.  One extra message is sent.  

### To Reproduce

To see this, call `send_periodic` with `period` argument and `duration` arguments such that the `duration` is not simply a multiple of the `period`.  For example, with `duration` of 2 seconds, specify a `period` of 0.6 seconds, or 0.8 seconds.  In both cases, the final transmitted message will be at 2.4 seconds (which is 0.6 x 4 repeats, or 0.8 x 3 repeats), which is well past the expiry of the 2 seconds `duration`.

### Expected behavior

The messages should stop transmitting when the `duration` time elapses (give or take a little wiggle-room).

### Additional context
We seem to have effectively an off-by-one error that's arising from the way that the `ThreadBasedCyclicSendTask` has things ordered inside its while loop:
1. [send a message](https://github.com/hardbyte/python-can/blob/f0634d2445f74f8c5c3b90236975f73fe3d8ba2a/can/broadcastmanager.py#L304)
2. [check whether it's time to stop yet](https://github.com/hardbyte/python-can/blob/f0634d2445f74f8c5c3b90236975f73fe3d8ba2a/can/broadcastmanager.py#L322) 
3. [wait for another period](https://github.com/hardbyte/python-can/blob/f0634d2445f74f8c5c3b90236975f73fe3d8ba2a/can/broadcastmanager.py#L334)

I'm thinking that this means that the final message can be later than the `duration` by an amount of time up to the size of the `period`.

OS and version: Windows10
Python version: 3.11.6
python-can version: 4.3.0

<details><summary>Sample output where devices on a bus are asked periodically to claim addresses:</summary>

#### With `duration` 2 seconds, `period` 0.6 seconds

```
Timestamp:       15.414987    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       16.016019    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       16.615183    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       17.215543    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       17.816031    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
```
Time of final transmission after first transmission = 17.816031 - 15.414987 = 2.401044 seconds

#### With `duration` 2 seconds, `period` 0.8 seconds

```
Timestamp:       19.160761    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       19.961824    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       20.761057    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
Timestamp:       21.561177    ID: 18eafffe    X Tx                DL:  3    00 ee 00                    Channel: 1
```
Time of final transmission after first transmission = 21.561177 - 19.160761 = 2.400416 seconds

</details>

Tbruno25:
Closes #1710 

The proposed fix is to check whether `duration` is exceeded before the transmit of the next message.

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -297,11 +297,8 @@ class ThreadBasedCyclicSendTask(
             win32event.WaitForSingleObject(self.event.handle, 0)
 
         while not self.stopped:
-            if not USE_WINDOWS_EVENTS:
-                msg_due_time_ns += self.period_ns
             if self.end_time is not None and time.perf_counter() >= self.end_time:
                 break
-            msg_index = (msg_index + 1) % len(self.messages)
 
             # Prevent calling bus.send from multiple threads
             with self.send_lock:
@@ -321,6 +318,11 @@ class ThreadBasedCyclicSendTask(
                     if not self.on_error(exc):
                         self.stop()
                         break
+
+            if not USE_WINDOWS_EVENTS:
+                msg_due_time_ns += self.period_ns
+
+            msg_index = (msg_index + 1) % len(self.messages)
 
             if USE_WINDOWS_EVENTS:
                 win32event.WaitForSingleObject(
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -297,6 +297,12 @@ class ThreadBasedCyclicSendTask(
             win32event.WaitForSingleObject(self.event.handle, 0)
 
         while not self.stopped:
+            if not USE_WINDOWS_EVENTS:
+                msg_due_time_ns += self.period_ns
+            if self.end_time is not None and time.perf_counter() >= self.end_time:
+                break
+            msg_index = (msg_index + 1) % len(self.messages)
+
             # Prevent calling bus.send from multiple threads
             with self.send_lock:
                 try:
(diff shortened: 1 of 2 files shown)


grant-allan-ctct:
I'm delighted to see that you've added a new test case.  🏆

zariiii9003:
`msg_due_time_ns` and `msg_index` should only be incremented after the message was sent.

hardbyte (approved):
Thanks both @grant-allan-ctct and @Tbruno25 for the bug report, fix and test.

zariiii9003:
Now the [CI is failing](https://github.com/hardbyte/python-can/actions/runs/7397622800/job/20125084518) 😅
@Tbruno25 Could you try to improve the robustness of `test_send_periodic_duration`?

Tbruno25:
@zariiii9003 absolutely! Although it appears to only be failing on the very specific windows/pypy combinations. Would it be wrong if I simply skipped this test for this environment?

grant-allan-ctct:
It's a interesting thing.  I wonder what it's telling us.  

Perhaps it has found a bug for which there is no other test coverage yet existing.  For example, perhaps the message `timestamp` values are coming back as integers on this platform.  Or perhaps one the calls to `self.bus1.recv` is choking on one of the arguments that the testcase is passing it.

Disclaimer: I just had to Google what pypy is, so please disregard if you think this comment could well be nonsense.

zariiii9003:
> @zariiii9003 absolutely! Although it appears to only be failing on the very specific windows/pypy combinations. Would it be wrong if I simply skipped this test for this environment?

Yes that's okay. I think we disable most timing sensitive tests in CI.

PR #1727, opened 2024-01-17
add parser for TRC 1.3 format

the-working-rene:
added columns "bus" and reserverd. bus is the number of the CAN Bus as an integer, reserved is not parsed.

Reference: https://www.peak-system.com/produktcd/Pdf/English/PEAK_CAN_TRC_File_Format.pdf


the-working-rene:
todo: 
* [ ] has channel the same meaning as bus ?
* [ ]  if so: remove new bus attribute and parse bus as channel
* [ ]  if not: update docs accordingly

the-working-rene:
this "works for me" for an example log generated with PCAN Explorer 5, but I'd appreciate any comments for this implementation

zariiii9003:
Closing in favor of #1753

PR #1731, opened 2024-01-19
Use ctypes.util.find_library to find vxlapi dll

zariiii9003:


luojiaaoo:
old version vector dll may not support "xlCanReceive", only except FileNotFoundError sometion cause crash...

PR #1733, opened 2024-01-19
Added VariableRateCyclicTaskABC and updated ThreadBasedCyclicSendTask

sschueth:
* Modified `bus.py` to include an optional argument, `period_intra`, in `send_periodic` and `_send_periodic_internal`
* Implemented `VariableCyclicTaskRateABC` within `broadcastmanager.py` to be inherited by `ThreadBasedCyclicTask`
* Added attributes to `CyclicSendTaskABC` to help with handling a group of messages
* relaxed `arbitration_id` requirement within group of messages and updated the `_check_modified_messages` to check for each updataed message's `arbitration_id` to preserve that `arbitration_id` does not change when a message is modified
* Modified `ThreadBasedCyclicTask._run()` for variable rate transmission of messages in a group

### DIFF
--- a/can/broadcastmanager.py
+++ b/can/broadcastmanager.py
@@ -67,12 +67,13 @@ class CyclicSendTaskABC(CyclicTask, abc.ABC):
         :raises ValueError: If the given messages are invalid
         """
         messages = self._check_and_convert_messages(messages)
-
-        # Take the Arbitration ID of the first element
-        self.arbitration_id = messages[0].arbitration_id
+        self.msgs_len = len(messages)
+        # Take the Arbitration ID of each message and put them into a list
+        self.arbitration_id = [self.messages[idx].arbitration_id for idx in range(self.msgs_len)]
         self.period = period
         self.period_ns = int(round(period * 1e9))
         self.messages = messages
+        self.msg_index = 0
 
     @staticmethod
     def _check_and_convert_messages(
@@ -81,8 +82,7 @@ class CyclicSendTaskABC(CyclicTask, abc.ABC):
         """Helper function to convert a Message or Sequence of messages into a
         tuple, and raises an error when the given value is invalid.
 
-        Performs error checking to ensure that all Messages have the same
-        arbitration ID and channel.
+        Performs error checking to ensure that all Messages have the same channel.
 
         Should be called when the cyclic task is initialized.
 
@@ -96,12 +96,6 @@ class CyclicSendTaskABC(CyclicTask, abc.ABC):
         if not messages:
             raise ValueError("Must be at least a list or tuple of length 1")
         messages = tuple(messages)
-
-        all_same_id = all(
-            message.arbitration_id == messages[0].arbitration_id for message in messages
-        )
-        if not all_same_id:
-            raise ValueError("All Arbitration IDs should be the same")
 
         all_same_channel = all(
             message.channel == messages[0].channel for message in messages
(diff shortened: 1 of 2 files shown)


sschueth:
@tysonite @hardbyte @zariiii9003 

I referenced PR #785 related to the relaxation of the uniqueness of the `arbitration_id` in SocketCAN and was looking for some insight on the reason why the `arbitration_id` uniqueness checking exists for a Sequence of messages within `ModifiableCyclicTaskABC`.

* Is this uniqueness restriction related to a hardware / driver limitation?
  * I've only been able to test this on my available hardware (vector and kvaser) and have not had any issues with relaxing the uniqueness of `arbitration_id`
* Is this uniqueness restriction to ensure a modified or updated message is not changing an existing message `arbitration_id` and therefore removing it from the bus?
  * If this is the case, I think my changes within `_check_modified_messages()` should suffice?

I removed the `all_same_id` check within `_check_and_convert_messages()` and converted `self.arbitration_id` to a list of `arbitration_id`. To preserve the same behavior of "The arbitration ID of new cyclic messages cannot be changed from when the task was created" I added a series of checks within `_check_modified_messages()` to check that the new sequence of messages have respecitvely matching `arbitration_id` of the original sequence of messages.

For context, I'm making this change to allow for messages with different arbitration ids to be passed in the same task ensuring the order of transmission and improved timing. For example, with this update the J1939-76 Safety Header Message and Safety Data Message can be explicitly linked and transmitted as a sequence of messages within the same task.

Any feedback / guidance is greatly appreciated. Thanks!

can/broadcastmanager.py
```diff
@@ -67,12 +67,13 @@ def __init__(
         :raises ValueError: If the given messages are invalid
         """
         messages = self._check_and_convert_messages(messages)
-
-        # Take the Arbitration ID of the first element
-        self.arbitration_id = messages[0].arbitration_id
+        self.msgs_len = len(messages)
+        self.messages = messages
+        # Take the Arbitration ID of each message and put them into a list
+        self.arbitration_id = [self.messages[idx].arbitration_id for idx in range(self.msgs_len)]
```
hardbyte, line 73: prefer `self.arbitration_ids`

can/broadcastmanager.py
```diff
@@ -154,16 +148,17 @@ def _check_modified_messages(self, messages: Tuple[Message, ...]) -> None:
...
-            raise ValueError(
-                "The arbitration ID of new cyclic messages cannot be changed "
-                "from when the task was created"
-            )
+        for idx in range(self.msgs_len):
+            if self.arbitration_id[idx] != messages[idx].arbitration_id:
+                raise ValueError(
+                    "The arbitration ID of new cyclic messages cannot be changed "
+                    "from when the task was created"
+                )
```
hardbyte, line 161: I haven't tried but do you know if socketcan supports reordering the messages as well? In which case prefer to count the arbitration ids match.

can/bus.py
```diff
@@ -235,6 +236,10 @@ def send_periodic(
             Function which should be used to modify each message's data before
             sending. The callback modifies the :attr:`~can.Message.data` of the
             message and returns ``None``.
+        :param period_intra:
+            Period in seconds between each message when sending multiple messages
+            in a sequence. If not provided, the period will be used for each
+            message.
```
hardbyte, line 242: Period in seconds between individual message within the sequence. 

I assume no impact for sending individual messages? Could you also update the docstring for `period` 🙏🏼

hardbyte (commented):
I like the direction. 

Would appreciate this being tested and the user facing docs being updated too.

DjMike238:
Hi, will this PR ever be merged? I'm trying to use this library to send multipacket J1939 packets (in which the first packet will have a different arbitration ID than the others) periodically, and having the relaxed `arbitration_id` requirement for this would be very useful for my use case.

I'm currently using a PCAN interface, I just tried commenting out the code that checks `all_same_id` in `_check_and_convert_messages()` from `broadcastmanager.py` and it's working without any issues.

DjMike238:
Hi, just wanted to say that I've been testing this PR for a while and it properly allows sending multipacket CAN messages periodically (which I'm supposing is the reason behind this PR) both on PCAN (tested on PCAN-USB X6) and Vector (tested on VN1630A).

Is there any possibility to have this merged in a future release? I could also try and take care of any issues, if this PR is not worked on anymore (is that fine with you, @sschueth?).

sschueth:
Hi, @DjMike238. Apologies this PR fell behind. Thank you for testing with PCAN. I have successfully tested with Vector and KVaser. I think my biggest hang up going forward is SocketCAN. I will try to make some time this week to clean up this branch.

Issue #1699, opened 2023-11-29
The neovi interface `use_system_timestamp` feature has compatibility issues


grant-allan-ctct:
### Describe the bug

I am a beginner with python-can, and am up to learning about the asyncio feature.  After I adapted the [asyncio sample code](https://python-can.readthedocs.io/en/stable/asyncio.html) to use the `neovi` interface, I encountered exceptions being thrown in the code which was logging an `asc` file.  The code in `asc.py` tries to perform a `split` on the decimal point in [this line](https://github.com/hardbyte/python-can/blob/2e58a21578cf4451538e4e1f45cddbd85d12b488/can/io/asc.py#L406), and this was causing an exception to be thrown due to my timestamps being an integral number of milliseconds.  I realized that this was related to the `use_system_timestamp` feature that is specific to the `neovi` interface.  It turns out that if this Boolean is set to `True` when creating the `Bus` instance, timestamps from the Bus are emitted as `DWORD` values rather than floats, and the emitted timestamps are in terms of milliseconds rather than seconds.  

### To Reproduce

After getting the asyncio example code running successfully as is, change the constructor arguments passed to `can.Bus` ([this line](https://github.com/hardbyte/python-can/blob/2e58a21578cf4451538e4e1f45cddbd85d12b488/examples/asyncio_demo.py#L23)) to instead specify interface of "neovi" and appropriate device serial number, channel, baudrate..., and be sure to include `use_system_timestamp=True` as well.  Run up the example to see exceptions being thrown by the ASC logger.

### Expected behavior

I expected no exceptions 😎 .  (You can get to the expected behavior by changing `use_system_timestamp=False` or removing that constructor argument altogether.)

### Additional context

OS and version: Windows10
Python version: 3.11.3
python-can version: 4.3.0
python-can interface/s (if applicable): neovi

[...]

PR #1734, opened 2024-01-20
Fix ASCWriter millisecond handling

lumagi:
With this PR, I would like to improve the string formatting of milliseconds in the `ASCWriter` class.
Unfortunately, the format seems to rely on localized datetimes. I added a commit to change the date time format to 24h instead of 12h. Let me know what you think.

### DIFF
--- a/can/io/asc.py
+++ b/can/io/asc.py
@@ -7,14 +7,13 @@ Example .asc files:
 """
 import logging
 import re
-import time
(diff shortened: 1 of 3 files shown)


lumagi:
Supposed to fix part of the problem presented in #1699

zariiii9003 (approved):
Looks good 👍

PR #1752, opened 2024-03-08
Add notes/kvaser support for non-iso support

jacobschaer:
Some older devices in the wild use non-iso CAN FD (Bosch). Some hardware supports choosing the mode in the driver, others are configurable during install or with a separate application. Kvaser is in the canlib API. PCAN requires either socket-can driver or the Windows GUI. Socket-CAN is an initialization flag.

Add documentation and hooks for kvaser. Relatively unintrusive. It didn't fit nicely with the bit-timings concept as it's purely associated with CRC calculation and stuffing.

### DIFF
--- a/can/interfaces/kvaser/canlib.py
+++ b/can/interfaces/kvaser/canlib.py
@@ -415,6 +415,10 @@ class KvaserBus(BusABC):
             In this case, the bit will be sampled three quanta in a row,
             with the last sample being taken in the edge between TSEG1 and TSEG2.
             Three samples should only be used for relatively slow baudrates.
+        :param bool fd_non_iso:
+            Open the channel in Non-ISO (Bosch) FD mode. Only applies for FD buses.
+            This changes the handling of the stuff-bit counter and the CRC. Defaults
+            to False (ISO mode)
 
         :param bool driver_mode:
             Silent or normal.
@@ -454,6 +458,7 @@ class KvaserBus(BusABC):
         accept_virtual = kwargs.get("accept_virtual", True)
         fd = isinstance(timing, BitTimingFd) if timing else kwargs.get("fd", False)
         data_bitrate = kwargs.get("data_bitrate", None)
+        fd_non_iso = kwargs.get("fd_non_iso", False)
 
         try:
             channel = int(channel)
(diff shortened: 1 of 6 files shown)


jacobschaer:
Fixed the test that was failing the black formatting check. The unit test that was failing appears to be intermittent and unrelated to these changes

doc/interfaces/kvaser.rst
```diff
@@ -39,6 +39,11 @@ in the ``recv`` method. If a message does not match any of the filters,
 ``recv()`` will return None.
 
 
+ISO/Non-ISO CAN FD Mode
+-----------------------
```
zariiii9003, line 43:
```suggestion
~~~~~~~~~~~~~~~~~~~~~~~
```

This heading is a bit too large. 
image
Also, i think the docstring for the new parameter is sufficient,  this paragraph does not add any additional information.

can/interfaces/kvaser/canlib.py
```diff
@@ -415,6 +415,10 @@ def __init__(
             In this case, the bit will be sampled three quanta in a row,
             with the last sample being taken in the edge between TSEG1 and TSEG2.
             Three samples should only be used for relatively slow baudrates.
+        :param bool fd_non_iso:
```
zariiii9003, line 418: Move this closer to the *fd* parameter.

doc/interfaces/pcan.rst
```diff
@@ -37,7 +37,14 @@ Here is an example configuration file for using `PCAN-USB <https://www.peak-syst
 ``bitrate`` (default ``500000``)
  Channel bitrate
 
+ISO/Non-ISO CAN FD Mode
```
zariiii9003, line 40: The heading here also looks out of place. IMO the paragraph is not necessary.

zariiii9003:
@lumagi What do you think about adding `CanProtocol.CAN_FD_NON_ISO`?

lumagi:
> @lumagi What do you think about adding `CanProtocol.CAN_FD_NON_ISO`?

Although it is very specific to this interface, I generally don't see any reasons against it. The enum already has multiple values so there's no real harm in one more. And it would certainly do away with the mess of book flags.

jacobschaer:
What value would there be in tracking this in bus's protocol property? For all use cases I've seen ISO and Non-ISO CANFD should behave identically at the app layer - the only difference is some bits in the header change but that's at a layer below python-can. A bus can't mix/match frames between the two types and I don't see any interesting uses of protocol outside of the bus implementations.

Ultimately you still need a kwarg for the Bus init, there's no way to guess if it's ISO or not from say bittimings or other flags.

I agree that it's unfortunate that more drivers don't expose this in a way that python-can can exploit.

I can update the docs per the recommendations later.

jacobschaer:
I made some updates per the suggestions.

jacobschaer:
Fixed the black issue -would be nice if that was in the default workflow but oh well.

PR #1638, opened 2023-07-21
add read support for trc v1.3

oliverbl:

PR #1753, opened 2024-03-12
Add trc 1.3 support

1atabey1:
Adds support for .trc log files as generated by e.g. PCAN Explorer v5

After implementing the change I noticed that there is already two open PRs implementing the same (:
-> https://github.com/hardbyte/python-can/pull/1638 and https://github.com/hardbyte/python-can/pull/1727

However this might be easier to merge since build passes (when i run it locally) and a corresponding test has been added


1atabey1:
hmm not really sure why tests are failing for python 3.11 (windows) only, maybe someone can trigger a re-run?

zariiii9003 (approved):
Thanks for adding a test! Looks good.

PR #1757, opened 2024-03-15
Add ability to read and write FlexRayVFrReceiveMsgEx in BLF

paperplane110:


danielhrisca:
How is this supposed to work? there is another set o functions that need to be used to setup the FlexRay bus, it can only be setup on proper FlexRay hardware channels ... etc

paperplane110:
> How is this supposed to work? there is another set o functions that need to be used to setup the FlexRay bus, it can only be setup on proper FlexRay hardware channels ... etc

Hi @danielhrisca, I only add the ability to modify Flexray message from BLF, not from the hardware.

PR #1758, opened 2024-03-16
Support 11-bit identifiers in the serial interface

KB1RD:
I'm working on a piece of software that will eventually run on a Raspberry Pi (ideally using the `SocketCAN` interface), but I would like to easily test on my local computer. I have several microcontrollers, but no way to connect the CAN bus to my computer. With the CAN serial interface, I can program the microcontrollers to act as an gateway for my computer, but the serial interface doesn't support 11-bit CAN IDs. Since my project uses 11-bit CAN IDs, I'm looking to add this support.

I added support for 11-bit CAN IDs by placing them at the end of the `arbitration_id` space for 29-bit IDs, and using the 30th bit to enable 11-bit IDs. I thought it might be useful to others, so I tested and documented the change.

I have considered making a new interface that adds more features to serial. I have also considered using the SLCAN interface, but it's honestly easier to make this change than it is to implement full SLCAN compatibility.

### Additional context
I am new to this project, and I'm not sure if this is a feature that is really wanted in the upstream. If it is, feel free to make edits to this PR.

### DIFF
--- a/can/interfaces/serial/serial_can.py
+++ b/can/interfaces/serial/serial_can.py
@@ -135,7 +135,8 @@ class SerialBus(BusABC):
 
         # Pack arbitration ID
         try:
-            arbitration_id = struct.pack("<I", msg.arbitration_id)
+            arbitration_id = msg.arbitration_id + (0 if msg.is_extended_id else 0x20000000)
+            arbitration_id = struct.pack("<I", arbitration_id)
         except struct.error:
             raise ValueError(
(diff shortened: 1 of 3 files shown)


hardbyte (approved):
Nice job, thanks for the contribution

zariiii9003:
Formatting test is failing now, but i'll follow up with another PR to add remote and error frames in the same way.

PR #1763, opened 2024-04-06
Added extra info for Kvaser dongles

FedericoSpada:
Changed Kvaser auto-detection function to return additional info regarding the found dongles, such as Serial Number, Name and Dongle Channel number.
In this way it's possible to discern between different physical dongles connected to the PC and if they are Virtual Channels or not.

### DIFF
--- a/can/interfaces/kvaser/canlib.py
+++ b/can/interfaces/kvaser/canlib.py
@@ -471,7 +471,7 @@ class KvaserBus(BusABC):
         log.info("Found %d available channels", num_channels)
         for idx in range(num_channels):
             channel_info = get_channel_info(idx)
-            channel_info = f"{channel_info["name"]}, S/N {channel_info["serial"]} (#{channel_info["dongle channel"]})"
+            channel_info = f"{channel_info["device_name"]}, S/N {channel_info["serial"]} (#{channel_info["dongle channel"]})"
             log.info("%d: %s", idx, channel_info)
             if idx == channel:
                 self.channel_info = channel_info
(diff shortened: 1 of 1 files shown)


can/interfaces/kvaser/canlib.py
```diff
@@ -788,8 +798,10 @@ def get_channel_info(channel):
         ctypes.sizeof(number),
     )
 
-    name_decoded = name.value.decode("ascii", errors="replace")
-    return f"{name_decoded}, S/N {serial.value} (#{number.value + 1})"
-
+    return {
+        "name": name.value.decode("ascii", errors="replace"),
```
hardbyte, line 802:
```suggestion
        "device_name": name.value.decode("ascii", errors="replace"),
```

hardbyte (approved):
Looks good to me. Thanks for the contribution.

can/interfaces/kvaser/canlib.py
```diff
@@ -471,6 +471,7 @@ def __init__(
         log.info("Found %d available channels", num_channels)
         for idx in range(num_channels):
             channel_info = get_channel_info(idx)
+            channel_info = f"{channel_info["device_name"]}, S/N {channel_info["serial"]} (#{channel_info["dongle channel"]})"
```
zariiii9003, line 474: Double quotes inside of double quotes is a syntax error in older Python versions.

FedericoSpada: Sorry, I just started using Python last year with the latest version and I'm unaware of these retrocompatibility issues. Can I change it using single quotes for the f-string but maitain the dict indexing inside it with double quotes? Thanks in advance

PR #1765, opened 2024-04-16
Raising CanOperationError if the bus is not open on some methods call

pierreluctg:
Raising :exc:`~can.exceptions.CanOperationError` if the bus is not open on some methods call


zariiii9003 (approved):
Looks good. The formatting needs to be fixed though.

pierreluctg:
> Looks good. The formatting needs to be fixed though.

Fixed

PR #1767, opened 2024-04-21
Fixed pcan Unpack error

FedericoSpada:
One of my colleagues encountered a "not enough values to unpack" error while scanning for connected dongles with PCAN drivers installed.
I couldn't replicate the error on my computer after installing the latest PCAN drivers. Still, analysing the instruction that raised that error, it seems that `PCANBasic.GetValue()` returned a wrong Tuple for a specific error condition.
That method must always return a Tuple with 2 elements, otherwise `PcanBus._detect_available_configs()` raises a "not enough values to unpack" error at line 718.

### DIFF
--- a/can/interfaces/pcan/basic.py
+++ b/can/interfaces/pcan/basic.py
@@ -968,6 +968,7 @@ class PCANBasic:
                 res = self.GetValue(Channel, PCAN_ATTACHED_CHANNELS_COUNT)
                 if TPCANStatus(res[0]) != PCAN_ERROR_OK:
                     return (TPCANStatus(res[0]),)
+                    return TPCANStatus(res[0]), ()
                 mybuffer = (TPCANChannelInformation * res[1])()
 
             elif (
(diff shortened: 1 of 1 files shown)


can/interfaces/pcan/basic.py
```diff
@@ -967,7 +967,7 @@ def GetValue(self, Channel, Parameter):
             elif Parameter == PCAN_ATTACHED_CHANNELS:
                 res = self.GetValue(Channel, PCAN_ATTACHED_CHANNELS_COUNT)
                 if TPCANStatus(res[0]) != PCAN_ERROR_OK:
-                    return (TPCANStatus(res[0]),)
+                    return TPCANStatus(res[0]), ()
```
zariiii9003, line 970:
```suggestion
                    return TPCANStatus(res[0]), None
```

FedericoSpada: Since you expect to receive an Iterable, returning an empty tuple is better so you don't have to check if it is "None" before using it in a For loop.

zariiii9003: This is the error case. There is no valid result, so it is not used here:

https://github.com/hardbyte/python-can/blob/7dba4490c6c61385dc8cbe917e85492b4777b8f6/can/interfaces/pcan/pcan.py#L720

PR #1771, opened 2024-04-24
Gracefully handle errors when binding SocketCAN fails

acolomb:
The parent class `BusABC` expects to be shutdown properly, checked via its `self._is_shutdown` flag during object deletion.  But that flag is only set when the base class' `shutdown()` is called, which doesn't happen if instantiation of the concrete class failed in `can.Bus()`.  So there is no way to avoid the warning message "SocketcanBus was not properly shut down" if the constructor raised an exception.

This change addresses that issue for the SocketCAN interface, by catching an OSError exception in `bind_socket()`, logging it, and calling `self.shutdown()` before re-raising the exception.

There might be a better approach which also works for the other backends?  The flag could be initialized to `True` and only reset to `False` when the `BusABC` constructor runs (which happens last in derived classes)?

Thus leaving this as a draft for further discussion.

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -716,8 +716,6 @@ class SocketcanBus(BusABC):  # pylint: disable=abstract-method
             )
         except OSError as error:
             log.error("Could not access SocketCAN device %s (%s)", channel, error)
-            # Clean up so the parent class doesn't complain about not being shut down properly
-            self.shutdown()
             raise
         super().__init__(
             channel=channel,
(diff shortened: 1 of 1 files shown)


acolomb:
The alternative approach is now proposed in #1774.  This PR could still be useful in addition, to have a proper error log message. If that is desired...

hardbyte (approved):
Happy to merge this too - perhaps remove the comment and the shutdown call

acolomb:
Adjusted as requested. Thanks for your consideration.

acolomb:
Ping @hardbyte, want to merge this?

PR #1774, opened 2024-04-25
Invert default value logic for BusABC._is_shutdown

acolomb:
The destructor of BusABC gives a warning when shutdown() was not previously called on the object after construction.  However, if the construction fails (e.g. when the derived bus class constructor raises an exception), there is no way to call shutdown() on the unfinished object, and it is not necessary either.

Initialize the _is_shutdown flag to False initially and flip it to True only when the parent class constructor runs, which usually happens last in derived classes.  That avoids the shutdown warning for objects that failed to initialize at all.

This is an alternative approach solving the same problem as #1771, but that PR could be merged independently to have an error log message before raising the constructor exception.

### DIFF
--- a/can/bus.py
+++ b/can/bus.py
@@ -65,7 +65,8 @@ class BusABC(metaclass=ABCMeta):
     #: Log level for received messages
     RECV_LOGGING_LEVEL = 9
 
-    _is_shutdown: bool = False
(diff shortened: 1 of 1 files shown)


hardbyte (approved):
Thanks - seems very reasonable.

PR #1776, opened 2024-05-08
fix mac os ip link not support error

Huibean:
```shell
  File "/opt/homebrew/Cellar/python@3.11/3.11.7_1/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['ip', '-json', 'link', 'list', 'up']' returned non-zero exit status 255.
win32com.client module required for usb2can
The SYSTEC ucan library has not been initialized.
```
fix ip link command not support raise error in mac os

### DIFF
--- a/can/interfaces/socketcan/utils.py
+++ b/can/interfaces/socketcan/utils.py
@@ -7,6 +7,8 @@ import json
 import logging
 import os
 import struct
+import sys
+import re
 import subprocess
 from typing import List, Optional, cast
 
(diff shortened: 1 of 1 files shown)


zariiii9003:
Too complicated. What you see there is just a log entry.
It would be sufficient to only log this exception on linux: `log.exception("failed to fetch opened can devices from ip link")`

zariiii9003:
Closing in favor of #1800

PR #1784, opened 2024-05-31
feat: Add zlgcan device support

zhuyu4839:
Add ZLG-CAN device support


zhuyu4839:
From the current point of view, using entry-points does not require embedding the code into python-can.
Just use `pip install zlgcan`.

PR #1789, opened 2024-06-07
Return timestamps relative to epoch in neovi interface

pierreluctg:
Return timestamps relative to epoch in neovi interface.
noeVI HW timestamps are relative to Jan 1, 2007 ("ICS epoch").
Adding logic to return messages with timestamps relative to January 1, 1970 (epoch).

### DIFF
--- a/can/interfaces/ics_neovi/neovi_bus.py
+++ b/can/interfaces/ics_neovi/neovi_bus.py
@@ -12,6 +12,7 @@ import functools
 import logging
 import os
 import tempfile
+import time
 from collections import Counter, defaultdict, deque
 from functools import partial
 from itertools import cycle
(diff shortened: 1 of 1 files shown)


can/interfaces/ics_neovi/neovi_bus.py
```diff
@@ -68,6 +69,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 open_lock = FileLock(os.path.join(tempfile.gettempdir(), "neovi.lock"))
 description_id = cycle(range(1, 0x8000))
 
+ICS_EPOCH = time.strptime("1/1/2007", "%m/%d/%Y")
+ICS_EPOCH_DELTA = time.mktime(ICS_EPOCH) - time.mktime(time.gmtime(0))
```
zariiii9003, line 73: `time.mktime(time.gmtime(0))` causes an OverflowError for me.

Maybe it's better to use datetime to calculate the delta:
```python3
ICS_EPOCH_DELTA = (
    datetime.fromisoformat("2007-01-01") - datetime.fromisoformat("1970-01-01")
).total_seconds()
```

PR #1790, opened 2024-06-09
gs_usb command-line support (and documentation updates and stability fixes)

BenGardiner:
with these changes it is now possible to use the gs_usb driver with e.g. can.logger from the command line (like many other interfaces) with `python -m can.logger -i gs_usb -c 0 -b 500000`

also now possible to send CAN frames from applications that set the timestamp field of the python-can can frame object (e.g. scapy)

### DIFF
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -17,7 +17,7 @@ class GsUsbBus(can.BusABC):
     def __init__(
         self,
         channel,
-        bitrate,
+        bitrate: int = 500_000,
         index=None,
         bus=None,
         address=None,
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -109,7 +109,7 @@ class GsUsbBus(can.BusABC):
         frame = GsUsbFrame()
         frame.can_id = can_id
         frame.can_dlc = msg.dlc
-        frame.timestamp_us = int(msg.timestamp * 1000000)
+        frame.timestamp_us = 0  # timestamp frame field is only useful on receive
         frame.data = list(msg.data)
 
         try:
--- a/can/interfaces/gs_usb.py
+++ b/can/interfaces/gs_usb.py
@@ -33,11 +33,16 @@ class GsUsbBus(can.BusABC):
         :param can_filters: not supported
         :param bitrate: CAN network bandwidth (bits/s)
         """
+        self._is_shutdown = False
         if (index is not None) and ((bus or address) is not None):
             raise CanInitializationError(
                 "index and bus/address cannot be used simultaneously"
             )
 
+        if index is None and address is None and bus is None:
+            index = channel
+
+        self._index = None
         if index is not None:
             devs = GsUsb.scan()
             if len(devs) <= index:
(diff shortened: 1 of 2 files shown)


can/interfaces/gs_usb.py
```diff
@@ -38,13 +38,18 @@ def __init__(
                 "index and bus/address cannot be used simultaneously"
             )
 
+        if index is None and address is None and bus is None:
+            index = channel
+
+        self.index = None
```
zariiii9003, line 44: Make these additional index and bitrate attributes private (add '_'-prefix).

can/interfaces/gs_usb.py
```diff
@@ -144,3 +150,13 @@ def _recv_internal(
     def shutdown(self):
         super().shutdown()
         self.gs_usb.stop()
+        if self.index is not None:
+            # Avoid errors on subsequent __init() by repeating the .scan() and .start() that would then fail here
+            devs = GsUsb.scan()
+            gs_usb = devs[self.index]
+            try:
+                gs_usb.set_bitrate(self.bitrate)
```
zariiii9003, line 158: Is this really necessary? Seems weird to call these methods in `shutdown()`. The comment does not really explain much.

doc/interfaces/gs_usb.rst
```diff
@@ -8,13 +8,14 @@ and candleLight USB CAN interfaces.
...
 Install: ``pip install "python-can[gs_usb]"``
 
-Usage: pass device ``index`` (starting from 0) if using automatic device detection:
+Usage: pass device ``index`` or ``channel`` (starting from 0) if using automatic device detection:
 
 ::
 
     import can
 
     bus = can.Bus(interface="gs_usb", channel=dev.product, index=0, bitrate=250000)
```
zariiii9003, line 17: Not related to your change, but `dev` is referenced before assignment here. Maybe you could improve this, too.

can/interfaces/gs_usb.py
BenGardiner: Yeah... it unfortunately is. Right now every second time python-can gs_usb interface tries to use the device there is a USBError thrown.

```
  File "C:\Users\Ben Gardiner\src\python-can\can\interfaces\gs_usb.py", line 65, in __init__
    self.gs_usb.start()
  File "C:\Users\Public\miniconda3\Lib\site-packages\gs_usb\gs_usb.py", line 50, in start
    self.gs_usb.reset()
  File "C:\Users\Public\miniconda3\Lib\site-packages\usb\core.py", line 961, in reset
    self._ctx.backend.reset_device(self._ctx.handle)
  File "C:\Users\Public\miniconda3\Lib\site-packages\usb\backend\libusb0.py", line 618, in reset_device
    _check(_lib.usb_reset(dev_handle))
  File "C:\Users\Public\miniconda3\Lib\site-packages\usb\backend\libusb0.py", line 447, in _check
    raise USBError(errmsg, ret)
usb.core.USBError: [Errno None] b'libusb0-dll:err [reset] could not reset device, win error: A device attached to the system is not functioning.\r\n\n'
```

I don't fully understand why though so the hamfisted way to fix it is to do the second open (start) that we know will fail in the shutdown method.

BenGardiner: I've added what I hope is a more descriptive comment there

zariiii9003 (approved):
Thanks!

PR #1795, opened 2024-06-12
Remove abstractmethod() decorator from Listener.stop()

### LINKED ISSUE #1770: Listener.stop() method breaks API since declared abstract
In #1724, the `can.listener.Listener.stop()` method was decorated as an `@abstractmethod`, without any further explanation.  This seems to work alright within the python-can codebase, but other dependent projects will get in trouble if they don't implement that method on derived classes.  At the very least should this API change be documented prominently.

This actually causes trouble with e.g. https://github.com/christiansandberg/canopen/, see https://github.com/christiansandberg/canopen/blob/713f005af8a6787a4ff54a26f55fa6736f86abaf/canopen/network.py#L346

### Expected behavior
This API change should be clearly documented OR reverted, going back to the empty default implementation.

### Additional context

OS and version: Linux Ubuntu 22.04
Python version: 3.11.6
python-can version: v4.3.1-16-g7dba4490 (current main branch)
python-can interface/s (if applicable): does not matter

<details><summary>Example script to reproduce</summary>

```python
import can

class MessageListener(can.Listener):

        pass

MessageListener()
# TypeError: Can't instantiate abstract class MessageListener with abstract method stop
```
</details>

zariiii9003:
Closes #1770 

This is breaking some downstream code.

### DIFF
--- a/can/listener.py
+++ b/can/listener.py
@@ -46,8 +46,7 @@ class Listener(metaclass=ABCMeta):
         """
         raise NotImplementedError()
 
-    @abstractmethod
-    def stop(self) -> None:
+    def stop(self) -> None:  # noqa: B027
         """
         Stop handling new messages, carry out any final tasks to ensure
         data is persisted and cleanup any open resources.


acolomb:
@hardbyte, @zariiii9003  Will you make a release with this fix soon?  We're kind of trying to hurry on a release of [canopen](https://github.com/christiansandberg/canopen/) now, but having this in a v4.4.1 on PyPI would reduce the amount of troubles for downstream users in the short term.

PR #1803, opened 2024-06-23
Release 4.4.2

zariiii9003:
@hardbyte I forgot to update the `__version__` in my 4.4.1 PR, so the PyPI upload failed. Then i changed the version and updated the release with a new git tag, but updating the release does not trigger a PyPI upload. 

So here is 4.4.2 😄


hardbyte:
I've done that a few times! 😅

Issue #1824, opened 2024-07-30
BLF reader CAN_MESSAGE and CAN_MESSAGE2 incomplete


pratik-mahamuni:
### Describe the bug
The class `BLFReader` from `can.io.blf` is implemented using Toby Lorenz's Vector_BLF. The implementation for reading CAN messages especially for `CAN_MESSAGE` and `CAN_MESSAGE2` is incorrect.

#### In this library

The code in this library related to this is as follows:
```python
if obj_type in (CAN_MESSAGE, CAN_MESSAGE2):
    channel, flags, dlc, can_id, can_data = unpack_can_msg(data, pos)
    yield Message(
        timestamp=timestamp,
        arbitration_id=can_id & 0x1FFFFFFF,
        is_extended_id=bool(can_id & CAN_MSG_EXT),
        is_remote_frame=bool(flags & REMOTE_FLAG),
        is_rx=not bool(flags & DIR),
        dlc=dlc,
        data=can_data[:dlc],
        channel=channel - 1,
    )
```
`data=can_data[:dlc]` uses the data length code to slice data bytes.

#### In Vector_BLF

For `CAN_MESSAGE`:
```cpp
void CanMessage::read(AbstractFile & is) {
    ObjectHeader::read(is);
    is.read(reinterpret_cast<char *>(&channel), sizeof(channel));
    is.read(reinterpret_cast<char *>(&flags), sizeof(flags));
    is.read(reinterpret_cast<char *>(&dlc), sizeof(dlc));
    is.read(reinterpret_cast<char *>(&id), sizeof(id));
    is.read(reinterpret_cast<char *>(data.data()), static_cast<std::streamsize>(data.size()));
}
```

For `CAN_MESSAGE2`:
```cpp
void CanMessage2::read(AbstractFile & is) {
    ObjectHeader::read(is);
    is.read(reinterpret_cast<char *>(&channel), sizeof(channel));
    is.read(reinterpret_cast<char *>(&flags), sizeof(flags));
    is.read(reinterpret_cast<char *>(&dlc), sizeof(dlc));
    is.read(reinterpret_cast<char *>(&id), sizeof(id));
    data.resize(objectSize - calculateObjectSize()); // all remaining data
    is.read(reinterpret_cast<char *>(data.data()), static_cast<std::streamsize>(data.size()));
    is.read(reinterpret_cast<char *>(&frameLength), sizeof(frameLength));
[...]

PR #1825, opened 2024-07-31
Modified data slicing in BLFReader for when the object type is CAN_ME…

pratik-mahamuni:
Merge request for #1824.

Please accept if it meets your internal standards.
Thank you

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,8 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-can"
-dynamic = ["readme", "version"]
+dynamic = ["readme"]
+version = "4.3.1"
 description = "Controller Area Network interface module for Python"
 authors = [{ name = "python-can contributors" }]
 dependencies = [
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -4,7 +4,8 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "python-can"
-dynamic = ["readme", "version"]
+dynamic = ["readme"]
+version = "4.3.1"
```
zariiii9003, line 8: Please undo this. The version is set by setuptools-scm from the git tag.

pratik-mahamuni: This is now undone. Please accept if no other problems. Thank you

zariiii9003: @pratik-mahamuni the tests are failing

pratik-mahamuni: @zariiii9003 modified the method `equals` from the `Message` class based on the intricacies of how the tests are modeled in the repository. This change could be deemed temporary as the current implementation of the loggers_test.py uses the same data and constructor for all file types (ASC, BLF, CSV etc.).

As `equals` is used only in tests and not for logic in the code (based on a quick `grep`), the small change should not have side-effects. I will let you be the judge for that.

PR #1844, opened 2024-08-26
Make Message a dataclass

NickCao:


NickCao:
`weakref_slot` is only available in python > 3.11.

zariiii9003:
The performance impact is not worth the benefits as of now, so i'm going to close this.
We can come back to this once python 3.11 is the minimum supported version of python-can.

Issue #1849, opened 2024-09-04
PCAN fails when PCAN_ERROR_ILLDATA is read via ReadFD


bures:
### Describe the bug
When there is an invalid frame on CAN bus (in our case CAN FD), PCAN first reports result PCAN_ERROR_ILLDATA and then it send the error frame. If the PCAN_ERROR_ILLDATA is not ignored, python-can throws an exception.

### To Reproduce
A packet like the one in this screenshot has to be received. 
![image](https://github.com/user-attachments/assets/c82a149d-329a-49c6-ad11-6e712c59824d)

### Expected behavior
python-can should ignore the PCAN_ERROR_ILLDATA because it carries no useful information. The error can be then detected by the error frame which PCAN generates right after.

### Additional context

OS and version: Windows 11
Python version: Python 3.12
python-can version: Latest commit in main - c7121310253c42c6410a70bc8b4f27a51b8b4411
python-can interface/s (if applicable): PCAN-USB FD (using pcan interface)

PR #1850, opened 2024-09-04
Fix for #1849 (PCAN fails when PCAN_ERROR_ILLDATA is read via ReadFD)

bures:
When there is an invalid frame on CAN bus (in our case CAN FD), PCAN first reports result PCAN_ERROR_ILLDATA and then it send the error frame. If the PCAN_ERROR_ILLDATA is not ignored, python-can throws an exception. Reported as #1849 

This fix adds a branch in pcan.py:_recv_internal, which ignores the PCAN_ERROR_ILLDATA.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -45,6 +45,7 @@ from .basic import (
     PCAN_ERROR_BUSLIGHT,
     PCAN_ERROR_OK,
     PCAN_ERROR_QRCVEMPTY,
+    PCAN_ERROR_ILLDATA,
     PCAN_FD_PARAMETER_LIST,
     PCAN_LANBUS1,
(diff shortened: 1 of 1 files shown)


can/interfaces/pcan/pcan.py
```diff
@@ -555,6 +556,10 @@ def _recv_internal(
             elif result & (PCAN_ERROR_BUSLIGHT | PCAN_ERROR_BUSHEAVY):
                 log.warning(self._get_formatted_error(result))
 
+            elif result == PCAN_ERROR_ILLDATA:
+                # Ignore this
```
zariiii9003, line 560: Add some explanation here and fix the import order, then i'll merge.

bures: Done. Thanks.

zariiii9003 (approved):

Issue #1780, opened 2024-05-15
Align with the new CAN frame structure for socketcan adding optional DLC element


rahagal:
### Is your feature request related to a problem? Please describe.
 
SocketCAN allows for passing an optional DLC element in its `can_frame` structure in addition to the length element, this allow accessing the raw DLC value that can be 0-15 (with 9-15 interpreted as 8 in classic CAN).
This feature is good for testing compliance with the ISO.
To enable this some changes are implemented in the Linux kernel and in can-utils.
You can see the change being introduced to can-utils in here: https://github.com/linux-can/can-utils/commit/c398e56afb9b60441232736e05fdfb5a525363ca
 
### Describe the solution you'd like
 
I would like to suggest an adaptation to the way CAN_FRAME_HEADER_STRUCT is being defined and un/packed.
 
### Describe alternatives you've considered
I did not find a way to achieve this with the current implementation.
 
### Additional context
See a patch used in a fork I've created that allowed me to achieve this capability.
[socktcan_over8_dlc.patch](https://github.com/hardbyte/python-can/files/15244735/socktcan_over8_dlc.patch)

eldadcool:
Your patch has a bug regarding the message length calculation.
You need to perform the len function on the msg.data and not data. because data is a padded buffer with length 64 or 8 and not the actual length of the original message.
I'm planning to attach my fix which is much like yours in a pull request.

PR #1851, opened 2024-09-05
CC-2174: modify CAN frame header structure to match updated struct ca…

eldadcool:
This fix allow to send CLASSIC CAN frames with a DLC value larger than 8 using the socketcan interface.

In addition it allow to parse incoming socketcan messages with the updated format
#1780 
@hartkopp


zariiii9003:
@lumagi could you take a look at this?

lumagi:
I'll try to take a look over the next couple of days. @eldadcool do you happen to have a link to the patch set in the kernel?

eldadcool:
> I'll try to take a look over the next couple of days. @eldadcool do you happen to have a link to the patch set in the kernel?

sure:
The patch can be found at version 6.10.7 of the kernel, see patch at https://github.com/torvalds/linux/commit/ea7800565a128c1adafa1791ce80afd6016fe21c

To configure the bus using "ip link" command you require a version of iproute2 5.12 or higher: https://lore.kernel.org/netdev/20210427165946.0bbd8fc0@hermes.local/

use the command 
```bash
ip link set can0 down
ip link set type can cc-len8-dlc on
ip link set can0 up
```

note: can-utils support this only since this patch https://github.com/linux-can/can-utils/commit/11edd1d05e6fac6f45706d98ccb9ddd621621874

@hartkopp - please review my fix as well to make sure I did not misinterpret  some edge cases regarding the new structure and that this will not break the usage of the old format.

hartkopp:
> use the command
> 
> ```shell
> ip link set can0 down
> ip link set type can cc-len8-dlc on
> ip link set can0 up
> ```

You can simply go with the latest can-utils where `candump` and `cangen` have a `-8` option to create and display the DLC values over 8.
WIth the virtual CAN interfaces (e.g. `vcan0`) you can test the cc-len8-dlc stuff without any additional configuration. The `vcan` is transparent for these values.

> @hartkopp - please review my fix as well to make sure I did not misinterpret some edge cases regarding the new structure and that this will not break the usage of the old format.

I'm not sure whether I can really review the patch, as I'm not that familiar with Pythons packing concept.
But if you strictly followed the data structure and the rules (len == 8 and ( len8_dlc > 8 && len8_dlc <= 0xF )) apply for valid DLC over 8 length's values - then it should be fine :-)

Many thanks,
Oliver

eldadcool:
> I'm not sure whether I can really review the patch, as I'm not that familiar with Pythons packing concept. But if you strictly followed the data structure and the rules (len == 8 and ( len8_dlc > 8 && len8_dlc <= 0xF )) apply for valid DLC over 8 length's values - then it should be fine :-)

Actually, the DLC value in python-can relate directly to the data length. and it's not even verified to be a valid message by default. So potentially this could lead to sending a message with unexpected DLC value.
This reminds me of an issue I think I found in socketcan (@hartkopp) allowing unexpected lengths for CAN FD frames (10 bytes for example...) because the length value used is socketcan also relate to the data length and not to the DLC. I think we need to be aligned regarding this and very careful because fixing it might break some existing usage of socketcan or python-can.

Note the issue with socketcan only relevant for virtual can interfaces because physical interfaces does not allow FD frames with 10 bytes of data (only <8, 8, 12, 16, 20, 24, 32, 48 and 64) but it makes virtual interfaces behave differently than physical interfeces.

hartkopp:
> This reminds me of an issue I think I found in socketcan (@hartkopp) allowing unexpected lengths for CAN FD frames (10 bytes for example...) because the length value used is socketcan also relate to the data length and not to the DLC. I think we need to be aligned regarding this and very careful because fixing it might break some existing usage of socketcan or python-can.
> 
> Note the issue with socketcan only relevant for virtual can interfaces because physical interfaces does not allow FD frames with 10 bytes of data (only <8, 8, 12, 16, 20, 24, 32, 48 and 64) but it makes virtual interfaces behave differently than physical interfeces.

Yes, this is intentional. 
The idea is that the content which is sent via virtual CAN devices is not extensively sanitized to be able to test with uncertain values and just reflect them back to user space. This is very useful to test user space applications.

Once the CAN frame content is given to a **real** CAN interface or received from them the functions

`can_fd_dlc2len()` `can_fd_len2dlc()` `canfd_sanitize_len()` are used.

https://elixir.bootlin.com/linux/v6.11/source/include/linux/can/length.h#L290

eldadcool:
> Yes, this is intentional.
> The idea is that the content which is sent via virtual CAN devices is not extensively sanitized to be able to test with uncertain values and just reflect them back to user space. This is very useful to test user space applications.

Cool,
In that case the current solution I proposed here sanitize the length field for FD valid values. Don't think this should lead to an issue. @lumagi, have any insights?

can/interfaces/socketcan/socketcan.py
```diff
@@ -260,11 +321,24 @@ def build_bcm_update_header(can_id: int, msg_flags: int, nframes: int = 1) -> by
...
 
 def dissect_can_frame(frame: bytes) -> Tuple[int, int, int, bytes]:
-    can_id, can_dlc, flags = CAN_FRAME_HEADER_STRUCT.unpack_from(frame)
+    can_id, data_len, flags, len8_dlc = CAN_FRAME_HEADER_STRUCT.unpack_from(frame)
     if len(frame) != constants.CANFD_MTU:
         # Flags not valid in non-FD frames
         flags = 0
-    return can_id, can_dlc, flags, frame[8 : 8 + can_dlc]
+
+    if data_len not in can.util.CAN_FD_DLC:
```
lumagi, line 329: Is there a specific reason you don't trust the value returned by the kernel? I'm not saying it's a bad thing, just interested in why you added this.

can/interfaces/socketcan/socketcan.py
```diff
@@ -260,11 +321,24 @@ def build_bcm_update_header(can_id: int, msg_flags: int, nframes: int = 1) -> by
...
+    # Allow deprecated can frames with old struct
+    if (
+        data_len == constants.CAN_MAX_DLEN
+        and constants.CAN_MAX_DLEN < len8_dlc <= constants.CAN_MAX_RAW_DLC
+    ):
+        can_dlc = len8_dlc
+    else:
+        can_dlc = data_len
+
+    return can_id, can_dlc, flags, frame[8 : 8 + data_len]
```
lumagi, line 341: This is very nit-picky: I know this wasn't any better before the new PR, but I don't like that we don't check that the expected `data_len` is actually present in the frame. Since it's Python it will simply return less data than expected.

lumagi (commented):
It looks fine to me. I only had some nit-picky stuff.

zariiii9003:
ping @eldadcool

can/interfaces/socketcan/socketcan.py
eldadcool: As discussed with @hartkopp socketcan allow to send non valid FD frames with length not supported by CAN FD. For example, DLC=0x9 (data_len=12) and actual data_len=10 - this is invalid CAN FD frame but still could be received from the kernel.
To handle this case I added code to round up the amount of bytes note that no read overflow will happen because the CANFD MTU is set to 72 (64 bytes of data) in the socketcan implementation so the additional bytes read from the frame will still be part of the struct and will probably contains 0x00s. If you prefer we can add padding with a defined value to be on the safe side.

can/interfaces/socketcan/socketcan.py
eldadcool: Sure I can add this verification to make it more robust. Should I add padding bytes , recude the data_len value or raise an exception in that case? Not sure what will be the best here...

eldadcool:
> ping @eldadcool

Hii sorry, I was not very available last month...
Added replies to your comments. :)

can/interfaces/socketcan/socketcan.py
eldadcool: @lumagi - what approach do you think will be better?

lumagi: I didn't think about the fact that the buffer returned from the kernel always has the maximum length and so the slicing operation should always provide the necessary amount of bytes. Please disregard my comment.

zariiii9003:
@lumagi feel free to approve and merge

can/interfaces/socketcan/socketcan.py
```diff
@@ -260,11 +321,24 @@ def build_bcm_update_header(can_id: int, msg_flags: int, nframes: int = 1) -> by
...
         # Flags not valid in non-FD frames
         flags = 0
-    return can_id, can_dlc, flags, frame[8 : 8 + can_dlc]
+
+    if data_len not in can.util.CAN_FD_DLC:
+        data_len = min(i for i in can.util.CAN_FD_DLC if i >= data_len)
+
+    # Allow deprecated can frames with old struct
+    if (
+        data_len == constants.CAN_MAX_DLEN
```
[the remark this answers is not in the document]
eldadcool: Amm yes you are correct, maybe I should add tests for this case as well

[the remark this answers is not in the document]
rahagal: I've added the non-FD condition - after syncing with @eldadcool

Issue #1848, opened 2024-08-30
Bus.send_periodic() should not start automatically


SWolfSchunk:
### Is your feature request related to a problem? Please describe.
I am writing an UI with a button to start and stop a can send task that initially should not send. Stopping the task immediately after creating leads to a single sending of the message(s).

### Describe the solution you'd like
 An optional kwarg eg. "start_on_init: bool = True" would help out here to keep backward compatibility and i could set it to False.

### Describe alternatives you've considered
There is no alternative else than changing the package itself in broadcastmanager.py and commenting out self.start() in __init__ of ThreadBasedCyclicSendTask.

### Additional context
I have no additional context so far

PR #1853, opened 2024-09-06
Add autostart option to BusABC.send_periodic() to fix issue #1848

SWolfSchunk:
I have fixed my own issue #1848

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -350,13 +350,13 @@ class CyclicSendTask(
 
         self.bcm_socket = bcm_socket
         self.task_id = task_id
-        self._tx_setup(self.messages, send=autostart)
+        if autostart:
+            self._tx_setup(self.messages)
 
     def _tx_setup(
         self,
         messages: Sequence[Message],
         raise_if_task_exists: bool = True,
-        send: bool = True,
     ) -> None:
         # Create a low level packed frame to pass to the kernel
         body = bytearray()
@@ -380,9 +380,7 @@ class CyclicSendTask(
         for message in messages:
             body += build_can_frame(message)
         log.debug("Sending BCM command")
-
-        if send:
-            send_bcm(self.bcm_socket, header + body)
+        send_bcm(self.bcm_socket, header + body)
 
     def _check_bcm_task(self) -> None:
         # Do a TX_READ on a task ID, and check if we get EINVAL. If so,
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -813,6 +813,7 @@ class SocketcanBus(BusABC):  # pylint: disable=abstract-method
         msgs: Union[Sequence[Message], Message],
         period: float,
         duration: Optional[float] = None,
+        autostart: bool = True,
         modifier_callback: Optional[Callable[[Message], None]] = None,
     ) -> can.broadcastmanager.CyclicSendTaskABC:
         """Start sending messages at a given period on this bus.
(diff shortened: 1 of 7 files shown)


hardbyte (approved):
Seems reasonable

SWolfSchunk:
It seems like the test "testProducerConsumer" fails randomly on random pypy versions (3.9, 3.10) on random platforms (windows here so far, but other requests show the same behavior on ubuntu and macos).
I can not appraise if it is really a problem depending on the changes of other requests but here i am sure it is not coming from the changes I made. Maybe I should open a bug report fort that but maybe this is a problem of the threading implementation of pypy?

zariiii9003:
The autostart parameter is ignored by socketcan and ixxat, right? This should either be fixed or documented.
And yes, the PyPy tests are flaky, it's probably not related to your changes.

can/interfaces/socketcan/socketcan.py
```diff
@@ -823,13 +824,16 @@ def _send_periodic_internal(
         :class:`CyclicSendTask` within BCM provides flexibility to schedule
         CAN messages sending with the same CAN ID, but different CAN data.
 
-        :param messages:
+        :param msgs:
```
grant-allan-ctct, line 827: 👍

zariiii9003:
> No that's not right the autostart parameter is not ignored by neither socketcan nor ixxat. 

Your change only affects the thread-based fallback. However, the socketcan broadcast manager start immediately.

SWolfSchunk:
Hopefully I have fixed the missing parts now. Thanks for your hint @zariiii9003 .
I am still not sure about the changes for socketcan devices.
From how i read the code of its CyclicSendTask.__init__() instead of other devices they where not autostarting before why i set the default value to autostart=False in this case for backward compatibility.

can/interfaces/socketcan/socketcan.py
```diff
@@ -376,7 +380,9 @@ def _tx_setup(
         for message in messages:
             body += build_can_frame(message)
         log.debug("Sending BCM command")
-        send_bcm(self.bcm_socket, header + body)
+
+        if send:
```
zariiii9003, line 384: I think this check should be in some other place, but the whole class needs some refactoring. I'll fix this later.

can/interfaces/socketcan/socketcan.py
```diff
@@ -823,13 +830,16 @@ def _send_periodic_internal(
...
+        :param msgs:
             The message(s) to be sent periodically.
         :param period:
             The rate in seconds at which to send the messages.
         :param duration:
             Approximate duration in seconds to continue sending messages. If
             no duration is provided, the task will continue indefinitely.
+        :param autostart:
+            If True (the default) the sending task will immediately start after creation.
+            Otherwise, the task has to be started by calling start() on it.
```
zariiii9003, line 842: Same as above, `start()` should be linked.

zariiii9003 (changes requested):

can/interfaces/socketcan/socketcan.py
```diff
@@ -349,10 +350,13 @@ def __init__(
...
         self.task_id = task_id
-        self._tx_setup(self.messages)
+        self._tx_setup(self.messages, send=autostart)
 
     def _tx_setup(
-        self, messages: Sequence[Message], raise_if_task_exists: bool = True
+        self,
+        messages: Sequence[Message],
+        raise_if_task_exists: bool = True,
+        send: bool = True,
```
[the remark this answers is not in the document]
SWolfSchunk: To make everything clearer I have had a closer look at the code and decided to guard the whole  call of self._tx_setup() with if autostart to remove the send parameter as the name autostart would not really fit.
I also think that the name of the _tx_setup() should be changed to make clearer what it really does.

can/interfaces/socketcan/socketcan.py
SWolfSchunk: Se the comment above

zariiii9003 (approved):

zariiii9003:
Thank you 👍

PR #1858, opened 2024-09-12
Faster Message string representation

pierreluctg:
Improved the speed of data conversion to hex string.
This also as a result improves the `can.Printer` logger performance.

### Main Branch
```
$ python -m timeit -s "import can" "str(can.Message(data=(_ for _ in range(64))))"
20000 loops, best of 5: 14.7 usec per loop
```

### This PR
```
$ python -m timeit -s "import can" "str(can.Message(data=(_ for _ in range(64))))"
100000 loops, best of 5: 3.47 usec per loop
```


zariiii9003 (approved):

felixdivo (approved):
🥳

PR #1878, opened 2024-10-21
Fix Kvaser timestamp

### LINKED ISSUE #1877: kvaser timestamps broken since 4.4.1
### Describe the bug

Timestamps from kvaser messages are errouneous, due to the timestamp offset applied is incorrectly initialized in init function. 
Looks like the root cause is that the calls to canBusOn were moved in version 4.4.1 to now be done after the calls to kvReadTimer used to initialize the timestamp offset, this is causing the kvReadTimer to return a value that is not right (a way much bigger value than what is returned as timestamps on messages received later!). 

### To Reproduce

Check timestamps on received messages from _recv_internal function and compare them with the value of time.time() when the function returned, it should be close to zero but is not. 

### Expected behavior

Timestamps should be accurate and in sync with values returned by time.time()

### Additional context

OS and version: Windows 10
Python version: 3.11.7
python-can version: 4.4.2
python-can interface/s (if applicable): Kvaser Leaf light v2

```python
def func():
    return "hello, world!"
```
</details>

zariiii9003:
Fixes #1877

### DIFF
--- a/can/interfaces/kvaser/canlib.py
+++ b/can/interfaces/kvaser/canlib.py
@@ -586,7 +586,7 @@ class KvaserBus(BusABC):
             canBusOn(self._read_handle)
         canBusOn(self._write_handle)
 
-        # timestamp must be set after bus is online
+        # timestamp must be set after bus is online, otherwise kvReadTimer may return erroneous values
         self._timestamp_offset = self._update_timestamp_offset()
 
     def _update_timestamp_offset(self) -> float:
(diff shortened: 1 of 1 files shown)


can/interfaces/kvaser/canlib.py
```diff
@@ -573,35 +573,39 @@ def __init__(
...
+            **kwargs,
+        )
+
+        # activate channel after CAN filters were applied
+        log.debug("Go on bus")
+        if not self.single_handle:
+            canBusOn(self._read_handle)
+        canBusOn(self._write_handle)
+
+        # timestamp must be set after bus is online
```
johanb321, line 589: To be picky, it would be nice with a comment on why here: 
_...since otherwise kvReadTimer seem to return erroneous values_

johanb321 (approved):

PR #1882, opened 2024-10-30
socketcan: support use of SO_TIMESTAMPING for hardware timestamps

pevsonic:
The current implemenation of socketcan utilises SO_TIMESTAMPNS which only offers system timestamps.

I've looked at how can-utils candump.c configures hardware timestamping and implemented this in socketcan as an option which is disabled by default to avoid any potential adverse impact on existing usage. This is using the same param 'use_system_timestamp' as established by neovi_bus.py.

I've also modified logger.py to provide an additional '-H' flag in the same way that candump does in order to use this functionality.


can/interfaces/socketcan/socketcan.py
```diff
@@ -619,6 +634,7 @@ def __init__(
         self,
         channel: str = "",
         receive_own_messages: bool = False,
+        use_system_timestamp: bool = True,
```
hartkopp, line 637: The new option should tell about the hardware timestamping and not about the system timestamps which is the default for a good reason: It is always available.
The hardware timestamps are read from some CAN controller chips and its support is not very common.

Instead of `use_system_timestamp` I would suggest to provide a new option:
`can_hardware_timestamps: bool = False,`

So when people really know what they are doing, enabling this feature should be possible.

pevsonic: Ill modify to adopt option naming as suggested - makes sense to me.

WRT defaults etc, my initial suggestion of defaulting to SO_TIMESTAMPNS makes sense as I'm a cautious kinda guy. However, reading the kernel docs :
[](https://www.kernel.org/doc/Documentation/networking/timestamping.txt)

I think one should be able to request SOF_TIMESTAMPING_RAW_HARDWARE and either setsockopt() will fail allowing one to fallback to system generated timestamps, or one can use the system timestamp in the cmsg array that's returned.   Annoyingly I don't have enough hardware to test this otherwise Id have had a crack at the same time! Worth thinking about for the future however...

hartkopp: To provide the best user experience, the standard (sw) timestamp should be provided in the best resolution available. In whatever way this can be retrieved by the system.

But the hardware timestamping provides values that might not fit to the common system time. And therefore this should be an explicit setting for people that are aware of such details.

pevsonic: > But the hardware timestamping provides values that might not fit to the common system time. And therefore this should be an explicit setting for people that are aware of such details.

If referencing system time is helpful, given that the system timestamp and hardware timestamp are both returned in a cmsg array, one could potentially read the system timestamp as well for the first received timestamp and calculate the offset which you apply equally to all subsequent frame timestamps. This would add little overhead but would give you equivalent clock time but with the precision of the hardware timestamp?

can/interfaces/socketcan/socketcan.py
```diff
@@ -642,6 +658,9 @@ def __init__(
             channel using :attr:`can.Message.channel`.
         :param receive_own_messages:
             If transmitted messages should also be received by this bus.
+        :param bool can_hardware_timestamps:
+            Use hardware timestamp for can messages if available instead of
+            the system timestamp.
```
zariiii9003, line 663: You could mention `SO_TIMESTAMPNS` and `SO_TIMESTAMPING` here. You could also mention possible pitfalls as @hartkopp mentioned

can/interfaces/socketcan/socketcan.py
```diff
@@ -703,12 +723,25 @@ def __init__(
...
-        self.socket.setsockopt(socket.SOL_SOCKET, constants.SO_TIMESTAMPNS, 1)
+        if not self.can_hardware_timestamps:
+            # Utilise SOF_TIMESTAMPNS interface :
+            # we can always do this since
+            #  1) it is guaranteed to be at least as precise as without
+            #  2) it is available since Linux 2.6.22, and CAN support was only added afterward
+            #     so this is always supported by the kernel
+            self.socket.setsockopt(socket.SOL_SOCKET, constants.SO_TIMESTAMPNS, 1)
+        else:
+            # Utilise SOF_TIMESTAMPNS interface :
```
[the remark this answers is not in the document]
pevsonic: good spot, yes.

can/logger.py
```diff
@@ -46,6 +46,13 @@ def _create_base_argument_parser(parser: argparse.ArgumentParser) -> None:
         choices=sorted(can.VALID_INTERFACES),
     )
 
+    parser.add_argument(
```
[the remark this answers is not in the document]
pevsonic: From my perspective, enabling hardware timestamping is a core feature I use and I imagine others might see it that way too. As an argument, it's availability is advertised by --help. Sending as you suggest would (I don't believe?) have any way for the user to see the args availability without understanding the mechanism and digging into the code to find the arg name. Additionally theres no validation on them so if I used `--can-hard-we-ar-timestamps=True` there'd be no detection of this, failure or feedback to the user. 

Ive not got a big issue with this if you'd like me to drop the argument, but I think there's a solid case for leaving as an argument?

can/interfaces/socketcan/socketcan.py
```diff
@@ -659,6 +678,7 @@ def __init__(
         self.socket = create_socket()
         self.channel = channel
         self.channel_info = f"socketcan channel '{channel}'"
+        self.can_hardware_timestamps = can_hardware_timestamps
```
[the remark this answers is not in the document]
pevsonic: done

can/interfaces/socketcan/socketcan.py
pevsonic: done

can/interfaces/socketcan/socketcan.py
```diff
@@ -642,6 +656,17 @@ def __init__(
             channel using :attr:`can.Message.channel`.
         :param receive_own_messages:
             If transmitted messages should also be received by this bus.
+        :param bool can_hardware_timestamps:
+            Use raw hardware timestamp for can messages if available instead
+            of the system timestamp. By default we use the SO_TIMESTAMPNS
+            interface which provides ns resolution but low accuracy. If your
+            can hardware supports it you can use this parameter to
```
zariiii9003, line 663:
```suggestion
            CAN hardware supports it you can use this parameter to
```

can/logger.py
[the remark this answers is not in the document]
zariiii9003: If only one interface supports this argument, i'd prefer to remove it. Otherwise we could add 100 more interface specific "important" arguments.

can/interfaces/socketcan/socketcan.py
```diff
@@ -642,6 +656,17 @@ def __init__(
...
+            alternatively use the SO_TIMESTAMPING interface and request raw
+            hardware timestamps. These are much higher precision but will
+            almost certainly not be referenced to the time of day. There
+            may be other pitfalls to such as loopback packets reporting with
```
zariiii9003, line 667:
```suggestion
            may be other pitfalls too such as loopback packets reporting with
```

can/interfaces/socketcan/socketcan.py
```diff
@@ -642,6 +656,17 @@ def __init__(
             channel using :attr:`can.Message.channel`.
         :param receive_own_messages:
             If transmitted messages should also be received by this bus.
+        :param bool can_hardware_timestamps:
+            Use raw hardware timestamp for can messages if available instead
```
zariiii9003, line 660:
```suggestion
            Use raw hardware timestamp for CAN messages if available instead
```

pevsonic:
> Any update on this?

I’d love to get it in as I’m using it in derived code we have here and I’m sure it’s helpful, but I’m not sure what changes I need to make to get accepted?

jayceslesar:
> I’d love to get it in as I’m using it in derived code we have here and I’m sure it’s helpful, but I’m not sure what changes I need to make to get accepted?

I think you can just remove the argparse code you added and things should just work as the library knows to just pass in kwargs to the bustype by using the following:

```py
    parser.add_argument(
        "extra_args",
        nargs=argparse.REMAINDER,
        help="The remaining arguments will be used for the interface and "
        "logger/player initialisation. "
        "For example, `-i vector -c 1 --app-name=MyCanApp` is the equivalent "
        "to opening the bus with `Bus('vector', channel=1, app_name='MyCanApp')",
    )
```

So you would just pass 
`--can-hardware-timestamps=True`
like @zariiii9003 commented in the initial review.

This is beneficial because this library acts as an abstraction over many bustypes abstracted over and made to look like CAN buses even though they are not and allows users to pass in arbitrary "kwargs" to the underlying interface that actually gets initialized without having to worry about what arguments actually live on the argument parser object

The rest just look like minor docs changes

pevsonic:
> > I’d love to get it in as I’m using it in derived code we have here and I’m sure it’s helpful, but I’m not sure what changes I need to make to get accepted?
> 
> I think you can just remove the argparse code you added and things should just work as the library knows to just pass in kwargs to the bustype by using the following:
> 
> ```python
>     parser.add_argument(
>         "extra_args",
>         nargs=argparse.REMAINDER,
>         help="The remaining arguments will be used for the interface and "
>         "logger/player initialisation. "
>         "For example, `-i vector -c 1 --app-name=MyCanApp` is the equivalent "
>         "to opening the bus with `Bus('vector', channel=1, app_name='MyCanApp')",
>     )
> ```
> 
> So you would just pass `--can-hardware-timestamps=True` like @zariiii9003 commented in the initial review.
> 
> This is beneficial because this library acts as an abstraction over many bustypes abstracted over and made to look like CAN buses even though they are not and allows users to pass in arbitrary "kwargs" to the underlying interface that actually gets initialized without having to worry about what arguments actually live on the argument parser object
> 
> The rest just look like minor docs changes

That's really helpful thanks! I didn't know that's how kwargs works - while I can get by with Python, I'm really a 'C' guy..! Hopefully that should be enough for a merge now...

Issue #1883, opened 2024-10-30
FEATURE REQUEST: Support for MDF4 CAN bus logging files [PR ready]


MatinF:
### Is your feature request related to a problem? Please describe.

The current MDF4 reader in the `python-can` project is only able to load MDF4 files generated by the MDF4 writer in the project, or MDF4 files having the exact same internal structure. This severely limits the number of MDF4 files with CAN data which can be loaded into the `python-can` ecosystem for analysis and playback.

### Describe the solution you'd like

An expansion of the MDF4 reader capabilities to support MDF4 CAN bus logging files in general. I.e. instead of always expecting CAN DataFrames to be group 0, allow these to be any group, and even span multiple groups. The `asammdf` package the reader/writer leverages for the heavy lifting behind the scenes already supports loading of more complex files/other structures.

We can provide a PR for expanding the MDF4 readers capabilities to support more generalized CAN bus logging files (in fact we have already prepared and tested one).

Please let us know whether this is something you are open to, then we will formally prepare the PR.

MatinF:
As an update, we have now made a PR for this:
https://github.com/hardbyte/python-can/pull/1884

johnMediaOnFire:
I think you can close this as a duplicate of #1884.

PR #1884, opened 2024-11-01
Updates to the MF4 reader to enable more general MF4 support

MatinF:
Related to feature request posted below:
https://github.com/hardbyte/python-can/issues/1883

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,10 @@ jobs:
       with:
         python-version: ${{ matrix.python-version }}
         allow-prereleases: true
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install tox
     - name: Setup SocketCAN
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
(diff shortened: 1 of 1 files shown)


hardbyte (commented):
Thanks, will be great to support more Mf4 files.

My only concern really is how this will behave with large mf4 files on resource constrained systems as this looks to load all DataFrame data into memory at once during class initialization instead of iterating through the data frame by frame. 

Obviously there should be upsides to this bulk processing if a device has the resources for it, but consider a lower footprint approach.

MatinF:
Hi Brian, thanks for your feedback! 

We have made an updated PR and include a few comments below:

1) The previous PR was focused on speed performance, rather than RAM utilization. However, our updated PR should match the RAM usage of the current implementation, while drastically improving speed and interoperability. 

Below is an example comparison of performance:

Data: 10 MB finalized MF4 file 

Close mimic of current implementation:
- Speed:  128.7 s
- max RAM: 14 MB

Previous PR:
- Speed: 1.3 s
- max RAM: 128 MB

Updated PR (using 1000 row chunk size):
- Speed: 7.6 s
- max RAM: 14 MB

In other words, our proposed alternative PR would yield the same RAM usage, while improving the readout speed by 20x+. If you believe the stats above and the proposed alternative PR would be acceptable, we can update the PR accordingly.


2) For the other comments, we have tried to address these in the updated PR

zariiii9003:
I recommend to rebase onto the main branch, there was a problem with CI.

PR #1893, opened 2024-11-10
Improve speed of TRCReader

lebuni:
Optimized TRCReader for speed, mainly for TRC files v2.x. According to cProfile it's double as fast now.

### DIFF
--- a/can/io/trc.py
+++ b/can/io/trc.py
@@ -11,15 +11,12 @@ import logging
 import os
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from typing import Any, Callable, Dict, Generator, List, Optional, TextIO, Union
+from typing import Any, Callable, Dict, Generator, Optional, TextIO, Tuple, Union
 
 from ..message import Message
 from ..typechecking import StringPathLike
-from ..util import channel2int, dlc2len, len2dlc
-from .generic import (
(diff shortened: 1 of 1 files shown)


can/io/trc.py
```diff
@@ -244,7 +229,7 @@ def _parse_cols_v2_x(self, cols: List[str]) -> Optional[Message]:
     def _parse_line(self, line: str) -> Optional[Message]:
         logger.debug("TRCReader: Parse '%s'", line)
         try:
-            cols = line.split()
+            cols = tuple(line.split(maxsplit=self._num_columns))
```
zariiii9003, line 232: Why do you convert the list to a tuple?

can/io/trc.py
```diff
@@ -58,13 +55,16 @@ def __init__(
         """
         super().__init__(file, mode="r")
         self.file_version = TRCFileVersion.UNKNOWN
-        self.start_time: Optional[datetime] = None
+        self.start_time: float = 0
```
zariiii9003, line 58: This attribute is public, so changing the type could break user code. I'd suggest creating a private float attribute `_timestamp` and a property `timestamp`, that returns a datetime instance like before.

Otherwise this looks good to me.

lebuni: Good point! Changed it

PR #1894, opened 2024-11-12
Added support for ZLG CAN

luojiaaoo:
Support for python-can in accordance with the python-can interface


zariiii9003:
Please contribute to the existing plugin instead: https://github.com/zhuyu4839/zlgcan-driver

luojiaaoo:
OK, I would look into zhuyu4839/zlgcan-driver

PR #1903, opened 2024-12-16
Resolve SW-13543 "Update the first line of asc file to use logging start time instead of asc file created time"

AmanKAggarwal:
This is because we create the asc file when the user requests to export their log file into asc format. Using the file created time means when they open the generated ASC file in CANoe, CANoe identifies that created time as the session start time for data which is incorrect


can/io/asc.py
```diff
@@ -394,6 +394,14 @@ def stop(self) -> None:
         # This is guaranteed to not be None since we raise ValueError in __init__
         if not self.file.closed:
             self.file.write("End TriggerBlock\n")
+            if self.started is not None:
```
zariiii9003, line 397: Why do you check for None?

PR #1907, opened 2025-01-04
Disable error frames when ignore_rx_error_frames is set

polymorf:
Changing `ignore_rx_error_frames` for socketcan interfaces has no effect since error filter is only applied when `ignore_rx_error_frames` is false

### DIFF
--- a/can/interfaces/socketcan/socketcan.py
+++ b/can/interfaces/socketcan/socketcan.py
@@ -702,6 +702,14 @@ class SocketcanBus(BusABC):  # pylint: disable=abstract-method
                 )
             except OSError as error:
                 log.error("Could not enable error frames (%s)", error)
+        else:
+            # disable error frames
+            try:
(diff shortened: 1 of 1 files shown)


can/interfaces/socketcan/socketcan.py
```diff
@@ -702,6 +702,14 @@ def __init__(
                 )
             except OSError as error:
                 log.error("Could not enable error frames (%s)", error)
+        else:
```
zariiii9003, line 705: You could simplify this a bit like this:
```python3
            # set error frame filter
            try:
                self.socket.setsockopt(
                    constants.SOL_CAN_RAW,
                    constants.CAN_RAW_ERR_FILTER,
                    0x0 if ignore_rx_error_frames else 0x1FFFFFFF,
                )
            except OSError as error:
                log.error("Could not enable error frame filter (%s)", error)
```

PR #1915, opened 2025-01-30
mf4: asammdf update to 8.0.0 for macOS build

tuna-f1sh:
The current dependency of 6.0.0 fails to build on modern Apple Silicon: https://github.com/danielhrisca/asammdf/issues/1038.

Requires bump of min python-can version to 3.9 too. It looks like this was added in the 8.0.0 release. The >= 7.0.0 would still work with 3.8 if this is an issue.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "typing_extensions>=3.10.0.0",
     "msgpack~=1.1.0; platform_system != 'Windows'",
 ]
-requires-python = ">=3.8"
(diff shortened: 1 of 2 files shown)


zariiii9003:
It's a lower limit, what's stopping you from installing a newer version?

tuna-f1sh:
You're right apologies. For some reason (I guess cache) uv resolved 6.0.0 when I added the mf4 feature. I then thought the default was to install 6.0.0 but it appears to install latest.

PR #1916, opened 2025-01-30
Adding env marker to pywin32 requirements (extra)

pierreluctg:
As part of #1833, the `pywin32` requirements environment markers got dropped when moving the requirement to an extra.


zariiii9003 (approved):
That was intentional. I thought: If the user explicitely requests pywin32, then installation should fail if it is not available.

But i can see use cases, where that might be annoying, so approved :)

PR #1920, opened 2025-02-18
add FD support to slcan according to CANable 2.0 impementation

ssj71:
The canable 2.0 project has extended the slcan interface to add support for CAN-FD frames and bitrate switching (see https://github.com/normaldotcom/canable2-fw).  I needed this support for my work so I've implemented it in the library. If this work is useable I'd love to have it merged in. Happy to make tweaks or whatever, just need some direction. 

Thank you for maintaining this project. It's immensely useful.


ssj71:
fixed the black and ruff failings.

CallmeLins:
Hi Bro, I have modify the code as your commit, send message is ok, but seems can not received message (windows platform), do you have this issue?

Nakakiyo092:
Hello,

Thank you for submitting this pull request.
This update will be very useful for me.

However, I noticed an issue regarding compatibility with classical devices that do not support CAN FD.
In the latest release, a device can be initialized as follows:
```
bus = can.interface.Bus(interface="slcan", channel="COM9", bitrate=500000)
```
When using your branch, the initialization fails with the following exception:
```
Invalid data bitrate, choose one of 2000000, 5000000.
```
In my opinion, the initialization should still work to support classical slcan devices.

By the way, I tested the CAN FD capability with a 500k/2Mbps bus.
I had any issue with sending/receiving on a Windows PC.

Thank you.

ssj71:
I agree my changes should not have that effect. I will update and check that useage.

ssj71:
> Hi Bro, I have modify the code as your commit, send message is ok, but seems can not received message (windows platform), do you have this issue?

I send and receive without issue. I have found that ListenOnly mode doesn't work properly. I believe it is a canable firmware issue.

ssj71:
Apologies for the delays. That fix is in to allow non-FD rates again and I believe this to be ready to merge.

can/interfaces/slcan.py
```diff
@@ -157,10 +161,12 @@ def __init__(
 
         super().__init__(channel, **kwargs)
 
-    def set_bitrate(self, bitrate: int) -> None:
+    def set_bitrate(self, bitrate: int, dbitrate: int = 0) -> None:
```
zariiii9003, line 164: I'd prefer `data_bitrate`  over `dbitrate` for consistency with other interfaces. The default value should be `None`, not zero.

zariiii9003:
Overall this looks fine except for a lack of tests.
What happens if a user tries to open an FD bus on a device, that does not support it? Will there be some exception or just silent failure?

ssj71:
I added some tests and they made it clear to me I had an error in the DLC handling in the initial implementation. That's resolved and I hope this makes the PR a bit more compelling. Let me know if there's anything else needed. Thanks!

ssj71:
> What happens if a user tries to open an FD bus on a device, that does not support it? Will there be some exception or just silent failure?
Looking at the original CANable firmware source it just silently fails. I'm not sure about other HW devices that use slcan.

ssj71:
ok, I think this one is ready (finally)!

zariiii9003 (approved):
Looks okay to me. I hope you'll be available to fix potential issues :smile:

PR #1923, opened 2025-02-22
Add can bridge

pkess:
Hello,

it might be useful to have the possibility to use the RedirectReader in a script so that one can connect two can busses with a single cli program. This is extremly helpful if one would like to connect e.g. socketcand with a proprietary hardware interface.

Comments are appreciated

### DIFF
--- a/can/bridge.py
+++ b/can/bridge.py
@@ -49,8 +49,8 @@ def get_config_list(it, separator, conf):
         el = next(it)
         if el == separator:
             break
-        else:
-            conf.append(el)
+
+        conf.append(el)
 
 
 def split_configurations(arg_list, separator="--"):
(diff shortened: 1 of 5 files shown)


zariiii9003:
I think this is may be very useful 👍 

Not a review yet, but a few notes:

- New code should have type annotations, at least for the function signatures.
- The other CLI tools have tests in `test_scripts.py` and e.g. `test_player.py`, `test_logger.py` etc., i think this bridge could also be tested with two virtual buses.
- Minor: I think the plural of bus is usually buses, not busses 🤷‍♂️

pkess:
Ok. I will work on all those three

pkess:
I addes some very basic tests now. I do not know at the moment how to test that a message that was received on one bus is forwarded to the other bus. Maybe i can try this again in the next days

zariiii9003:
Closing in favor of #1961

PR #1702, opened 2023-12-06
implement TRCWriter feature for v1.1 

hiteshhedwig:
current `TRCWriter` does not implement method for fileversion 1.1. So i have implemented it. And trc can be generated as follows

```
;$FILEVERSION=1.1
;$STARTTIME=45248.7383051968
;
;   Start time: 18-11-2023 17:43:09.569.0
;   Generated by PCAN-View v5.0.1.822
;
;   Message Number
;   |         Time Offset (ms)
;   |         |        Type
;   |         |        |        ID (hex)
;   |         |        |        |     Data Length
;   |         |        |        |     |   Data Bytes (hex) ...
;   |         |        |        |     |   |
;---+--   ----+----  --+--  ----+---  +  -+ -- -- -- -- -- -- --
     1)         2.6  Rx         0020  8  80 00 A0 06 68 03 3A 00 
     2)         3.2  Rx     18DAF101  8  10 18 62 02 2F 81 90 02 
     3)         4.7  Rx     18DA01F1  8  30 0A 0A 55 55 55 55 55 
     4)        11.7  Rx     0C0AA7F0  8  00 00 00 00 00 00 00 00 
```

Please feel free to leave any feedback

PR #1924, opened 2025-02-23
Trc file writer add version param

pkess:
While i tried to update and finish the changes in [PR 1702](https://github.com/hardbyte/python-can/pull/1702) i saw some basic change that are quite usefull but are not the trace file version 1.1 changes itself. I collected them here now.

I also added the docstring as it was requested by @zariiii9003 .

I will create another PR for the addition of the TRC version 1.1 itself.

### DIFF
--- a/can/io/trc.py
+++ b/can/io/trc.py
@@ -310,8 +310,15 @@ class TRCWriter(TextIOMessageWriter):
         self.header_written = False
         self.msgnr = 0
         self.first_timestamp = None
-        self.file_version = TRCFileVersion.V2_1
-        self._msg_fmt_string = self.MESSAGE_FORMAT_MAP[self.file_version]
+        self._setup_file_version(TRCFileVersion.V2_1)
+
+    def _setup_file_version(self, file_version: Union[int, TRCFileVersion]):
+        try:
+            self.file_version = TRCFileVersion(file_version)
+            self._msg_fmt_string = self.MESSAGE_FORMAT_MAP[self.file_version]
+        except (KeyError, ValueError) as exc:
+            err_msg = f"File version is not supported: {file_version}"
+            raise NotImplementedError(err_msg) from exc
 
     def _write_header_v1_0(self, start_time: datetime) -> None:
         lines = [
--- a/can/io/trc.py
+++ b/can/io/trc.py
@@ -378,7 +378,7 @@ class TRCWriter(TextIOMessageWriter):
         ]
         self.file.writelines(line + "\n" for line in lines)
 
-    def _format_message(self, msg, channel):
+    def _format_message(self, msg: Message, channel: int) -> str:
         if msg.is_extended_id:
             arb_id = f"{msg.arbitration_id:07X}"
         else:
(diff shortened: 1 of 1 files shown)


pkess:
I tried to but i was not able to fix the sphix build. It looks like sphinx cannot resove the TRCFileVersion to `can.TRCFileVersion`.

Can anyone help me?

zariiii9003:
> I tried to but i was not able to fix the sphix build. It looks like sphinx cannot resove the TRCFileVersion to `can.TRCFileVersion`.
> 
> Can anyone help me?

Sphinx fails to create a link to TRCFileVersion, because it does not exist in the documentation yet. You could add 
```
.. autoclass:: can.TRCFileVersion
    :show-inheritance:
    :members:
    :undoc-members:
```
somewhere in `file_io.rst`.

pkess:
Thank you for that hint. I will try that later

can/io/trc.py
```diff
@@ -308,9 +322,15 @@ def __init__(
...
         self.msgnr = 0
         self.first_timestamp = None
-        self.file_version = TRCFileVersion.V2_1
-        self._msg_fmt_string = self.FORMAT_MESSAGE_V1_0
-        self._format_message = self._format_message_init
+        self._setup_file_version(file_version)
+
+    def _setup_file_version(self, file_version: Union[int, TRCFileVersion]) -> None:
+        try:
+            self.file_version = TRCFileVersion(file_version)
```
zariiii9003, line 329: instance attributes should be set in `__init__`. You could have version and format string as return values of a static method or just remove this method and put the try/except block into `__init__`

pkess: ok. The branch where i picked this from handled it with a tuple returned. I think this does not improve readability.

As this method is invoked from the __init__ function it is always ensured that the attributes are initialized. Can you tell my why it should be added explicit in __init__?

pkess: I googled a bit for this and it seems like the only requirement is that all attributes are already initialized after the __init__ function was executed. It is not enforced that the assignment is inside of the __init__ function. So for me it looks like the approach here is correct.

I prefer to try to make a function implementation fit on one or two screen page. With this implementation we are quite prepared for more additions and changes not to blow up the function.

If we return a tuple we would achieve the same that is correct, but i think it is better readable like this.

So please take another look and tell me your thoughts.

zariiii9003: My thoughts are still the same. Instance attributes belong inside `__init__`. Here is the output of pyrefly for example:
```
ERROR /home/user/repositories/python-can/can/io/trc.py:317:13-34: Object of class `TextIO` has no attribute `reconfigure` [missing-attribute]
ERROR /home/user/repositories/python-can/can/io/trc.py:329:18-30: Attribute `file_version` is implicitly defined by assignment in method `_setup_file_version`, which is not a constructor [implicitly-defined-attribute]
ERROR /home/user/repositories/python-can/can/io/trc.py:330:18-33: Attribute `_msg_fmt_string` is implicitly defined by assignment in method `_setup_file_version`, which is not a constructor [implicitly-defined-attribute]
```

So, it would be nice if you fix it, so i can merge.

PR #1927, opened 2025-02-26
Fix float arithmetic in BLF reader

pierreluctg:
Using the Python decimal module for fast correctly rounded decimal floating-point arithmetic when applying the timestamp factor.

The BLF unit tests are also updated to reflect this new accuracy in the read timestamps.

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -17,6 +17,7 @@ import logging
 import struct
 import time
 import zlib
+from decimal import Decimal
 from typing import Any, BinaryIO, Generator, List, Optional, Tuple, Union, cast
 
 from ..message import Message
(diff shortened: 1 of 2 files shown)


can/io/blf.py
```diff
@@ -264,8 +265,8 @@ def _parse_data(self, data):
                 continue
 
             # Calculate absolute timestamp in seconds
-            factor = 1e-5 if flags == 1 else 1e-9
-            timestamp = timestamp * factor + start_timestamp
+            factor = Decimal("1e-5") if flags == 1 else Decimal("1e-9")
+            timestamp = float(Decimal(timestamp) * factor) + start_timestamp
```
zariiii9003, line 269: I think that can be achieved without Decimal:
```python3
            factor = 100_000 if flags == 1 else 1_000_000_000
            timestamp = (timestamp + round(start_timestamp * factor)) / factor
```

pierreluctg: I agree that we can achieve it without `Decimal`.
But what is the downside of using `Decimal`?

IMO, using the Python built-in module `Decimal` makes it very obvious what we are doing here.

felixdivo (approved):

Issue #1403, opened 2022-10-04
Support for PCAP file format


pkess:
### Is your feature request related to a problem? Please describe.
I would like to analyze some can traffic with wireshark. With Wireshark it is possible to check e.g. CANopen communication as you see here: https://wiki.wireshark.org/CANopen.md

### Describe the solution you'd like
There is the pcap file formt supported by wireshark to load can traces https://wiki.wireshark.org/Development/LibpcapFileFormat
It would be nice to use python-can to record or convert into pcap file format for further processing

Tbruno25:
It appears like wireshark includes a utility for converting regular ascii hex logs to pcap -- https://www.wireshark.org/docs/man-pages/text2pcap.html

PR #1936, opened 2025-04-08
support pcapng file format

dsseng:
This PR adds support for Pcapng files to the I/O module

Related to #1403, pcap should be rather simple to implement based on these changes.

- **Factor out common SocketCAN ser/des parts**
- **Add PcapngWriter**
- **Add PcapngReader**


can/io/logger.py
```diff
@@ -40,6 +40,14 @@
 from .printer import Printer
 from .sqlite import SqliteWriter
 from .trc import TRCWriter
+from .pcapng import PcapngWriter
+
+
+try:
+    import pyzstd
+except ImportError:
+    pyzstd = None
```
dsseng, line 49: Zstandard addition should be a distinct commit. Please let me know if you would like to merge with squashing so that I can split it out into a new PR

zariiii9003:
Keep the changes minimal. Move the socketcan files back to where they belong, remove zstd and all other unrelated changes. Add test data and documentation (see file_io.rst)

dsseng:
> Keep the changes minimal. Move the socketcan files back to where they belong, remove zstd and all other unrelated changes. Add test data and documentation (see file_io.rst)

I abstracted out some common SocketCAN files for the purpose of not duplicating the code. pcapng expects packets to be serialized as SocketCAN packets, so this code has to be shared. Let me know if there is a better place or way to refactor this.

zariiii9003:
you could just import from socketcan without moving the files

dsseng:
> you could just import from socketcan without moving the files

Is this a correct pattern for an IO module to import an interface module? That felt a bit counter-logical

PR #1942, opened 2025-05-27
Add metadata object type to BLFReader

jonmon6691:
BLF files have metadata objects that encode useful information such as CAN channel mapping configuration. This PR provides the minimal additions to the BLFReader class to support extracting that data along with the parsing of the message objects. Parsing the data structures used in the metadata objects is outside the scope of this PR. However this still provides a significant benefit to applications that are can perform context specific interpretation of the metadata without having to duplicate the BLF parsing performed by BLFReader.

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -77,9 +77,13 @@ CAN_ERROR_EXT_STRUCT = struct.Struct("<HHLBBBxLLH2x8s")
 # group name length, marker name length, description length
 GLOBAL_MARKER_STRUCT = struct.Struct("<LLL3xBLLL12x")
 
+# Metadata object type, 4 context-specific bytes, metadata length, 4 more
+# context-specific bytes
+METADATA_HEADER_STRUCT = struct.Struct("<I4BI4B")
 
 CAN_MESSAGE = 1
 LOG_CONTAINER = 10
+METADATA = 65
 CAN_ERROR_EXT = 73
 CAN_MESSAGE2 = 86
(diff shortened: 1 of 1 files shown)


zariiii9003:
Object type 65 should be `APP_TEXT`. You can look [here](https://bitbucket.org/tobylorenz/vector_blf/src/master/src/Vector/BLF/AppText.cpp) or [here](https://github.com/zariiii9003/vblf/blob/3eff6d3f0f7e3b622052876383d1bbf30ad90bb2/src/vblf/general.py#L479) for reference.

jonmon6691:
Thanks for pointing that out @zariiii9003 and I wasn't aware of the vblf project. Does it make sense to have this functionality in BLFReader? It seems like the API here is focused on just returning objects that appeared on the bus in the iterator, and abstracting away the fact that there are other object types in the BLF. It seems a little awkward to have a side channel in self.app_text that is only populated as a side-effect of iterating for messages.

can/io/blf.py
```diff
@@ -173,6 +182,11 @@ def __init__(
         )
         # Read rest of header
         self.file.read(header[1] - FILE_HEADER_STRUCT.size)
+
+        # APP_TEXT objects are appended when type 65 is encountered
+        # while iterating
+        self.app_text: List[Tuple] = []
```
zariiii9003, line 188: Instead of a tuple, i'd prefer an AppText or BlfAppText dataclass with timestamp, source and text attributes.

PR #1953, opened 2025-07-10
Add TRCReader RTR frames support

denis-jullien:
TRCReader currently fails to read trc files containing RTR messages (from PCAN-View). This PR provide support for reading files with RTR messages for all trc versions.


zariiii9003 (approved):
Thank you!

PR #1957, opened 2025-07-18
Convert BusState to enum when read with configparser

Gregox273:
Fix bug in pcan interface where state values cannot be read from config files.

### DIFF
--- a/can/interfaces/pcan/pcan.py
+++ b/can/interfaces/pcan/pcan.py
@@ -268,12 +268,6 @@ class PcanBus(BusABC):
 
         self.check_api_version()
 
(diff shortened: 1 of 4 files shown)


can/interfaces/pcan/pcan.py
```diff
@@ -268,6 +268,12 @@ def __init__(
 
         self.check_api_version()
 
+        if not isinstance(state, BusState):
```
zariiii9003, line 271: I think this is the wrong place for this. The config value should be converted in `can.util._create_bus_config()`.

zariiii9003 (approved):
Thank you

PR #1961, opened 2025-07-25
Add CAN bridge script

zariiii9003:
This is a rebased continuation of #1923 using the new functions from #1949.

@pkess Could you take a look?


pkess:
Looks good to me

PR #1967, opened 2025-08-04
reader: mf4: test for 'S' type .Dir like asammdf

tuna-f1sh:
I could not decode a mf4 file created by [mdflib](https://github.com/ihedvall/mdflib). It creates `.Dir` channels with string values not int: https://github.com/ihedvall/mdflib/blob/main/mdflib/src/canconfigadapter.cpp#L321

Not sure if this is standard in MDF but since the note said the reader iterators are based on the asammdf GUI code, I took a look. It does a check on the [attached type](https://github.com/danielhrisca/asammdf/blob/ca8182a80125bbb01f6bb394710b8d7f9c84b1f5/src/asammdf/mdf.py#L6135).

Commit adds the same for python-can so that it can decode files with a string value for `.Dir`.

### DIFF
--- a/can/io/mf4.py
+++ b/can/io/mf4.py
@@ -352,7 +352,10 @@ class MF4Reader(BinaryIOMessageReader):
                     if "CAN_DataFrame.BusChannel" in names:
                         kv["channel"] = int(data["CAN_DataFrame.BusChannel"][i])
                     if "CAN_DataFrame.Dir" in names:
-                        kv["is_rx"] = int(data["CAN_DataFrame.Dir"][i]) == 0
+                        if data["CAN_DataFrame.Dir"][i].dtype.kind == "S":
+                            kv["is_rx"] = data["CAN_DataFrame.Dir"][i] == b'Rx'
+                        else:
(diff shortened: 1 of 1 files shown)


zariiii9003:
According to the [standard](https://www.asam.net/standards/detail/mdf/), it should be 0 or 1.

tuna-f1sh:
Ah ok - unfortunately I don't have access to the specification. I'll create an issue in mdflib instead then. Are you able to provide a snippet of the spec saying this? Do you know why asammdf would support values as string? Is there some legacy non-explicit definition here?

zariiii9003:
The spec says it's a bit signal, not a string:
image

According to the [asammdf commit message](https://github.com/danielhrisca/asammdf/commit/bb1653b5c5e73f0892def1da6ed072a04cd45a3b#diff-a5304e9e1256fa6554ae2e4a20c0ffdd590e2b1b04e2f71161618f69aae673a6), it was added to fix CANalyzer measurements.

Anyhow, i guess there's no harm in merging, since this does not affect the files we produce. Could you fix the formatting?

tuna-f1sh:
Yes I've updated commit with `black` format.

zariiii9003 (approved):

PR #1971, opened 2025-08-07
Fix zizmor warnings and update dependabot.yml

zariiii9003:
## Summary of Changes

Fix all [zizmor](https://github.com/zizmorcore/zizmor) security warnings and use dependabot to update actions regularly.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring
- [x] Other (please describe):  GitHub Actions workflow

## Additional Notes

@hardbyte Could you activate dependabot if it isn't already? You can find it in "Settings->Advanced Security".

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,6 +6,21 @@
 version: 2
 updates:
   - package-ecosystem: "uv"
+    # Enable version updates for development dependencies
     directory: "/"
     schedule:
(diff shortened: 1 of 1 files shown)


hardbyte:
Sure, it is enabled but needs a config. Placeholder [here](https://github.com/hardbyte/python-can/pull/1972)

PR #1978, opened 2025-08-11
Change the default to 0, which is a no-op for data_bitrate.

dcasnowdon:
## Summary of Changes

- 

- Closes #
- Related to #

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [ ] I have followed the [contribution guide](https://python-can.readthedocs.io/en/main/development.html).
- [ ] All checks and tests pass (`tox`).

### DIFF
--- a/can/interfaces/slcan.py
+++ b/can/interfaces/slcan.py
@@ -166,7 +166,7 @@ class slcanBus(BusABC):
 
         super().__init__(channel, **kwargs)
 
-    def set_bitrate(self, bitrate: int, data_bitrate: Optional[int] = None) -> None:
+    def set_bitrate(self, bitrate: int, data_bitrate: Optional[int] = 0) -> None:
         """
         :param bitrate:
             Bitrate in bit/s


zariiii9003:
Hi, thank you for the fix. But I'd prefer the function signature as it is, so instead you could change the `_DATA_BITRATES` value to `None: ""`

dcasnowdon:
> Hi, thank you for the fix. But I'd prefer the function signature as it is, so instead you could change the `_DATA_BITRATES` value to `None: ""`

I wasn't sure whether 0 might be passed in as a no-op. Is it OK to change the interface in this way? Instead, I've changed it to add an explicit check for `None` in the body of the method. I can change the value in the `_DATA_BITRATES` table if you'd prefer.

dcasnowdon:
> Yes, looks good now. Could you add a little bit more information to the news fragment? I'll create a patch release afterwards

I have added a little more detail (is this what you'd like?), and I've changed it to be a `fix`, rather than `added`.

Issue #1995, opened 2025-10-07
SeeedBus: Allow setting hardware filter/mask during Bus Init


varunpenumudi:
## Main Problem
I don't like that the SeeedBus interface does not allow for the configuration of the hardware acceptance filter and mask during the bus initialization.

My use case involves a bootloader flasher client that communicates with a device over the CAN bus. The client sends a message with one CAN ID and needs to listen only for a response on another specific CAN ID. To achieve this efficiently, I need to use the hardware filter. The current implementation forces me to create the bus object, then manually change its internal attributes (filter_id, mask_id), and then re-call the internal init_frame() method. This workflow is not intuitive and feels like a workaround.

## Current Alternative
The only alternative at the moment is to manually re-configure the bus after it has been created. This involves the following steps:

```Python

import struct

# 1. Create the bus with default (disabled) filters
bus = can.interface.Bus(interface='seeedstudio', channel='/dev/ttyUSB0')

# 2. Manually overwrite the internal attributes
bus.filter_id = struct.pack('<I', 0x456 << 21)
bus.mask_id = struct.pack('<I', 0xFFFFFFFF)

# 3. Call the internal init_frame() method again to send the new config to the hardware
bus.init_frame()
```
This is not an ideal solution because it is recalling init_frame() method again after it is called in __init__, and is little more complicated.

## Want to change Like this
I would like to have filter_id and mask_id added as optional keyword arguments to the SeeedBus.__init__ method. This would allow the hardware filter to be configured cleanly and directly when the bus is created.

Example of the desired usage:

Python

```
# Initialize the bus to only accept messages with ID 0x456
bus = can.interface.Bus(
    interface='seeedstudio',
    channel='/dev/ttyUSB0',
    bitrate=500000,
    frame_type='STD',
    filter_id=0x456, 
    mask_id=0xFFFFFFFF
)

# The bus is now ready to use with the hardware filter active.
msg = bus.recv()
```

**Note:** If the filter_id and mask_id arguments are not provided, they will default to 0x00. This maintains the current behavior where the filter is disabled and all CAN IDs are received, ensuring the change is fully backward-compatible.

```
# Initialize the bus 
# Filter ID, Mask ID here defaults to 0x00 so no filtering of can messages
bus = can.interface.Bus(
    interface='seeedstudio',
    channel='/dev/ttyUSB0',
    bitrate=500000,
    frame_type='STD',
)

# The bus is now ready to use with the hardware filter active.
msg = bus.recv()
```

[...]

PR #1998, opened 2025-10-14
SeeedBus: Added can_filters parameter in Bus Init

varunpenumudi:
This pull request resolves issue #1995 by adding support for configuring the hardware acceptance filter and mask during SeeedBus initialization.

Previously, setting a hardware filter was not possible directly through the constructor. Users had to use a workaround by manually overwriting internal attributes (filter_id, mask_id) and re-calling the internal init_frame() method after the bus was already created. This process was unintuitive, broke encapsulation, and was prone to errors.

This PR implements the standard can_filters parameter in the SeeedBus.__init__ method, adhering to the BusABC interface. This allows users to set a hardware filter cleanly in a single step upon bus creation.

Key Changes:

- Added the can_filters parameter to `SeeedBus.__init__`.
- The implementation validates that only one filter is provided, raising a can.CanInitializationError if more are given, in accordance with the Seed Bus protocol limitation.
- Updated the seeedstudio.rst documentation to reflect the new parameter and its usage 
- Added a news fragment for the changelog.

**Note:** This change is fully backward-compatible. If can_filters is not provided, the bus defaults to the previous behavior of accepting all messages.

Testing
I have personally tested this feature on a Waveshare USB to CAN Adapter (Model A), which uses the same serial protocol, and can confirm that it works as expected for both default (no filter) and filtered modes.


varunpenumudi:
Dont merge these changes, this PR have some issues

PR #1999, opened 2025-10-16
SeedBus: Added can_filters parameter in SeedBus.__init__ method

varunpenumudi:
This pull request resolves issue #1995 by adding support for configuring the hardware acceptance filter and mask during SeeedBus initialization.

Previously, setting a hardware filter was not possible directly through the __init__ call of SeedBus. Users had to use a workaround by manually overwriting internal attributes (filter_id, mask_id) and re-calling the internal init_frame() method after the bus was already created. This process was unintuitive, broke encapsulation, and was prone to errors.

This PR implements the standard can_filters parameter in the SeeedBus.__init__ method, adhering to the BusABC interface. This allows users to set a hardware filter cleanly in a single step upon bus creation.

Key Changes:

- Added the can_filters parameter to `SeeedBus.__init__`.
- The implementation validates that only one filter is provided, raising a can.CanInitializationError if more are given, in accordance with the SeedBus protocol limitation.
- Updated the seeedstudio.rst documentation to reflect the new parameter and its usage 
- Added a news fragment for the changelog.

**Note:** This change is fully backward-compatible. If can_filters is not provided or None, the bus defaults to the previous behavior of accepting all messages.

Testing
I have personally tested this feature on a Waveshare USB to CAN Adapter (Model A), which uses the same serial protocol, and can confirm that it works as expected for both default (no filter) and filtered modes.

### DIFF
--- a/can/interfaces/seeedstudio/seeedstudio.py
+++ b/can/interfaces/seeedstudio/seeedstudio.py
@@ -103,8 +103,10 @@ class SeeedBus(BusABC):
 
         can_id = 0x00
         can_mask = 0x00
+        self._is_filtered = False
 
         if can_filters and len(can_filters) == 1:
+            self._is_filtered = True
             hw_filter = can_filters[0]
             can_id = hw_filter["can_id"]
             can_mask = hw_filter["can_mask"]
@@ -252,8 +254,9 @@ class SeeedBus(BusABC):
                 This parameter will be ignored. The timeout value of the
                 channel is used.
 
-        :returns:
-            Received message and False (because not filtering as taken place).
+        :return:
+            1.  a message that was read or None on timeout
+            2.  a bool that is True if hw_filter is enabled, else False
 
         :rtype:
             can.Message, bool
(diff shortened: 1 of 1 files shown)


can/interfaces/seeedstudio/seeedstudio.py
```diff
@@ -94,11 +101,23 @@ def __init__(
                 "the serial module is not installed"
             )
 
+        if can_filters is None:
+            can_filters = [{"can_id": 0x00, "can_mask": 0x00}]
+
+        if len(can_filters) > 1:
```
zariiii9003: You could use the software fallback of python-can if the user passes multiple filters. Check the implementation in BusABC.

zariiii9003:
The `_recv_internal` method should also be updated. It returns a message and a boolean value: https://github.com/hardbyte/python-can/blob/efd4de5619331d3ae8da10688979624cbb6d0342/can/bus.py#L169

If you are using hardware filtering, you should return `True`, for software filtering return `False`. You can add an attribute `self._is_filtered` and return that in in `_recv_internal`

PR #2026, opened 2026-02-05
interfaces/cantact: Handle None timeout by setting to 0xFFF...

BenGardiner:
## Summary of Changes

There are definitely cases of client code that end up passing timeout=None to _recv_internal(). Several other interfaces/*.py are prepared for this e.g. canalystii usb2can canlib_vcinpl2 gs_usb.py

For cantact I think the right thing to do is set default timeout to 0 if None is provided.
There are definitely cases of client code that end up passing timeout=None to _recv_internal(). Several other interfaces/*.py are prepared for this e.g. canalystii usb2can canlib_vcinpl2 gs_usb.py

For cantact I think the right thing to do is set default timeout to 0 if None is provided.

## Type of Change

- [X] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [X] I have followed the [contribution guide](https://python-can.readthedocs.io/en/main/development.html).
- [X] All checks and tests pass (`tox`).

### DIFF
--- a/can/interfaces/cantact.py
+++ b/can/interfaces/cantact.py
@@ -125,9 +125,7 @@ class CantactBus(BusABC):
 
     def _recv_internal(self, timeout: float | None) -> tuple[Message | None, bool]:
         if timeout is None:
-            raise TypeError(
-                f"{self.__class__.__name__} expects a numeric `timeout` value."
-            )
+            timeout = 2**64 - 1
 
         with error_check("Cannot receive message"):
             frame = self.interface.recv(int(timeout * 1000))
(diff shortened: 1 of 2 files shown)


zariiii9003:
The usb2can implementation seems correct to me. If our `timeout` is `None`, then `CanalBlockingReceive` is called with 0, which means "wait forever" according to the documentation.

The gs_usb implementation seems wrong indeed.

The candle_bus implementation also seems incorrect, but that's not part of this repo.

BenGardiner:
thank you @zariiii9003 . 

I fixed this change to cantact with a force-push. 

I made this https://github.com/BIRLab/python-can-candle/pull/5 for python-can-cantact

I am not sure how to fix gs-usb and also these days I find it hard to test it since it requires libusbK. I don't use it almost ever since the cantact (and now python-can-candle) drivers work with the exact same devices...

Issue #1746, opened 2024-02-22
slcan interface needs some time-tweaking (Possible solution provided)


K1BSS:
### Describe the bug
Messages are truncated while using send command.
Messages are truncated while using close command.

### To Reproduce
Code as example. Requires python-can and pyserial modules.
Windows PC and CANable with original firmware: https://github.com/normaldotcom/canable-fw
Counter-interface: PCAN USB dongle and oscilloscoop

Problem 1: When looping messages within the "with can.Bus", one can see on the scope that the messages are corrupted/truncated.
Problem 2: When looping messages main, one can see that the last message in the "with can.Bus" is truncated.

Possible solution:
End the _write-function with: "time.sleep(1.0e-5)"

### Expected behavior
Truncated messages: seen on a oscilloscoop. PCAN viewer tels me stuffing errors. The error will result in resending infinite retrying messages.

### Additional context
OS and version: Windows 11
Python version: Python 3.12
python-can version: 4.3.1
pyserial version: 3.5
python-can interface/s (if applicable): slcan
Hardware interface: CANable 2.0 Pro
IDE: PyCharm recent version
Oscilloscope: Rigol DS1000-series
Counter-interface: PCAN USB isolated version and PCAN-View software

Possible solution:
End the _write-function with: "time.sleep(1.0e-5)"

<details>
<summary>Traceback and logs</summary>

None

See code example.

```
def send_one():
    with can.Bus(interface="slcan", channel="COM7@115200",  bitrate=500000, sleep_after_open=0.0) as bus:
        msg = can.Message(
            arbitration_id=0x555, data=[0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55], is_extended_id=False
        )

        try:
            j = 0
            while j < 2:
                bus.send(msg)
                j = j + 1
        except can.CanError:
            print("Message NOT sent")


if __name__ == "__main__":
    i = 0
    while i < 2:
        send_one()
        i = i + 1
```

</details>

Issue #1701, opened 2023-12-04
CAN-FD Message Truncation Issue When Sending More Than 8 Bytes


PMCSilva:
### Describe the bug
When attempting to send a CAN-FD message with a payload larger than 8 bytes using 'slcan' bustype connected to a CANABLE device, the message appears to be truncated or not sent at all, despite setting `is_fd=True` in the `can.Message` object and `fd=True` in the `can.interface.Bus` initialization. Messages with 8 bytes or less are sent and received correctly.

### To Reproduce
1. Initialize a CAN interface with CAN-FD enabled.
2. Create a CAN-FD message with more than 8 bytes of data.
3. Send the message over the CAN bus.
4. Observe that the message is not sent correctly or is truncated.

### Expected behavior
The CAN-FD message with more than 8 bytes should be sent and received correctly, as per the CAN-FD specifications.

### Additional context

OS and version: Windows 11 Pro 22H2
Python version: 3.11.3
python-can version: 4.3.0
python-can interface/s (if applicable): slcan

### Code Snippet
```python
import can
 
# Initialize CAN interface with CAN-FD enabled
can_bus = can.interface.Bus(channel='COM16', bustype='slcan', bitrate=500000, fd=True)
 
# Create a CAN-FD message with more than 8 bytes of data
data = bytes.fromhex("0102000000AAAAAAA000000A")  # Example data with more than 8 bytes
message = can.Message(arbitration_id=0x100, is_extended_id=False, data=data, is_fd=True)
 
# Send the message
try:
    can_bus.send(message)
    print("Message sent")
except can.CanError as e:
    print(f"Error sending CAN message: {e}")
```

**Additional Information:**
Successfully tested sending the same CAN-FD message using the Cangaroo application, where it was received correctly. This suggests that the issue may be specific to the 'python-can' library or its interaction with slcan and the CANABLE device.

zariiii9003:
The python-can `slcan` does not support CAN-FD

hartkopp:
Hi @PMCSilva ,

the `slcan` protocol officially does not support CAN FD but only Classical CAN.

Your referenced CANABLE app obviously uses some proprietary extension with extra frame types `b`,`B`,`d`,`D`
See  https://github.com/normaldotcom/cangaroo/blob/master/src/driver/SLCANDriver/SLCANInterface.cpp#L350

A similar suggestion was made here:
https://www.mikrocontroller.net/topic/501223

[...]

Issue #1541, opened 2023-03-16
SLCAN only receives a few messages and then it hangs?


arnolddej:
### Describe the bug

Currently I am using a SLCAN adapter to receive and transmit some CAN messages. 
The problem I encounter is that there are only a handful(mostly only 3) of messages printed to the console. While the SLCAN adapter works fine when using it in a serial terminal. 

Console output of receive_all.py example in console:
```Python
Timestamp: 1678957835.071922    ID: 0bb00c02    X Rx                DL:  4    2f 0a 00 00
Timestamp: 1678957835.072918    ID: 0bb01c02    X Rx                DL:  4    32 0a 00 00
Timestamp: 1678957835.073915    ID: 0bb01c02    X Rx                DL:  4    c2 0c 00 00
```
And that's it. It will hang..

### To Reproduce

I used the receive_all.py example with the interface initialized like this:
```Python
with can.Bus(interface="slcan", channel="COM15@115200",  bitrate=500000, 
                 ) as bus:
```

### Expected behavior

I expected a similar result as when using it manually in a serial terminal. 

Manual serial terminal output snippet
```
V
V0101
S6

O

T1D20F2F2102
T090781008FFFF0F00FFFF0F00
T0BB01C02403040000
T0140800088C72F70B07118010
T00A00C008AC008C72F7070200
T00A01C008AC008C72F7070200
T00200E2280080000000009F44
T00201E22800C3000000003032
T020382221AC
T020782221AC
T0C201F00680119F48CC4F
T00201F0387BCC5B4387F159DA
T0C201F006810804883B4E
T0C201F006820B8A74E34E
T04201F0068DA4C113604F
T04201F0068BA44C03604F
T04201F00697A46DBB614E
T04201F0069DA42A43B04F
T04201F0069EA4637AC04F
T00201F0287BCD5B43780F56DA
T00201F0587BCC5B43780F56DA
T0CF40000403020000
T08E10000403020000
T00200000403020000
T007E0000403020000
T0C200F006800EE0DFDA4F
T0C200F006810843643A4E
T0C200F006820B8837DB4E
T04200F0068877E3C1B04E
T04200F0068977322CB04E
T04200F0068A77E3C1B04E
T04200F0069474AD1DD64D
T04200F006955F94F91DB0
T04200F006965C50A1084D
T04200F0069777E3C1B04E
T00200F01884BC4443780F56DA
T0000000080503FFFE0FFFFE0F
T0000000080502FFFF0F000000
[...]

PR #2028, opened 2026-02-13
Changed slcan serial read routine to be faster and to manage the buffer more carefully

### LINKED ISSUE #2027: Performance fix for slcan
### Describe the bug
There is a performance issue with the way the reads are implemented in slcan.py. 

_read() reads serial data one byte at a time via read(1), even when multiple bytes (or complete frames) are already buffered in the OS serial receive buffer. Each read(1) incurs a syscall and, when no data is available, blocks for up to the serial timeout (default 1ms). A typical SLCAN frame is ~27 bytes, so each frame requires 27 separate read(1) calls minimum. This causes per-frame read latency of 50–850ms instead of <1ms (particularly on windows which seem to have sub-optimal serial support).

I'm using CAN-TP over CAN-FD, so I'm pushing this to the limits, which is why the performance issue has shown up.

### Expected behavior
The serial is read as fast as possible

### Additional context

OS and version: Windows 11
Python version: 3.10
python-can version: 4.6.1
python-can interface/s (if applicable): Canable2 running SL CAN firmware with FD support

### The Fix
I'll submit a pull request with a suggested fix that alters the serial read behaviour to be significantly faster.

aaronfultonnz:
## Summary of Changes
Changes the serial read in slcan to be faster by reading all available bytes at once rather than one at a time. This is a huge performance boost (at least on windows).

- Closes #2027
- Potentially related to #1746
- Potentially related to #1701 (buffer management issue)
- Potentially related to #1541

### DIFF
--- a/doc/changelog.d/2028.changed.md
+++ b/doc/changelog.d/2028.changed.md
@@ -1,0 +1,1 @@
+Improved performance in slcan interface


zariiii9003 (approved):
Looks good. Please add a news fragment so i can merge.

PR #2041, opened 2026-03-16
Fix seedstudio dongle bug

### LINKED ISSUE #2024: SeeedBus: Crash occsionally.
### Describe the bug
python-can crashes rarely with the error:
```
    msg = self.bus.recv(0.0)
  File "/nix/store/0amvifzhc5sv3p1b29dzwxd53njs74z3-python3.13-python-can-4.6.1/lib/python3.13/site-packages/can/bus.py", line 121, in recv
    msg, already_filtered = self._recv_internal(timeout=time_left)
                            ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/nix/store/0amvifzhc5sv3p1b29dzwxd53njs74z3-python3.13-python-can-4.6.1/lib/python3.13/site-packages/can/interfaces/seeedstudio/seeedstudio.py", line 279, in _recv_internal
    end_packet = ord(self.ser.read())
TypeError: ord() expected a character, but string of length 0 found
```

### To Reproduce
Use a while true loop to  open the device quickly.
For my use-case, I needed to implement a flush_rx_buffer() function on user-level:
Please ignore my comments in the code, they may be outdated.
```
    def flush_rx_buffer(self):
        """Discarding the received buffer."""
        # As python-can does not implement any discarding of old messages, see:
        # https://github.com/hardbyte/python-can/issues/1721
        # The recv(0.0) is probably a non-blocking function, see:
        # https://python-can.readthedocs.io/en/3.3.2/interfaces/socketcan.html
        # That non-blocking function should return immediately.
        while True:
            msg = self.bus.recv(0.0)
            if msg is None:
                return

````

### Expected behavior
Instead of a TypeError I expected to receive a None.

### Additional context

OS and version: NixOS
Python version: 3.13
python-can version: 4.6.1
python-can interface/s (if applicable): seeedstudio

I'm using an asus nuc 15, so quite a fast CPU. I don't remember this error to appear on the slow raspberry pi pico.

```python
def func():
    return "hello, world!"
```
</details>

LucienDufour:
## Summary of Changes

It fixes all my issues with seeedstudio can dongle.
Also, python-can does not throw any TypeError's anymore due to "ord()" not failing due to missing data from the serial port.

- Closes: https://github.com/hardbyte/python-can/issues/2024

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [x] I have followed the [contribution guide](https://python-can.readthedocs.io/en/main/development.html). (I did not use tox)
- [ ] All checks and tests pass (`tox`). (Don't know how to run tox on nixos)

### DIFF
--- a/can/interfaces/seeedstudio/seeedstudio.py
+++ b/can/interfaces/seeedstudio/seeedstudio.py
@@ -265,7 +265,13 @@ class SeeedBus(BusABC):
 
         if rx_byte_1 and ord(rx_byte_1) == 0xAA:
             try:
-                rx_byte_2 = ord(self.ser.read())
+                if self.ser.in_waiting >= 1:
+                    rx_byte_2_opt = self.ser.read()
+                else:
+                    # Should not get to this line
+                    logger.debug("received 0xAA, then serial stopped")
+                    rx_byte_2_opt = 0x00
+                rx_byte_2 = ord(rx_byte_2_opt)
 
                 time_stamp = time()
                 if rx_byte_2 == 0x55:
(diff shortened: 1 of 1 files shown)


can/interfaces/seeedstudio/seeedstudio.py
```diff
@@ -273,7 +273,9 @@ def _recv_internal(self, timeout):
 
         if rx_byte_1 and ord(rx_byte_1) == 0xAA:
             try:
-                rx_byte_2 = ord(self.ser.read())
+                if self.ser.in_waiting >= 1:
+                    rx_byte_2_opt = self.ser.read()
+                rx_byte_2 = ord(rx_byte_2_opt)
```
zariiii9003, line 274: if you rebase onto the latest main branch, pylint will complain about possible-use-before-assignment here.
if  `in_waiting` returns 0, this will raise an error.

LucienDufour: Should be solved

LucienDufour:
I have to take this back, as it does not correctly solve the issue.

The claimed end-byte is actually not missed by the seeedstudio device. However, it came slightly later, in a later (USB-) packet. The packets are split at arbitrarily positions.

The best way to handle this framing issue would probably be a ringbuffer that collects all the new incoming bytes, then processes them as soon as they are correctly processable.

PR #2086, opened 2026-08-04
fix: do not assume BLF files only contain log containers

spnda:
## Summary of Changes

The current BLFReader cannot load BLF files which do not make use of log containers but instead store object files at the top level. Vector tools accept these files, so I think this library should, too. This PR simply adds a second call that then parses the objects directly instead of going through the log container route.

- Closes #
- Related to #

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Checklist

- [ ] I have followed the [contribution guide](https://python-can.readthedocs.io/en/main/development.html).
- [x] All checks and tests pass (`tox`).

### DIFF
--- a/can/io/blf.py
+++ b/can/io/blf.py
@@ -202,6 +202,8 @@ class BLFReader(BinaryIOMessageReader):
                     LOG.warning("Unknown compression method (%d)", method)
                     continue
                 yield from self._parse_container(data)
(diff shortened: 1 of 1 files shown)


spnda:
@mergifyio queue

PR #2089, opened 2026-08-15
fix: optimize CAN receive timeout timing with perf_counter

luojiaaoo:
### Description

Replace `time.time()` with `time.perf_counter()` for CAN receive timeout calculations.

`perf_counter()` is better suited for measuring elapsed time and avoids issues caused by system clock adjustments, while also providing better timing precision for the receive loop.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
         # See: https://github.com/pypy/pypy/issues/3808
         TEST_SOCKETCAN: "${{ matrix.os == 'ubuntu-latest' && ! startsWith(matrix.env, 'pypy' ) }}"
     - name: Coveralls Parallel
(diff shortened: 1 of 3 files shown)


fallenmi (changes requested):
Please keep the Coveralls action pinned to a full commit SHA. On exact base b4f82abede25ff83376be793a2935c41f81c3869, all 9 external Actions references are full-SHA pinned; exact head 0cd66392e4151930a50835b38d9ba2827fa6813a changes both Coveralls uses to the mutable v2.3.8 tag, making these the only two mutable refs. Moving that tag would change third-party code executed with the workflow token without a reviewed repository diff. v2.3.8 currently resolves to 8d6379e14d29928660c4ba802d8e85393440b329, so please pin that SHA with a version comment (or retain the prior pin). I separately exercised the timeout loop on the exact base and head; both exit correctly, and the live upstream checks are green.

Reviewed with OpenAI Codex.

